Issue #39, opened 2013-12-31
Edit program crash


applejag:
If the file it tries to open is empty (not a single line written) it crashes at start

```
/bin/edit.lua:77: bad argument #1 to 'len' (string expected, got nil)
```

PR #43, opened 2013-12-31
Attempted Late-Night fix for edit.lua (Issue #39) 

MFernflower:
Sorry if its wrong. Its late at night and I'm tired.


fnuecke:
Thanks, your fix wasn't wrong, but it was just treating the symptoms, not the cause ;)

PR #47, opened 2014-01-02
Some improvements to de_DE.lang

Vexatos:
I think I/O should stay English, the rest is mostly grammar improvements and commas and hyphens, since German language uses quite a lot of them.


fnuecke:
Hmmm, why should I/O stay English? I think E/A is pretty well established. For example it's also used in the Windows device manager: it uses E/A-Bereich as translation for of I/O Range in the disk controller resource information - I just double-checked, didn't want to rely on memories from Windows 95.

As for the commas... yeah. I've long since given up on those, so thanks for fixing those ;)

So, convince me it should be I/O instead of E/A and I'll pull it :)

Vexatos:
Well, it's a computer science term, the I/O, computer science uses mostly English terms, and the I/O is one of those terms which are used in German as well (Don't use Windows as an example, I've seen better translations than the German one of them ;) ). Also, in other mods, the term I/O is used in German as well, so I would stick to that convention. I've honestly never seen the term E/A outside from the Windows translation.

fnuecke:
To some extent, surely. But at least at university there were quite a few German terms flying around ;)
If other mods translate it as I/O (or rather don't), too, though, fine.

PR #71, opened 2014-01-07
Create fr_FR.lang

Pyrolusite:
May need some feedback from fellow French speakers, but this should be fine.


fnuecke:
Thanks :-)

PR #83, opened 2014-01-17
Added package module to /lib

rmellema:
I created a small package module within Lua for Opencomputers and I thought that you might be interested to include it. 

It is based of the original module, but it uses the filesystem of the computer you load it on, so it also lacks the functionality to load C libraries. 
I also added a simple system to detect loops within packages.

### DIFF
--- a/assets/opencomputers/lua/rom/lib/package.lua
+++ b/assets/opencomputers/lua/rom/lib/package.lua
@@ -1,0 +1,87 @@
+package = {}
+
+package.path = "./?.lua;/lib/?.lua;/usr/lib/?.lua;/home/lib/?.lua"
+
+local loading = {}
+
+local loaded = {}
+package.loaded = setmetatable({}, {__index = loaded, __newindex = function(t, k, v) loaded[k] = v end})
+
+local preload = {}
+package.preload = setmetatable({}, {__index = preload, __newindex = function(t, k, v) preload[k] = v end})
+
+package.searchers = {}
+
+function package.searchpath(name, path, sep, rep)
+	assert(name, "You must specify a name!")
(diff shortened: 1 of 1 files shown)


fnuecke:
Sounds great, I'll have a closer look at it later. Thanks!

assets/opencomputers/lua/rom/lib/package.lua
```diff
@@ -0,0 +1,87 @@
...
+	sep = sep or '.'
+	rep = rep or '/'
+	sep, rep = '%'..sep, '%'..rep
+	name = string.gsub(name, sep, rep)
+	local errorFiles = {}
+	for sPath in string.gmatch(path, "([^;]+)") do
+		sPath = string.gsub(sPath, "?", name)
+		sPath = shell.resolve(sPath)
+		if fs.exists(sPath) then
+			local file = io.open(sPath, "r")
```
fnuecke, line 27: Is this part only to test whether the path leads to a directory? In that case `fs.isDirectory` would be quicker, I suppose. I refer to the file = io.open(...) if file then file:close() part.

rmellema, line 27: In the Lua reference, it says that search path only returns files that it can open, so I simply used the same mechanic. But for the default case `fs.isDirectory` would work just as well since all files can be opened for reading.

fnuecke, line 27: Ah, thanks for bringing up the reference. I expected something along those lines, just wanted to make sure. Since it's possible for people to hook in custom file systems it makes sense to keep it as is, I guess. Thanks again!

PR #93, opened 2014-01-24
serializer in lua and change in event dispatch

pingbird:
added my serializer to lua
fixed text.serialize a bit
event.dispatch was changed so when a listener returns false it will run event.ignore on it
ukxodwy

### DIFF
--- a/assets/opencomputers/lua/rom/bin/lua.lua
+++ b/assets/opencomputers/lua/rom/bin/lua.lua
@@ -34,6 +34,6 @@ while term.isAvailable() do
       print(table.unpack(result,2,result.n))
     end
   else
-    print(text.serialize(result, true))
+    print(result)
   end
 end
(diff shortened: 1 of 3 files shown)


fnuecke:
Nice, thanks. I really like the automatic result serialization in the interpreter. Couple of questions, though:
- Github's diff is terrible for those so it's kinda hard to see what actually changed - did you change spaces to tabs or something? If so, please change it to spaces again.
- I'm not a fan of having another serialization implementation in the Lua program, could you change that to somehow re-use text.serialize (with only post-processing like the truncate being in lua.lua)?
- Not that it really matters, but could you give me a test case for why the `^[%a_][%w_]*$` -> `^[%a_]([%w_]*)$` change is necessary? I'm curious.

pingbird:
fixed
as for the pattern, i derped and thought
^[a][b]*$ matched abab instead of abbb :/

PR #97, opened 2014-01-25
Updated de_DE.lang

Vexatos:


fnuecke:
Universal Electricity definitely has a space in its name. http://universalelectricity.com/

(Oh, und: Minecraft duzt seine Spieler? Na gut :-P ich hab's immer auf Englisch, daher k.A. was die Konvention ist)

Vexatos:
@calclavia can't decide where to put spaces in the names :|
I'll update it.
And yes, since I used "Du" in all of my other translations since I think it fits better into Minecraft, I chose it here as well (I haven't seen any "Sie" in Minecraft yet)

PR #100, opened 2014-01-26
Well, this is German grammar! (Updated de_DE.lang)

Vexatos:


fnuecke:
Exactly. _Die_ CPU. Unit = Einheit. If I had written "Hauptprozessor" it'd be another story, but as it stands "keine" is correct. ;-)

Vexatos:
Everyone in Germany calls it "Der CPU", I know it is wrong, but well, there's a lot wrong in German-English _cough Handy cough_, so maybe think about this twice.

fnuecke:
Everyone I know calls it "die CPU" ^^ Maybe it's a local thing. Change it to "kein Hauptprozessor" if you want and I'll pull it.

PR #105, opened 2014-01-28
debug.traceback

pingbird:
this will be extremely useful in debugging, and allows for functions to "track" what called them


fnuecke:
Thanks for the reminder, I wanted to have a look at that at some point, to check if it's safe to keep in the sandbox. Just did, and I'd say it is.

Vexatos:
Soo... what does this do exactly?

SpiritedDusty:
It allows you to find out what function called which function.

PR #115, opened 2014-02-11
Added Russian localization.

YuRaNnNzZZ:


lordjoda:
Thank you!

src/main/resources/assets/opencomputers/lang/ru_RU.lang
```diff
@@ -0,0 +1,170 @@
+# Russian localization by YuRaNnNzZZ
+# Use UTF-8 for this file.
+
+# Blocks
+oc:block.Adapter.name=Адаптер
+oc:block.Cable.name=Кабель
+oc:block.Capacitor.name=Аккумулятор
```
mishkapp, line 7: "Кондесатор" instead "Аккумулятор", it will be closer to original translation

fnuecke:
If I read this correctly, "Конденсатор" (Кондесатор is a typo, no?) would be equivalent to the German "Kondensator", whereas "Аккумулятор" is more like "Akkumulator"? Yeah, then Конденсатор would be more appropriate I think. I'll change it.

mishkapp:
Yes, these words are similat to each other in German and Russian in writing, pronunciation and sense.

PR #134, opened 2014-02-23
Mc1.7

Iworb:
Please, add to the robots way to spill of liquids from cells.


fnuecke:
Err, this isn't really a PR is it? Please open this again as an issue instead, thanks.

PR #141, opened 2014-02-28
Update de_DE.lang

Vexatos:
Extremely dangerous comma mistake, could have caused MassivePunctuationMistakeException. This comma there is the most important thing you'll have ever seen in your whole life.
</irony>


fnuecke:
I couldn't agree more!

PR #168, opened 2014-03-22
An important one.

Vexatos:
If you did not know, search for it, you'll maybe find out.


rmellema:
if you mean [this one](http://www.newscientist.com/blog/technology/2007/07/programmable-robot-from-60ad.html), it might be better to place it at the real robots section I guess. 

Otherwise, I have no clue what it might be

Vexatos:
You don't need to know ;) It's a pretty deep insider.

PR #203, opened 2014-04-08
Because German. And punctuations.

Vexatos:
Check Vanilla localization.


fnuecke:
Again! Thanks as usual ;)

PR #240, opened 2014-04-29
Further improved irc.lua

Techokami:
Properly renders the /me command in chat, and uses computer.beep() when your nick is said in chat.


fnuecke:
Thanks!

PR #243, opened 2014-04-30
Created note API

Vexatos:
Functionality:
- `note.midi(n: number or string)`: Converts a note in string form (e.g. A#4 or Gb3, will be documented on the wiki) or a given frequency into the respective MIDI code
- `note.freq(n: number or string)`: Converts a note in string form (e.g. A#4) or a given MIDI code into the respective frequency
- `note.name(n: number)`: Converts a MIDI value back into a string; if you have a frequency to convert, just use note.get(note.midi(frequency))
- `note.ticks(n: number)`: Converts note block ticks (0-24) into MIDI code (34-58, respectively) and vice-versa. Useful for use with note blocks and OpenComponents
- `note.play(tone: string or number,duration: number)`: Plays a note from a string or MIDI code via computer.beep with the specified duration

### DIFF
--- a/src/main/resources/assets/opencomputers/lua/rom/lib/note.lua
+++ b/src/main/resources/assets/opencomputers/lua/rom/lib/note.lua
@@ -3,7 +3,7 @@
 local computer = require("computer")
 
 local note = {}
---The table containing all the notes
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks again!

PR #248, opened 2014-05-06
lua now displays return values

Wuerfel21:
Lua.lua displays return values of lines ehich do not contain = or end (this should be everything which does not like to be put inside of term.write, feel free to add stuff i missed)

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OCPiano/piano.lua
+++ b/src/main/resources/assets/opencomputers/loot/OCPiano/piano.lua
@@ -1,0 +1,86 @@
+local shell = require("shell")
+local component = require("component")
+local keyboard = require("keyboard")
+local event = require("event")
+
+local args,options = shell.parse(...)
+
+local MagicTable = {}
+
+--stuff to make life easier
+local note
+local gpu
+
+--this saves me from typing "keyboard" over and over again
+local keys = keyboard.keys
+
+local function MakeTheMagicHappen() --Fill MagicTable with keybindings
+  MagicTable[keys["1"]] = 1
+  MagicTable[keys["q"]] = 2
+  MagicTable[keys["2"]] = 3
+  MagicTable[keys["w"]] = 4
+  MagicTable[keys["3"]] = 5
+  MagicTable[keys["e"]] = 6
+  MagicTable[keys["r"]] = 7
+  MagicTable[keys["5"]] = 8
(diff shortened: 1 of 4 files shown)


fnuecke:
Thanks; printing the simple, non-serialized output by default seems fair enough. I'd prefer a more robust check than a test for '=' or 'end' anywhere in the command, though, and prefer io.write over term.write, since that's more general.

So, something like `if not string.find(command, "^=") and load("io.write(tostring(" .. command .. "))") then`, maybe? (note: untested, just an idea)

Also, if this is added, the message regarding the leading '=' on program start should go.

Wuerfel21:
Don't quite understand the last part(i am from germany) about the =
Also i cant remember any valid line that will crash it without end or = (besides from comments, but why would ever someone type comments in a interpreter)
Also please forgive me if i done something wrong (first pull request, also bit more experience on CC(but then found OC and "WOW")) some last note, why is io better than term?

fnuecke:
Currently, when starting `lua` there's a message printed saying one needs to prefix the expression with `=` to print its value. That would be no longer true with such a change.

Example for it 'breaking' besides for comments (which is bad enough, since it could be avoided):
`"this is the " .. "end"`
Expected: the resulting strings to be displayed.
Actual result: error.
Disclaimer: this is without testing, but what I'd expect from looking at the code.

You committed something unrelated to the branch the PR is based on. Feel invited to do a fresh PR with an improved version, but I'll be closing this one for now.

`io` is prefereable IMHO because it is 'standard Lua', as opposed to `term`, which is OC specific. I'd encourage non-specialized code, since it a) makes it easier for people with Lua experience to understand the code b) makes it easier for people who learn Lua that way to apply their knowledge outside OC.

Wuerfel21:
You committed something unrelated to the branch the PR is based on. Feel invited to do a fresh PR with an improved version, but I'll be closing this one for now. Whats a PR?

AEnterprise:
PR is a pull request

Wuerfel21:
thanks

Wuerfel21:
See me releasing some loot discs soon
BTW i cant edit comments on smartphone

Wuerfel21:
See me releasing some loot discs soon
BTW i cant edit comments on smartphone

Wuerfel21:
Oh an sorry for double posting omg my S3 dont likes github

Issue #252, opened 2014-05-09
OpenPeripherals methods inssue


marcin212:
Whenever I call any method from OpenPeripherals that returns an array, it returns a nil instead. The console output looks like this: http://pastebin.com/raw.php?i=v3M3fmHz

ComputerCraft1.63
OpenComponents-MC1.6.4-0.1.1.37-universal
OpenPeripheralCore-1.6.4-0.4-snapshot-127
OpenModsLib-1.6.4-0.4-snapshot-196
OpenComputers-MC1.6.4-1.2.9.362-universal

fnuecke:
The method in OP tries to 'yield' / pull an event, which is not supported by the CC peripheral wrapper (as the name of the exception indicates). Sorry, but this cannot be 'fixed' due to the way OC works (specifically, its persistence).

marcin212:
You need both fixes to have working events

PR #254, opened 2014-05-13
Add support for ComputerCraft Events

marcin212:
Add support for ComputerCraft Events
fix: https://github.com/MightyPirates/OpenComputers/issues/252
Now it wraps the peripheral correctly and returns proper values instead of null/nil.

"It always seems impossible until its done." - Nelson Mandela


fnuecke:
Hmm, I'm not sure how this makes a difference? Lua does not have 'nil' values in tables - setting something to nil in a table effectively removes that entry, so I don't understand why this case would be necessary?

Kubuxu:
I was writting this patch with him. For some weird reason if map contained key with null value assigned to it it was converted to scala.Tuple2 and crashing table conversion.

fnuecke:
Ah, right, the pattern matcher will fail for those cases. Thanks for pointing this out.

PR #277, opened 2014-05-28
Update default.recipes

erinpentecost:
The recipe for cpu2 now uses diamonds instead of emeralds. It didn't match the nuggetIron -> nuggetGold -> diamond pattern used by all other components.


fnuecke:
I see your point, but that makes the diamond cost spike too much IMHO. Introducing 'diamond chips' (as an equivalent to nuggets) has been suggested before... those might make sense in this recipe, amongst others. It's something I'll have a look into when I find the time.

PR #278, opened 2014-05-30
Recursive conversion with Converters. Circural reference protection and ...

Kubuxu:
...caching

Signed-off-by: Jakub Sztandera kubuxu@gmail.com


fnuecke:
Thanks. I couldn't merge it directly, but I expanded it to also support cyclic maps / lists :-)

Issue #259, opened 2014-05-18
Hologram rendering performance


evg-zhabotinsky:
Relevant part of initial message (after testing):

> I use version 1.2.9.362 for MC 1.6.4
> Minecraft  freezes for a moment _when_ you are **looking at** hologram _and_ its part is updated.

**More testing**

It turned out that hologram updates cripple performance less on my integrated video (intel hd 4600)
With all minimum settings it runs 80 fps normally, 24 when I see random hologram and 15-20 when randomly updating

On discrete video (nvidia gt 730m) with all minimum settings it runs **300 fps** normally, same with empty hologram, 120 fps when I see fully filled hologram, 25 (!) fps when I see randomly filled hologram and **5 (!!!) fps** when I it updates randomly

Amount of pixels covered by hologram does not affect fps so it has nothing to do with transparency, etc.

Also I took a look at render code and it seems to me that update lags are more noticable due to render list recompilation and. Fps show that it also lags when not updating (although not so notably).

Fps changes and code suggest that amount of rendered quads (borders between 'on' and 'off' voxels) is what affects performance most, but my GPU should be able to render much more than 5 millions triangles a second (~200000 triangles at ~25 fps) (Minecraft renders at least 10 times more!) (again, it has nothing to do with their area, only amount).

I do not know what can be done. Maybe I just should not draw random holograms :)

fnuecke:
Yeah, depending on the complexity of the geometry you make the projector display the framerate can drop quite badly. The worst-case is pretty much 50% 3D checkerboard fill. I do plan to invest some time into optimizing this, but it's a complex topic, and time is the one thing that there's never enough of.
Some ideas I have are
- culling 'internal' geometry (i.e when some area is never visible from 'outside' it will just not be rendered, even when stepping into the hologram).
- merging quads on large planar faces.
- possibly taking the player's position into account to cull faces beforehand.
- divide the hologram into 'chunks' and update them individually (it's pretty small, so I had hoped that wouldn't be necessary, but...)

Will have to see how much can be 'optimized away' before the optimization gets more expensive than the actual rendering, though...

If someone who has more experience with stuff like this reads this, some contributions - and if it's just ideas and pointers - would be very much appreciated.

KFAFSP:
[...]

PR #294, opened 2014-06-06
VBO hologram renderer

evg-zhabotinsky:
I finally did it.
Render code totally rewritten. Fps is much better.
Line in PacketHandler.scala is removed because rescaling no longer requires recompilation.
~~Also, consider doing something with d99bc5815c680198dfa7cd94eede40e5b4e66296 because it causes synchronous hologram updates every 2 seconds which causes freezes.~~ solved

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -66,6 +66,12 @@ opencomputers {
       # hologram projector when at maximum scale. Render distance is scaled
       # down with the actual scale of the hologram.
       hologramRenderDistance: 64
+
+      # The distance at which to start fading out the hologram (as with
+      # hologramRenderDistance). This is purely cosmetic, to avoid image
+      # disappearing instantly when moving too far away from a projector.
+      # It does not affect performance. Holograms are transparent anyway.
+      hologramFadeStartDistance: 48
   }
 
   # Computer related settings, concerns server performance and security.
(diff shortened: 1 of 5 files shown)


Vexatos:
Don't tell me I'll be able to use Hologram projectors with more than 0.2 FPS :o
Thank you!

evg-zhabotinsky:
Well, if you have 0.2 FPS now I'm afraid it's not going to help a lot. It will be faster but still unplayable.

evg-zhabotinsky:
Now this issue #259 is almost irrelevant on my machine. No lag if I am not looking at ~10 simultaneously updating checkerboard holograms. So I guess it should be playable on older hardware now.

fnuecke:
Wow, this is amazing! Thanks so much for working on this :)

I had a look at it, and what I am noticing is that there's no 'lighting' at all - right now the faces are lit slightly differently (like normal blocks) to get a better feeling for the hologram's depth. In your implementation it's completely 'flat'. Which is probably because you don't set any normals in the VBO, yes? Could you add those?

Current:

Patched:

evg-zhabotinsky:
Hmm I explicitly turned lighting off and commented on it (line 250):
`GL11.glDisable(GL11.GL_LIGHTING) // Hologram that reflects light... It would be awesome! But...`
Well, it really looks better with lighting no matter how realistic it is.
~~**UPD** Looks like tesselator does all the lighting in Minecraft and I'll have to emulate it's behavior.~~
**UPD2** Looked through decompiled tesselator code. It does nothing special to normals. ~~I do not understand why lighting does not work in my code.~~
**UPD3** Okay. How obvious is that glScale also scales normals? They are just directions after all!

fnuecke:
Brilliant! I'm looking through it right now :)

fnuecke:
And it's in. Again, thanks a lot!

PR #299, opened 2014-06-06
Added functionality to robots to allow item sorting

erinpentecost:
Added itemID() function to robots, allowing them to extract the ID, subtype, and name of an item inside the inventory.

Extended compare() and compareTo() to also return the difference of the id
- (subtype/1000) of both items being compared. This allows for item
  sorting. The first returned value is still a boolean, so this should be
  backwards compatible with existing Lua scripts.

### DIFF
--- a/src/main/resources/assets/opencomputers/lua/component/robot/lib/robot.lua
+++ b/src/main/resources/assets/opencomputers/lua/component/robot/lib/robot.lua
@@ -61,6 +61,10 @@ function robot.transferTo(slot, count)
   return component.robot.transferTo(slot, count)
 end
 
+function robot.itemID(slot)
+  return component.robot.itemID(slot)
+end
+
 -------------------------------------------------------------------------------
 -- Inventory + World
 
--- a/src/main/scala/li/cil/oc/server/component/robot/Robot.scala
+++ b/src/main/scala/li/cil/oc/server/component/robot/Robot.scala
@@ -96,11 +96,25 @@ class Robot(val robot: tileentity.Robot) extends ManagedComponent {
   @Callback
   def compareTo(context: Context, args: Arguments): Array[AnyRef] = {
     val slot = checkSlot(args, 0)
-    result((stackInSlot(selectedSlot), stackInSlot(slot)) match {
-      case (Some(stackA), Some(stackB)) => haveSameItemType(stackA, stackB)
-      case (None, None) => true
-      case _ => false
-    })
+    (stackInSlot(selectedSlot), stackInSlot(slot)) match {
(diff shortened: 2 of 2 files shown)


evg-zhabotinsky:
Good. But this is a bit... unrealistic. Why the hell should dirt come after stone but before cobblestone? Item ID is a meaningless and really inconsistent metric.
Even the already present compare() function works strange. For example, it distinguishes 3 kinds of floppies: OpenOS, treasure disks (dig, besh, etc.) and all others.

ComputerCraft developers refused to allow getting item IDs. They only provide various compare functions (using turtles). It works the same way in OpenComputers at the moment.
Eloraam (in her long dead but anyway epic RedPower mod) decided to provide hashes for item ID and damage as a way to identify items (using Sortron) and these hashes differ between worlds but are the same if item ID and world are the same. More like comparing item to it's digital photo instead of another item.
I don't think @fnuecke is going to accept this pull request as is.

However, having a way to compare items without having them both at the same time (or even remotely) is definitely interesting. So I'd like to see a function that returns a few identifying hashes but I am not sure which exactly. For example,one hash could be a string (like peripheral ID) generated from the same values that compare() uses and some per-world random number. Also, compare() needs consistency improvement in this case.

erinpentecost:
I agree that the id difference doesn't make much sense in the context of what an item is, but that isn't really in the scope of what I was trying to accomplish. I want some way to impose a sorted order on items so it will be possible to use robots for item storage and retrieval. I'd prefer not to hash the combined id and subtype, because, for example, it would keep all the different types of saplings next to eachother.

What is the reason for making these results world-specfic? I think I'm missing something here.

Thanks!

evg-zhabotinsky:
World-dependent hashes are for "realism" if it can be called that way. It is a bit odd when every object kind is assigned an arbitrary fixed number for no reason.
Also, I don't think ordering by ID is meaningful. For example, in OpenComputers memory tier 1.5 goes after all others with a lot of other items in between and diamond server goes before the other two. And many other mods have much less meaningful ID order. I really hate to search for something in NEI when it refuses to find that item by name or when I forget the name. And you can enforce your own sorting order if you can identify items.

erinpentecost:
Ok, you convinced me. I can change id/subtype to a hash of type string if needed.

Are unlocalized names unique? If so, we could just use those. Then the compare functions would just use alphabetical order, which would still group subtypes together.

Techokami:
Unlocalized names should be considered unique, as they are the strings used in the /give command.

erinpentecost:
Ok, I'll rewrite itemID() to only return the unlocalized name, and I'll rewrite the compare functions to use alphabetical comparison on the same. Sound acceptable?

PR #303, opened 2014-06-07
Updated the mod ID of gregtech

xbony2:
The mod ID of gregtech changed in 1.7, so I fixed it.

### DIFF
--- a/src/main/scala/li/cil/oc/util/mods/Mods.scala
+++ b/src/main/scala/li/cil/oc/util/mods/Mods.scala
@@ -12,7 +12,7 @@ object Mods {
     }
   }
   val ForgeMultipart = new SimpleMod("ForgeMultipart")
-  val GregTech = new SimpleMod("gregtech_addon")
+  val GregTech = new SimpleMod("gregtech")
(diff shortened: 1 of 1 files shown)


fnuecke:
Hmm, interesting. It changed in the mcmod.info, but it's [still the old in the annotation of the class](http://i.imgur.com/oqV23bj.png), and I think that one wins (in particular with `useMetadata=false`)? Someone should probably tell Greg...

Are you 100% sure it doesn't work as-is, but will with that change? (Sorry, don't have the time to test it myself right now)

xbony2:
I haven't tested it actually. I was going to work on gregtech support for my mod, and I heard the mod ID changed from "gregtech_addon" to "gregtech" from 1.6 to 1.7. After that, I looked for all the mods on github that supported gregtech (just this and gany's nether) and edited their code.

fnuecke:
Ah wait, I derped, I read that the wrong way around. Yeah, `gregtech` will probably work. The `mcmod.info` file in GT should probably still be updated, though. Thanks for letting me know!

PR #304, opened 2014-06-08
Fixes #31 for current 1.3 upstream.

### LINKED ISSUE #31: OpenComputers tends to crash in ScreenRenderer
I'm finding that when using OpenComputers, there is a chance that it will end up crashing the entire JVM down due to something with ScreenRenderer

Note that may be helpful:
I was playing on a server yesterday and my friend kept crashing out completly. After looking through his crash logs it was again atioglxx.dll and li.cil.oc.client.renderer.tileentity.ScreenRenderer We got him to update his drivers and it seemed to go away after that, but today his issue came back, both involving li.cil.oc.client.renderer.tileentity.ScreenRenderer and atioglxx.dll
I've also had this crash and have attached my log to the bottom. Mine also involves li.cil.oc.client.renderer.tileentity.ScreenRenderer and atioglxx.dll.
I had a different friend who couldn't join the server at all without completly crashing the moment on join. His involves atio6axx.dll though but li.cil.oc.client.renderer.tileentity.ScreenRenderer was still there.

My Log: http://pastebin.com/8NhuKcKg
Friends log: http://pastie.org/private/ix5kfnxk3e2746ybhfzaka

evg-zhabotinsky:
Removed nested display lists from screen renderer. It theems to work.
Should not slow anything down.

### DIFF
--- a/src/main/scala/li/cil/oc/client/renderer/MonospaceFontRenderer.scala
+++ b/src/main/scala/li/cil/oc/client/renderer/MonospaceFontRenderer.scala
@@ -38,51 +38,15 @@ object MonospaceFontRenderer {
   })
 
   private class Renderer(private val textureManager: TextureManager) {
-    /** Display lists, one per char (renders quad with char's uv coords). */
-    private val charLists = GLAllocation.generateDisplayLists(256)
-    RenderState.checkError("MonospaceFontRenderer.charLists")
-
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks! I was aware display lists couldn't be built recursively, but not that they mustn't call other display lists in turn. It's still odd that this worked _most_ of the time, and only crashed occasionally.

PR #305, opened 2014-06-08
Fixes #31 for 1.2.12 (I guess it's time for 1.2.13)

### LINKED ISSUE #31: OpenComputers tends to crash in ScreenRenderer

evg-zhabotinsky:
Removed nested display lists from screen renderer. It seems to work.
I think it should end up in v1.2_backports branch but I don't know how to tell that to Github. That said, you might have to merge it manually into that new branch. This commit is already branched from v1.2.12 so nothing should go wrong.

### DIFF
--- a/src/main/scala/li/cil/oc/client/renderer/MonospaceFontRenderer.scala
+++ b/src/main/scala/li/cil/oc/client/renderer/MonospaceFontRenderer.scala
@@ -25,51 +25,15 @@ object MonospaceFontRenderer {
   })
 
   private class Renderer(private val textureManager: TextureManager) {
-    /** Display lists, one per char (renders quad with char's uv coords). */
-    private val charLists = GLAllocation.generateDisplayLists(256)
-    RenderState.checkError("MonospaceFontRenderer.charLists")
-
-    /** Buffer filled with char display lists to efficiently draw strings. */
-    private val listBuffer = GLAllocation.createDirectIntBuffer(512)
-    RenderState.checkError("MonospaceFontRenderer.listBuffer")
-
     private val (charWidth, charHeight) = (MonospaceFontRenderer.fontWidth * 2, MonospaceFontRenderer.fontHeight * 2)
     private val cols = 256 / charWidth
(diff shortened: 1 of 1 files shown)


Techokami:
I don't think 1.2.13 is going to happen... probably will be in 1.3, though.

gamax92:
There's a possibility, 1.3 is still unstable

@evg-zhabotinsky There is an Edit button when you make the Pull Request, you can change the branches to pull from and where to merge.

evg-zhabotinsky:
I pressed that button, chosen v1.2.12 and then "create pull request" button disappeared. Switched it back to master and the button is back. Maybe I just didn't understand something.
@Techokami As one alternative it might remain just a "fixed 1.2.12 jar" and not merged into git. Anyway it differs from the other commit in single line. Filed it just for convenience.

fnuecke:
Thanks again; I'm hopeful I can push a first prerelease of 1.3 any day now, though (haven't had any crashes in the native lib reported in a while, so I guess the fix worked). If I do hit another roadblock I will do a 1.2.13, though.

PR #306, opened 2014-06-08
Allow installation on 1 Tier1 memory

gamax92:
Before, installation would crash half way through with "not enough memory", and a partially copied OpenOS is left on the disk.


Vexatos:
:+1: T1s shouldn't be entirely useless.

Wuerfel21:
Agree, but in the meanwhile you could try floppypirate(in my repo in openprograms)

fnuecke:
Thanks!

PR #310, opened 2014-06-10
Typo Wars, Episode V

Vexatos:
"The Punctuation strikes back"


fnuecke:
What a wonderful example for why I decided to ignore commas a while back :P I can't even, read that, sentence, fluidly anymore because, of all the, commas.

rmellema:
Although technically probably correct, (I don't know German comma-rules) I have to agree with @fnuecke here, it doesn't read as fluently if you insert a comma after every three to five words. Dutch has similar rules, but I never follow them for the same reason.

Vexatos:
German has far too many commas, I agree. But I cannot decide how my language is done, hence this Pull Request.

PR #311, opened 2014-06-10
Made dataBuffer be common for all holograms. Also thoughts about hologram memory usage.

evg-zhabotinsky:
The reason was that my client crashed because garbage collector wasn't able to free 486 MiB for all these dataBuffers for 36 holograms I was looking at (still alive @4FPS on my Intel HD 4600 by the way)

However, the main problem I didn't quite think before is the fact that **holograms consume tons of video memory**. Common 'skeleton' uses _47.25 MiB_ and each hologram (_even empty!_) eats up _extra 13.5 MiB_ of video memory. Although GPU is usually able to use part of system RAM when no free video memory remains, it has huge performance penalty and discrete GPUs are usually limited to 2 or 4 GiB total.

### DIFF
--- a/src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
+++ b/src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
@@ -15,15 +15,15 @@ import org.lwjgl.BufferUtils
 import li.cil.oc.Settings
 import java.nio.IntBuffer
 
-object HologramRenderer extends TileEntitySpecialRenderer with Callable[(Int, IntBuffer)] with RemovalListener[TileEntity, (Int, IntBuffer)] with ITickHandler {
+object HologramRenderer extends TileEntitySpecialRenderer with Callable[Int] with RemovalListener[TileEntity, Int] with ITickHandler {
   private val random = new Random()
 
   /** We cache the VBOs for the projectors we render for performance. */
   private val cache = com.google.common.cache.CacheBuilder.newBuilder().
     expireAfterAccess(10, TimeUnit.SECONDS).
     removalListener(this).
-    asInstanceOf[CacheBuilder[Hologram, (Int, IntBuffer)]].
-    build[Hologram, (Int, IntBuffer)]()
+    asInstanceOf[CacheBuilder[Hologram, Int]].
+    build[Hologram, Int]()
 
   /**
    * Common for all holograms. Holds the vertex positions, texture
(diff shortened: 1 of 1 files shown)


fnuecke:
Ah, the buffer can be flushed after it's been set, even in dynamic mode. That's great then.

As for the memory consumption, I guess it can be problematic for low-end systems and when people start spamming holograms all over the place to troll those low-end systems...

I've been experimenting a little, and using `dataBuffer.flip()` instead of `dataBuffer.rewind()` combined with a followup `dataBuffer.clear()` reduces the VBO size to what it needs to be vertex-index-wise; still uses the full space for color, though. I don't see an obvious way to interleave color data with element indices, and I'm not sure iterating once to get the actual number of visible quads would be worth it. I'll see if / how badly that'd impact framerate now.

_Edit_: ah, crap, the color buffer has to be the actual size of the verts since it's indexed and not read from linearly of course :- ( Oh well, at least it reduces the checkers (worst-case, I believe?) scenario to 10MB.

fnuecke:
For the memory problem in general, I think if clients do run into problems it's acceptable to have them reduce their hologram render distance. Oh, I also made it so the buffer is cleared if the hologram displays nothing at all, so randomly placed holograms don't take up video memory, people have to actually set something on them for them to do so now.

evg-zhabotinsky:
Maybe I can use 3D texture and multi-texturing to color cubes. It would be 74.25 MiB for common skeleton and just 288 KiB for each hologram to store 3D texture with contents.
~~**UPD** OpenGL can even fill this 3D texture using color-index image and palette so only 72 KiB will need to be processed on Java side. (Or even no extra Java-side processing if hologram is stored as byte array and not as several bitmasks.)~~ Indexed color is deprecated and generally poorly implemented.

evg-zhabotinsky:
A great idea came to me right now!
Fragments that are completely transparent are just thrown away and even don't update depth buffer.
Considering that, we don't need to draw individual cubes. We can just draw the whole plane with some transparent squares! It will be just (32 + 48 + 48) \* 2 = 256 quads! It's just a surface of single chunk on superflat.
_It will completely eliminate both memory usage and rendering performance problems!!!_
The only problem is that 2x2 hologram texture. It won't let transparent squares be absolutely transparent and they will update depth buffer. But there might be a way around that (aside from throwing this texture away)
**UPD** GL_MODULATE should work as I understand. I am already writing it. ~~Also, although direct use of VBOs is faster, I think we can afford cleaner tesselator-based approach with that miserable amount of quads.~~ Never mind, VBOs are already working and clean enough.

fnuecke:
So, no luck? :/ Thanks a lot for looking into this further nonetheless. Maybe splitting the hologram into 9 'chunks' might still keep render performance decent by reducing the memory size of the static geometry buffer a bit?

PR #317, opened 2014-06-12
Dynamic Maze Generator

ReTeYeZ:
I think the name speaks for itself. It uses Depth-First search algorithm.
I testet it and it should work fine :). Hope you and others like it.


fnuecke:
Nifty :-)

If you have the time, it'd be great if you could also add a manpage - it took me a bit of experimenting and reading the code to figure out how to properly use it ;-)

PR #325, opened 2014-06-14
Small fix to the error output

ReTeYeZ:
I wouldn't mind to add a small GUI but I thought this is how most programs are executed. I actually made it like this to prevent confusion. Maybe it's a bit easier to understand after this fix. Just tell me what you think.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/MazeGen/maze.lua
+++ b/src/main/resources/assets/opencomputers/loot/MazeGen/maze.lua
@@ -23,26 +23,9 @@ cellHeight   = tonumber(input[3]) or 1
 cellDiameter = tonumber(input[4]) or 1
 autoRefuel   = input[5] or "false"
 
-if #input == 1 and input[1] == "help" then
-    print("The maze will always start at the bottom left")
-    print("corner. The first two parameters are maze width")
-    print("and maze length. The third, forth and fifth")
-    print("parameters, which are optional, are cell")
-    print("height, cell diameter and auto refuel.")
-    print("Keep in mind that increasing the cell diameter")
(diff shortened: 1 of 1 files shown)


fnuecke:
Ah, while it'd probably make sense to adjust the 'usage' info to mirror the format of the other programs, that's not what I actually meant, I was referring to a little 'readme' alike [the one that comes with `dig`](https://github.com/MightyPirates/OpenComputers/blob/master/src/main/resources/assets/opencomputers/loot/TheDig/dig.man) and is present [for most built-in programs](https://github.com/MightyPirates/OpenComputers/tree/master/src/main/resources/assets/opencomputers/loot/OpenOS/usr/man).

ReTeYeZ:
Ok I hope this will do it for now.
I'll later take a look at the other programs how it's implemented there.

Feel free to bother me again if it's not clear enough.

ReTeYeZ:
Those are 'arguments' not 'parameters' right? xD
Not sure about that. If so can you pls change that?

ReTeYeZ:
Just checked the dig.man(and others). Are they accessable ingame? Or are they more of a manual for the GitHub users to get more in-depth knowledge of how a code works?

edit:
Finaly found the edit button^^

fnuecke:
Yeah, it's accessible ingame. The ones in /usr/man directly via `man x` or `help x` (e.g. `man ls`), the ones on disk via `man/dig` when the working directory is that disk (i.e. `cd /mnt/2af` followed by `man dig` would work, assuming the dig disk is mounted at /mnt/2af).

ReTeYeZ:
Ah didn't knew that. Ok I'm on it

PR #326, opened 2014-06-15
Added maze.man

ReTeYeZ:
Here you go. I hope this will do it. I'm not that good at explaining especially in english :/.


fnuecke:
Thanks, looks good :)

PR #334, opened 2014-06-16
Made history longer

Wuerfel21:
It was too short, i always needed to retype long commands

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -7,6 +7,14 @@ local shell = require("shell")
 local term = require("term")
 local text = require("text")
 
+local function getHistSize()
+  if os.getenv("HISTSIZE") then
+    return os.getenv("HISTSIZE")
+  else
+    return 10
+  end
+end
+
 local function expand(value)
   local result = value:gsub("%$(%w+)", os.getenv):gsub("%$%b{}",
     function(match) return os.getenv(expand(match:sub(3, -2))) or match end)
@@ -165,7 +173,7 @@ if #args == 0 and (io.input() == io.stdin or options.i) and not options.c then
         term.write("exit\n")
         return -- eof
       end
-      while #history > 20 do
+      while #history > getHistSize() do
         table.remove(history, 1)
       end
       command = text.trim(command)
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -7,14 +7,6 @@ local shell = require("shell")
 local term = require("term")
 local text = require("text")
 
-local function getHistSize()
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
```diff
@@ -165,7 +165,7 @@ if #args == 0 and (io.input() == io.stdin or options.i) and not options.c then
         term.write("exit\n")
         return -- eof
       end
-      while #history > 10 do
+      while #history > 20 do
```
Wuerfel21: derp, 2 times more history

fnuecke:
That's still kinda arbitrary, and the low limit was intended to avoid (relatively) non-obvious memory consumption. I think the better approach would be to add a `HISTSIZE` env var (defaulting to 10) and use that there. Then you can just `set HISTSIZE=20` in the shell and be happy ;-)

Wuerfel21:
Please hold the line, your request is beeing processed...

Wuerfel21:
Well, I done it(hopefully right)

src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
```diff
@@ -7,6 +7,14 @@ local shell = require("shell")
 local term = require("term")
 local text = require("text")
 
+local function getHistSize()
```
Wuerfel21: This Function does stuff(really?) to either return HISTSIZE or 10

src/main/resources/assets/opencomputers/loot/OpenOS/boot/01_os.lua
```diff
@@ -16,6 +16,7 @@ local env = {
   SHELL="/bin/sh",
   TMP="/tmp", -- Depricated
   TMPDIR="/tmp"
```
fnuecke: Missing a comma there. Test your stuff! :O

src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
```diff
@@ -165,7 +173,7 @@ if #args == 0 and (io.input() == io.stdin or options.i) and not options.c then
         term.write("exit\n")
         return -- eof
       end
-      while #history > 10 do
+      while #history > getHistSize() do
```
fnuecke: Maybe just a `(tonumber(os.getenv("HISTSIZE")) or 10)` here, do away with the function?

src/main/resources/assets/opencomputers/loot/OpenOS/boot/01_os.lua
fnuecke: Also, probably better make it `HISTSIZE="10"`, for consistency (don't mix in non-string types into env-vars).

Wuerfel21:
ok

fnuecke:
Thanks :)

Wuerfel21:
Phew... No more typeing madness...

PR #342, opened 2014-06-19
Made ls display amount of dirs/files

Wuerfel21:
Self explanatory

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/ls.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/ls.lua
@@ -66,7 +66,7 @@ for i = 1, #dirs do
     end
     setColor(0xFFFFFF)
     if options.c then
-      io.write("\nDirectorys: "..tostring(#lsd))
+      io.write("\nDirectories: "..tostring(#lsd))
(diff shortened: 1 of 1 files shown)


fnuecke:
a) Directories, not Directorys.
b) Why?
c) Does 'normal' ls ever do this? Is it a flag? If so, make it one. Otherwise, I'm not a fan :P

bendem:
`ls` is just a list of names on columns. The difference is visible if you do `ls -p` because it adds a trailing `/` to directories: https://gist.github.com/bendem/29b02230c90b9681b14b

Wuerfel21:
@fnuecke a)english is not my native language, and it was late in the night when i wrote this
b)if you have many files in a folder a d you want to know how many they are, currently you need to count them yourself c) i havent messed around with linux that much, so i cant talk about it, but why have an exact copy? Btw, dos did it,

Kubuxu:
Because whole OS is made similar to Linux/Unix and it is natural environment for some people.
`ls` was designed for parsing its output that why it they wanted to add something they added parameter to activate it.

Wuerfel21:
Made it by option

bendem:
It should still be "directories" and not "directorys" :)

PR #356, opened 2014-06-24
Made more.lua command accept more keys

jamesb6626:
Added more keys for the command to recognise ("page down" goes to the next page in the same way as "space", "enter" and "down" print just the next line)
Moved the code to print lines from the file to a function to reduce the space used by repeating it for the new keys

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/more.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/more.lua
@@ -13,7 +13,7 @@ local function nextLine(file, line, num)
     if not line then
       line = file:read("*l")
       if not line then -- eof
-        return 1
+        return true
       end
     end
     local wrapped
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks, will test tomorrow.

src/main/resources/assets/opencomputers/loot/OpenOS/bin/more.lua
```diff
@@ -5,6 +5,28 @@ local term = require("term")
...
+    end
+    local wrapped
+    wrapped, line = text.wrap(text.detab(line), w, w)
+    io.write(wrapped .. "\n")
+    i = i + 1
+  end
+  term.setCursor(1, h)
+  term.write(":")
+  term.setCursorBlink(true)
+  return 0
```
fnuecke: One question right away, though, the return value is used in a boolean check on line 48.
Are you aware that `not not 0 == true` in Lua? I.e. that zero is _not_ false as it is in, say, C?

jamesb6626:
Fixed this, thank you.
I didn't make a change to this that I made when testing it.
This should work now, I hope.

fnuecke:
Thanks :-)

PR #371, opened 2014-07-04
Fix some LuaJ bugs

gamax92:
LuaDouble:
Do not try Long.toString
Do not cast to float

LuaString:
Do not scan as a long
Add another check in scandouble (was scanlong)
Have substring create new bytes to work with


gamax92:
All patches have been tested in LuaJ 3 (On OpenComputers)

fnuecke:
Thanks!

PR #372, opened 2014-07-05
Adds support for the environment variable _

istasi:
http://istasi.dk/printenv.png yes, this clearly editted, due to my heighten paranoid level


fnuecke:
Hrm, wouldn't it be more correct to do something like

```
if #args > 0 then
  os.setenv("_", args[#args])
end
```

since

> it expands to the last argument to the previous command, after expansion

as [documented here](http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_03_02.html#sect_03_02_05)? At least that's also what the shell on my server behaves like, from what I can tell.

Update: nevermind, since env vars are all global in OC this is "wrong" anyway, so let's just stick with it holding the program last executed.

PR #391, opened 2014-07-11
Add some more functionality to string.format

gamax92:
The new checks in i,d and o,u,x,X probably aren't correct, but seem to work. (Wasn't sure how to translate that into LuaJ)


fnuecke:
Thanks!

PR #408, opened 2014-07-16
Changing shell.parse so it can handle abit more

istasi:
Making it possible for

./wierd -x1 -y 2 -z 4 -write "wooh" -l yah!
returns
{"yah!"},{z=4,y=2,x=1,write="wooh",l=true}

./wierd -write"wooh"
returns
{},{write="wooh"}

./wierd -write"ya -d"
returns
{},{write="ya -d"}

./wierd -write"asd
returns
{},{write"asd=true}

./wierd foo -write"hello" bar
returns
{"foo","bar"}, {write="hello"}

While i havn't super tested it with openOS, it can boot, mv -f respects it atleast, edit [file] works.
and while im not certainly sure why, it appears thats that shell.parse () gets a table thrown at it somewhere at boot?, while the original parse perverse it, i throw it away. OS seems fine without it, though if it is important, its easily added.

editted title

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/shell.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/shell.lua
@@ -163,33 +163,53 @@ function shell.execute(command, env, ...)
 end
 
 function shell.parse(...)
-  local params = table.pack(...)
-  local args = {}
+  local tmp = table.pack ( ... )
+  local args = ''
+
+	local arguments = {}
   local options = {}
-  local doneWithOptions = false
-  for i = 1, params.n do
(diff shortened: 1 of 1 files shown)


MyNameIsKodos:
Unless I'm completely oblivious (Which may very well be the case), are you trying to spell 'weird' as in strange or odd?  Because if so, you did it wrong =D  Unless I am indeed missing something which happens from time to time, ignore this comment.

istasi:
no i was attempting to spell weird and as you noted, failed hard.

istasi:
something is horrible wrong with this one, gonna take it off until i've gotten it fixed.

Except, i cant! :/, oh i can

PR #412, opened 2014-07-17
Update shell.lua

istasi:
same 'ish as #408

Needs abit of changes to sh.lua aswell, creating pull for that.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/shell.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/shell.lua
@@ -163,33 +163,53 @@ function shell.execute(command, env, ...)
 end
 
 function shell.parse(...)
-  local params = table.pack(...)
-  local args = {}
+  local tmp = table.pack ( ... )
+  local args = ''
+
+  local arguments = {}
   local options = {}
-  local doneWithOptions = false
-  for i = 1, params.n do
-    local param = params[i]
-    if not doneWithOptions and type(param) == "string" then
-      if param == "--" then
-        doneWithOptions = true -- stop processing options at `--`
-      elseif unicode.sub(param, 1, 2) == "--" then
-        if param:match("%-%-(.-)=") ~= nil then
-          options[param:match("%-%-(.-)=")] = param:match("=(.*)")
-        else
-          options[unicode.sub(param, 3)] = true
-        end
-      elseif unicode.sub(param, 1, 1) == "-" and param ~= "-" then
-        for j = 2, unicode.len(param) do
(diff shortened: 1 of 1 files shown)


istasi:
#413

istasi:
The other pull presumed that sh.lua gave exactly what was typed in after the file, like ./wierd -foo"bar" 
it presumed that i'd get -foo"bar" however text.tokenize removes all dem quotes and end up giving me -foobar instead.

Hi kodos, just for you, i kept my syntax of misspelling weird ;-)

Vexatos:
Shouldn't options consisting of more than one character be prefixed with two hyphens? I think that would make differentiation a little easier.

fnuecke:
@Vexatos that's how it is now and how I'd like to keep it, yes. What I like about this one is custom values for single-char options. Although that is... ambiguous, I suppose (consider `prog -n arg1 arg2`, how do we know `arg1` is an arg, not the value for `n`?)

The cleanest solution would probably be to support `-x=value` in addition to `--name=value`, which I believe is not currently possible? I.e. if it's a single one-char option also support value assignment.

rmellema:
Made make it possible to give a list of single char options that need an argument?
So `shell.parse(..., 'abc')` to make a, b and c single char options that take arguments?

Or made see how e.g. The python argument parser does this, that should be possible.

fnuecke:
Hmm, yeah, that sounds sensible, however the syntax would have to be different, of course (this way it'd be impossible to tell whether the 'abc' is an argument or the option names and 'abc' would have to come first or all but the first vararg would get dropped IIRC). Something like making an alternative call-mode with one sole argument that is a table should be safe enough? Would also be more future-proof. I.e. `shell.parse({args=table.pack(...), valueOpts="abc"})`.

The main danger here is breaking backwards compatibility, I think. It might even be best to add a second method entirely (shell.parseOptions or something like that?) leading to a change to `local args, options = shell.parse(...), shell.parseOptions('abc', ...)`. This would be compatible because the second returned value from `parse` would simply be ignored.

rmellema:
A different function might be a good idea, and a good example of named arguments for beginners. 

I/someone could also fix up and PR `getopt`. I can do this after my vacation. Would also make the system more POSIX :P

istasi:
http://pastebin.com/i3FpBg54 too crude?

t -f "foo bar" -l3 woh gets parsed into
{{"woh"}, {l=3,f="foo bar"}}

t -f "foo bar" -l 3 woh gets parsed into
{{"woh"}, {l=3,f="foo bar"}}

That way you can specify exactly how you want it look?, through the image argument.

And doesn't need to break sh.lua to do it

PR #413, opened 2014-07-17
Update sh.lua

istasi:
Keep dem quotes that text.tokenize so happy destroy!

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -104,7 +104,7 @@ local function execute(env, command, ...)
   if not program then
     return false, reason
   end
-  for i = 1, #args do
+  for i = 1, args.n do
     for _, arg in ipairs(evaluate(args[i])) do
       table.insert(eargs, arg)
     end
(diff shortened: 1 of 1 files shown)


istasi:
#412

fnuecke:
No, tokenize _should_ remove the quotes. They're syntax. To pass quotes, encase them in the other type of quotes, e.g. `text.tokenize([[this is a 'multi-word token "with" quotes']])`.

istasi:
grep "function tabIt" df.lua 

This kind of syntax is just not possible currently, because the program being callable does not receive the actually argument supplied but rather a parsed/stripped version of the same.

But why not give the arguments as supplied?

fnuecke:
What do you mean by 'not possible'? In that example `grep` would get two arguments, `function tabIt` and `df.lua`, which should be perfectly fine?

And passing the arguments 'as-is' (including quotes) will in _almost all cases_ force the programs to post-process the arguments themselves, stripping the quotes.

Lastly, this is like stuff is handled in C (at least on Linux), too:

```
pi@spiffy cat test.c
#include <stdio.h>
int main(int argc, const char** argv) {
  while (argc-- > 0)
    printf("%s\n", argv[argc]);
  return 0;
}
pi@spiffy ~ $ gcc test.c -o test
pi@spiffy ~ $ ./test "quoted argument" normal arguments
arguments
normal
quoted argument
./test
```

PR #414, opened 2014-07-17
Update components.lua

istasi:
http://istasi.dk/components.png

Adds -n [amount] -- Number n for number
Adds -l -- Makes it list all functions on the components its listing, yes this'll make it spam if you do not limit the number its showing. Also, for -n to work probably it needs the changes added by #413 and #412


Techokami:
ooooooo, this looks neat

fnuecke:
Thanks! Really useful :-)
I took the liberty of making some changes, in particular for the limit (it's now `--limit=n`), exactly because it won't work without the other PRs, which I still have to look through.

PR #416, opened 2014-07-18
Update df.lua

istasi:
http://istasi.dk/df.png

old versus new


fnuecke:
As it turned out that tabulate isn't really as versatile as we'd need it to be, could you revert the second commit?

istasi:
Im not home until Sunday late sadly, being without computer and shit :-( .. Will do then I suppose

fnuecke:
Sure, no hurry. Have a nice weekend!

istasi:
reversed

PR #417, opened 2014-07-18
Update ls.lua

istasi:
columns!, also doesn't break when max width adds up with resolution, thingy thingy

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/ls.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/ls.lua
@@ -30,8 +30,6 @@ for i = 1, #dirs do
     local lsf = {}
     local m = 1
     for f in list do
-      m = math.max(m,f:len())
-
       if f:sub(-1) == "/" then
         if options.p then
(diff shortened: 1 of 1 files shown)


istasi:
this is a rather ugly solution, sigh .... but works -.-'

fnuecke:
Hmm, some ideas for improving this:
- check if `gpu` is available, if it isn't just don't format.
- use the primary gpu (if available), not the first one in the list.
- use text.padRight.

For the first one, alternatively it might make sense to check if `io.output() == io.stdout`? E.g. to avoid formatting stuff when it's piped (in besh e.g., given someone/I find(s) the time to fix it...)

istasi:
Changed so newlines only are inserted if io.output() == io.stdout and component.gpu isn't nil, or enough files to reach math.huge

it still formats width between file entries based on max length though, regardless of whenever there's an gpu or not though...

Though in which case would ls ever be used where there's no gpu?, and in said case, why not just use filesystem.list directly?

fnuecke:
Looking over it now.

As for no GPU: servers where all IO is redirected through some pseudo-SSH thing, I guess?

PR #446, opened 2014-07-29
Update zh_TW.lang

mymagadsl:


fnuecke:
Thanks! I'm assuming this replaces #445 and that can be closed?

PR #466, opened 2014-08-04
Shell autocompletion

magik6k:
Added basic auto completion for shell that can recognize files and programs. Tapping tab key twice will print all the possibilities

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -207,12 +207,13 @@ local function hintHandler(line)
     if lastSearch == line then return matches end
   else
     local matches = getMatchingFiles(after)
-    if #matches == 1 then
-      lastSearch = ""
-      local ret = base .. space .. after .. matches[1]:gsub(after:match("[/]*(%w+)$"),"",1)
-      return ret:gsub("[^/]$","%1 ")
+    
+    for k in ipairs(matches)do
+      local ret = base .. space .. after .. (matches[k]):gsub(after:match("[/]*(%w+)$") or "","",1)
+      matches[k] = ret:gsub("[^/]$","%1 "):gsub("/$","")
     end
-    if lastSearch == line then return matches end
+    
+    return matches
   end
   lastSearch = line
 end
@@ -227,7 +228,8 @@ if #args == 0 and (io.input() == io.stdin or options.i) and not options.c then
       term.clear()
     end
     while term.isAvailable() do
-      local command = term.read(history, nil, hintHandler, drawPrompt)
+      drawPrompt()
+      local command = term.read(history, nil, hintHandler)
       if not command then
         term.write("exit\n")
         return -- eof
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -160,7 +160,7 @@ local function getMatchingPrograms(pattern)
   local res = {}
   for dir in string.gmatch(os.getenv("PATH"), "[a-zA-Z0-9/.]+") do
     for file in fs.list(dir) do
-      if string.match("/" .. file, "/" .. pattern) and file:match("(.+)[.]lua") then
+      if string.match(file, "^" .. pattern .. "(.+)[.]lua") then
         res[#res+1] = file:match("(.+).lua")
       end
     end
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -148,6 +148,76 @@ end
 local args, options = shell.parse(...)
 local history = {}
 
+local lastSearch
+
+local function drawPrompt()
+  local foreground = component.gpu.setForeground(0xFF0000)
+  term.write(expand(os.getenv("PS1") or "$ "))
+  component.gpu.setForeground(foreground)
+end
+
+local function getMatchingPrograms(pattern)
+  local res = {}
+  for dir in string.gmatch(os.getenv("PATH"), "[a-zA-Z0-9/.]+") do
+    for file in fs.list(dir) do
+      if string.match("/" .. file, "/" .. pattern) and file:match("(.+)[.]lua") then
+        res[#res+1] = file:match("(.+).lua")
+      end
+    end
+  end
+  return res
+end
+
+local function getMatchingFiles(pattern)
+  local res = {}
+  local dir = fs.isDirectory(pattern) and pattern or fs.path(pattern) or "/"
+  local name = (dir == pattern) and "" or fs.name(pattern) or ""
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
```diff
@@ -265,6 +277,8 @@ function term.read(history, dobreak)
       up()
     elseif code == keyboard.keys.down then
       down()
+    elseif code == keyboard.keys.tab then
```
fnuecke: Maybe better `elseif code == keyboard.keys.tab and not hint then`?

magik6k: Could be, but probably you meant `elseif code == keyboard.keys.tab and hint then`

src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
```diff
@@ -148,6 +148,76 @@ end
...
+  local foreground = component.gpu.setForeground(0xFF0000)
+  term.write(expand(os.getenv("PS1") or "$ "))
+  component.gpu.setForeground(foreground)
+end
+
+local function getMatchingPrograms(pattern)
+  local res = {}
+  for dir in string.gmatch(os.getenv("PATH"), "[a-zA-Z0-9/.]+") do
+    for file in fs.list(dir) do
+      if string.match("/" .. file, "/" .. pattern) and file:match("(.+)[.]lua") then
```
fnuecke: Hmm, why the `"/" ..`s? Or rather, why not `"^" .. pattern`? Won't this potentially match things that just _contain_ the current text (which I don't think it should)? Ditto for the `file:match`, should probably also start with a `^`, no?

src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
fnuecke: Gah, yes I did.

src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
magik6k: Using `"^"..pattern` performs better, second match does it's work well - checks for `*.lua`. There is way to make one match, as I didn't know the `"^"`, and it's `if string.match(file, "^" .. pattern .. "(.+)[.]lua")  then`

src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
```diff
@@ -148,6 +148,76 @@ end
...
+    if lastSearch == line then
+      term.write("\n")
+      for _, name in ipairs(matches) do term.write(name .. " ", true)end
+      term.write("\n")drawPrompt()
+    end
+  else
+    local matches = getMatchingFiles(after)
+    if #matches == 1 then
+      lastSearch = ""
+      return base .. space .. matches[1] .. " "
```
magik6k: This is broken, repairing it..

fnuecke:
This looks pretty cool, thanks :-)

I'd like to talk about a design decision, though, that I'm not sure about either way, myself: would it be better to have the 'preview' logic inside term itself? `hint` would then just return a table of 'completions'.
- Pro: more context, allows smarter behavior (e.g. caching results from `hint` on repetitive tab presses).
- Pro: centralized, less redundant code (people would just copy from sh.lua anyway).
- Con: must not print across multiple lines, so something like cycling through alternatives would have to be implemented instead (other ideas?).
- Con: less flexible.

I am currently slightly favoring putting it into term.read.

magik6k:
Well, the logic is hard to make useful in every use-case, for instance when you want to complete a file, when completion is a directory, it has `/` appended, else it gets a space. In fact `hint` should be able to return both string for completion and table (and/or iterator) for list of matching completions - and in case table contains only 1 element, it should also get printed, cause it would be hard to put it well into already written string. Let's say that user has written `/bin/som`, shall the completion be `/bin/somsomefile.txt` or `somefile.txt`? I guess neither of those are good. But still it's a good idea.

magik6k:
Okay, `hint` may now return tables, also I added `prompt` parameter which may be function drawing prompt or just prompt string.

fnuecke:
Uhm, from what I see (sorry, haven't had the time to actually run it), this now prints to multiple lines in `term.read()`? I don't really think that's a good idea, since it vastly limits the use of tables as hints. It's what I meant in my earlier reply - instead I think it'd be better to just 'cycle' through the alternatives (i.e. say you have `/b`, tab -> `/bin`, tab -> `/boot`, repeat). That would also make the prompt parameter superfluous.

magik6k:
I haven't thought of that, and it seems good, as shell creator still may print hint list inside . I'll code that and push later today.

magik6k:
Okay, it's now working in way you told. If you want test it grab some OC with os on hdd+internet card and do:
`pastebin run KaATrNkb`
`mpt ppa add m6koc`
`rm /lib/term.lua`
`mpt -Sy mesh`
`reboot`
`mesh`

src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
```diff
@@ -109,6 +109,9 @@ function term.read(history, dobreak)
   local scrollX, scrollY = 0, #history - 1
   local cursorX = 1
 
+  local hintCache = (type(hint)=="table" and #hint > 1)and hint
```
fnuecke, line 112: Sorry if I was unclear about that - I never really meant `hint` itself should be a table, I think it's fine to always have that as a function. I just meant it could _return_ a table that could then be cached. I think that'd simplify the code a bit, no?

I promise I'll find time to actually test it out tomorrow, busy day today.

magik6k, line 112: I understood what you mentioned above, but I thought that it may be simple addition, that can be used by someone creating simpler programs using static set of commands. It's just a simple feature.

fnuecke, line 112: Ah, all right then. I'll clean this up a little, there are some things that seem odd to me (e.g. the use of `after` in `tab()` which isn't defined anywhere).

I'll also try to adapt it to take into account the cursor position (e.g. if the text is `one[] two`, with `[]` being the cursor, it tries to complete `one`, not `two`).

src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
```diff
@@ -247,29 +250,61 @@ function term.read(history, dobreak)
     right(unicode.len(value))
   end
 
+  local function tab()
+    if not hintCache then
+      if type(hint) == "function" then
+        local h = hint(line())
+        if type(h) == "string" then
+          local _, cby = getCursor()
+          history[cby] = after
```
magik6k, line 259: Ouch, this should be `history[cby] = h`

fnuecke:
Thanks again! Now to find some people to break it... :-P

PR #486, opened 2014-08-13
transfer hologramMaxScaleByTier to the config file

xedp3x:


src/main/scala/li/cil/oc/Settings.scala
```diff
@@ -239,7 +239,10 @@ object Settings {
   val scriptPath = "/assets/" + resourceDomain + "/lua/"
   val screenResolutionsByTier = Array((50, 16), (80, 25), (160, 50))
   val screenDepthsByTier = Array(ColorDepth.OneBit, ColorDepth.FourBit, ColorDepth.EightBit)
-  val hologramMaxScaleByTier = Array(3, 4)
+  val hologramMaxScaleByTier = Array(
+    (config.getDouble("client.hologramMaxScaleTier1") max 20 min 1),
```
fnuecke: Hmm, please move that to the `Settings` class, then, and adjust usages to `Settings.get.hologramMaxScaleByTier`.

Also I think the min/max are the wrong way around, no?
`(x max 20 = 20+) min 1 = 1`, unless I'm too tired to think right now.

src/main/resources/application.conf
```diff
@@ -82,6 +82,13 @@ opencomputers {
       # may want to lower this a bit, to avoid it flickering too much.
       hologramFlickerFrequency: 0.025
 
+      # Maximum scale of the Hologram
+      # Normal size is 3x2x3
+      # 3 means 9x6x9 and 4 means 12x8x12
+      # min 1 max 20. Warning: bigger then 4 cane maks problems.
+      hologramMaxScaleTier1: 3
```
fnuecke: I think this would be nicer as an array (i.e. `hologramMaxScaleByTier: [3, 4]`).
Have a look at how the RAM or HDD size settings are implemented.

fnuecke:
Looks good, thanks!

PR #490, opened 2014-08-14
Add loot disk "parallel"

Akuukis:
The code works, its ok to call it beta.
I created man entries only for few functions, please review if I have started it correctly.
Comments on code, style, bugs and features are welcome.


Techokami:
I think the word you're looking for is parallel 
Might want to fix that ;)

fnuecke:
Interesting. I'll have a closer look over the weekend.

@Kilobyte22 since you're into the whole scheduling thing more than I am, mind also having a look at this?

@Akuukis please squash the commits. Also, I noticed a discrepancy in the upper/lowercase in the `parallel` manpage (e.g. in the description part kill and spawn start uppercase), might want to adjust that?

Akuukis:
@Techokami , @fnuecke thanks, done that.
Code looks working for me now, and you can check parallel.demonstration file for a tour for yourself. I have finished all documentation I could think of and sorry for my English over there. Let me know if there is anything else I have to do.

magik6k:
Changing parts of code in term lib seems to be really bad idea. IMHO you should make your own init.lua, and overwrite computer.pullSignal. Making that changes in term won't work well, as other tasks would be only performed when user is inside term.read. I think, that you should explore the way OpenOS works a bit more before making this kind of stuff, as it may be done much better way.

src/main/resources/assets/opencomputers/loot/parallel/parallel.lua
```diff
@@ -0,0 +1,489 @@
...
+      and (threads[focus].alarm or 0) < alarm 
+    then alarm = threads[focus].alarm or 0 end
+        
+    --logger.spam("\n")
+    if focus < lastThread then 
+      focus = focus + 1
+    else
+      focus = 1
+      event = require("event")
+      event = table.pack(event.pull(math.max(0,alarm - utils.getTime())))
```
magik6k: Ouch. You should require event lib on top of the program, and it shouldn't be overwritten each time it's used. IMHO event library shouldn't be even used for this purpose.

src/main/resources/assets/opencomputers/loot/parallel/parallel.lua
```diff
@@ -0,0 +1,489 @@
...
+  return coroutine.yield({
+    ["flag"] = "parallel_exit",
+  })
+end
+
+-- Magic happens here
+function parallel.manager( firstThread )
+
+  -- You should provide better versions of these functions.
+  local getUid = getUid or function () return tostring(math.random(10000000,99999999)) end
```
magik6k: In 'normal' systems Uid is just integer that get's increased every time thread is spawned, so processes have IDs like 1,2,3,10. Those are short, clean and easy to remember.

src/main/resources/assets/opencomputers/loot/parallel/parallel.lua
```diff
@@ -0,0 +1,489 @@
...
+    }
+  end
+  if not firstThread then 
+    firstThread = {
+      ["name"] = "Default console",
+      ["uid"] = getUid("coroutine"),
+      ["co"] = coroutine.create(
+        function ()
+          local term = require("term")
+          if not parallel then error("Load library into global table 'parallel' for default console to work") end
```
magik6k: Loading anything into global table is bad idea

src/main/resources/assets/opencomputers/loot/parallel/parallel.lua
```diff
@@ -0,0 +1,489 @@
...
+          while true do
+            term.write("\n@ ")
+            str = term.read(nil,nil,nil,true) -- yields here!
+            local fn, err = load(str)
+            if fn and str ~="" and str ~="\n" then 
+              print("executed:",pcall(fn))
+            else
+              if err then print("load error:", err) end
+            end
+          end
```
magik6k, line 265: This seems to be an attempt to create a basic lua console. It would be nice for it to have separated envirnoment with metatable to _G

src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
```diff
@@ -369,8 +369,13 @@ function term.read(history, dobreak, hint)
   term.setCursorBlink(true)
   while term.isAvailable() do
     local ocx, ocy = getCursor()
-    local ok, name, address, charOrValue, code = pcall(event.pull)
-    if not ok then
+    local ok, name, address, charOrValue, code
+    if not yieldable then
+      ok, name, address, charOrValue, code = pcall(event.pull)
+    else
+      name, address, charOrValue, code = coroutine.yield("[kc][el][yi][_p][db][oo][wa][nr]d?")
```
magik6k: I have no idea what does this may/should do, but I bet that it is meant to tell the system that it is call from.. system. And putting thing like that into OpenOS is bad. You should override coroutine in user envirnoment(note that this doesn't mean _G). the custom yield should somehow tell that the yield was caused by user. IIRC that kind of stuff was done in OC kernel.lua

src/main/resources/assets/opencomputers/loot/Parallel/usr/man/parallel.demonstration
```diff
@@ -0,0 +1,71 @@
...
+  4. Run your file at step #2
+  5. Test the commands below. Enjoy!
+  
+-- parallel demonstration functions, copy/paste lines in console
+-- expect very bad user interface, broken lines, etc.
+print("kuku!") -- expect it working
+print("kuku!""") -- expect error at loading
+print(nill[index]) -- expect error at executing
+a, err = parallel.spawn(function () print("kuku!") end) -- expect it working
+b, err = parallel.spawn(function () print("kuku!""") end) -- expect error
```
magik6k: It is normal, that it won't load, as it's syntax error, and it has nothing to do with the library

Akuukis:
@magik6k thanks for comments, all of them are implemented except two. Where you said about loading in globat table, imo I don't load anything into global table. 

``` lua
  local utils = utils or false
  local logger = logger or false
```

On lines 168 & 174 on fail local variables are initiated with value "false", therefore in next lines I make tables into those local variables. Does it work like that?

Secondly you said to make my own init.lua. Mainly I just wanted to complete and make available my library to others to use, and imo it is already compatible with OpenOS. But on other hand, there will be other libraries (pathfinding is next) coming out and I believe they should be put together in a separate OS. Nevertheless, today I tried few hours to understand how to make bare-bone init.lua without success, so help on that part would be very appreciated.

magik6k:
Well, I'm actually creating multithreaded init.lua, and it is going to be way more advanced, but this will be probably next OS rather than just parallel library. IMHO your lib is good, but I think it should be implemented in way the ComputerCraft library is done, as it's good for basic use. Also I think I'd put it to OpenPrograms repo and made it accessible through OPPM/MPT, as finding the disk in dungeon for normal player may be useless(I tried to find a nicer word + that's just my opinion), because he/she would have to be able to code something that depends on this library to actually make use of it. To get the repo visit irc.esper.net #oc and ask about it

fnuecke:
Re OPPM rather than loot disk: IMHO the main reason for loot disks is really just for use in environments where internet access is disabled and / or to "promote" them (since there are fewer than programs on OPPM). I don't see why you can't do both (so yes, I'd also suggest additionally putting it into OPPM).

But yeah, this does look good. Squash the commits and I'll merge it :-)

Vexatos:
I agree with @magik6k. Libraries like this shouldn't be on a loot disk, in my opinion, it might be too advanced for most players and thus useless to them.

Akuukis:
Thanks for comments. IMO lets don't put this to loot disk because for a player with no Lua knowledge it will be not useful. I will contact Vexatos and put this on OPPM.
Comments and suggestions about library are welcome at https://github.com/Akuukis/RobotColorWars

PR #499, opened 2014-08-17
Add edit features: Find, Delete-Line

WazWaz:
Adds incremental Find command (CTRL-F), and CTRL-Del to delete a line.

While I think most of us cheat and edit externally, it's useful when making small changes to be able to use the built-in edit command. But without Find, just getting to the point to make the small change is slow. Still needs Copy & Paste of course, but this change adds Highlight as a concept too which might help a little.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/edit.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/edit.lua
@@ -7,6 +7,8 @@ local term = require("term")
 local text = require("text")
 local unicode = require("unicode")
 
+local helpStatusText = "Save: [Ctrl+S] Close: [Ctrl+W] Find: [Ctrl+F]"
+
 if not term.isAvailable() then
   return
 end
@@ -118,6 +120,35 @@ local function setCursor(nbx, nby)
   component.gpu.set(w - 9, h + 1, text.padLeft(string.format("%d,%d", nby, nbx), 10))
 end
 
+local function highlight(x,y,n,yes)
+  local w, h = getSize()
+
+  local cy = y - scrollY
+  if cy > h or cy < 1 then
+    return
+  end
+
+  local cx = x - scrollX
+  if cx > w or cx < 1 then
+    return
+  end
+
+  local fg = component.gpu.getForeground()
+  local bg = component.gpu.getBackground()
+  if yes then
+    component.gpu.setForeground(bg)
+    component.gpu.setBackground(fg)
+  end
+  for c = cx,cx+n do
+    local ch,_ = component.gpu.get(c,cy)
+    component.gpu.set(c,cy,ch)
+  end
+  if yes then
+    component.gpu.setForeground(fg)
+    component.gpu.setBackground(bg)
+  end
+end
+
 local function home()
   local cbx, cby = getCursor()
   setCursor(1, cby)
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/bin/edit.lua
```diff
@@ -237,11 +276,70 @@ local function enter()
...
+end
+local function find()
+  findText = ""
+  refind()
 end
 
 local controlKeyCombos = {[keyboard.keys.s]=true,[keyboard.keys.w]=true,
-                          [keyboard.keys.c]=true,[keyboard.keys.x]=true}
+                          [keyboard.keys.c]=true,[keyboard.keys.x]=true,
+		  	  [keyboard.keys.f]=true,[keyboard.keys.g]=true}
```
fnuecke, line 342: Indent using spaces, please.

fnuecke:
This looks great, and I've given up on having OpenOS work on the lowest tier RAMs anyway (leave that to custom OSes :-P). I'll give it a go in a bit!

Edit: tho I'd vote to default to Shift+Del, I think that's more common (at least it's what I'm used to)?

WazWaz:
Shift-Del would be fine too. I just followed the "CTRL is for commands" model. If we had selection for cut-and-paste it wouldn't probably even be necessary (HOME, CTRL-downarrow, Del).

I could write an "ed"-like editor for lowest-end computers. Doesn't even need to hold the whole file in memory at once. I'm old enough to have used real teleprinters and quite enjoyed using ed.

Sorry about the spaces. I thought I'd configured vim correctly for doing that in lua. I'm not sure if you want me to fix and resubmit or what (I'm pretty new to github and pull requests).

rmellema:
I wouldn't go with shift+del, because not everybody(including me) has the delete key. You could do both, that would work as well.

Techokami:
> I wouldn't go with shift+del, because not everybody(including me) has the delete key. You could do both, that would work as well. 

You have a defective keyboard. Better replace it. (If it's a laptop keyboard, the FN key + Backspace should be mapped to Delete, so Shift + FN + Backspace)

As for the use of Shift versus the use of CTRL, we SHOULD be using CTRL, as that's what it was used for. Shift is the key to shift from capital letters to lowercase letters (and vice versa), CTRL is used for ConTRoL commands.

As for the creation of a clone of ed for barebones Tier 1 computers, I am very much down with that. That would be quite useful!

> I'm old enough to have used real teleprinters and quite enjoyed using ed.

aaaaa I romanticize about the olden days of computing, please tell me everything about everything

fnuecke:
I'll look into making the key combos configurable anyway - but for the heck of it, just some examples for programs I use that use Shift+Del: Sublime Text, IntelliJ IDEA. Are there ones that use Ctrl+Del? My argument here is pretty much "what's commonly in use", as was with ctrl+s and ctrl+w (and ctrl+f fits this, too).

Edit: also, the more text editor alternatives, the better! (someone please make a vi[m]-alike - please? :P)

Vexatos:
OC vim? Like vim++? :D :trollface:

rmellema:
Ah, I forgot about Fn + backspace, been mostly using x of c$ to delete things most of the time nowadays. 

Also, from that command, you should be able to infer my opinion about a vi[m]-alike editor :P

Techokami:
Am I the only person here who likes nano ._.

ghost:
@Techokami I have to add an alias to edit because I keep typing nano out of habit while messing around with OpenOS xD

fnuecke:
Thanks again. I just made a few minor adjustments and added configurable keybindings (saved in `/etc/edit.cfg` if possible).

PR #515, opened 2014-08-26
Update robot.names

jecowa:
suggesting some more robot names

### DIFF
--- a/src/main/resources/assets/opencomputers/robot.names
+++ b/src/main/resources/assets/opencomputers/robot.names
@@ -18,31 +18,83 @@ Kilobyte
 # Names of more or less famous robots, as a bit of filler material. Feel free
 # to add more via pull requests. Let's hope this won't get us sued...
 Adventure Core
+# video game "Portal 2"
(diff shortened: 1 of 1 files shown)


Vexatos:
Robbie AND Robby? I don't think that's legit.

fnuecke:
While I do appreciate the "source" annotations, that makes the file incredibly cluttered. I think I'll adjust the parser to allow for inline comments (e.g. `AdventureCore # Portal2`). That way it'll be somewhat cleanly and can still be kept sorted alphabetically (which reduces the chance of duplicates sneaking in).

fnuecke:
All right, done. Thanks!

I also used the opportunity to clean up the file a bit. I agree with Vexatos that the two names are too similar to justify keeping both (if that's what he meant, anyway :P). It's for the same reason I 'unified' the two Terminators in my cleanup, for example - I think it's better to keep the list varied over just making it longer.

jecowa:
Those inline comments look really spiffy!

Issue #546, opened 2014-09-07
Feature request: Allow a projection offset for hologram projectors


ingie:
Situation:  I have a wonderful [thanks to the darkest of dark things : black canvas from Extra Utilities] holographic projection suite. 

However, there's a big hole in the floor where the half-block projector sits, and i can't hide this except with a carpet or multi-part cover on the block above, which ruins the room's aesthetic, and more importantly means that the floor-space isn't clear of collisions [i.e. a robot can't walk "through" a carpet like a player can ]
more than that, unless i project at y>=8 i can't see the pixels, as they're embedded in the floor :)

see attached screenshot - diagonal coming out of the floor is from 1,1,1 > 32,32,32 [note, 1st 7 pixels missing below the floor ] - horizontal lines at 1,y,1 -> 32,y,32 [ where y = 1>8 ]  visible across the projector

if i want to not have the projector visible/ something hiding it in the block above, then my only option is to lower the projector by another block, but that means the offset above now needs to be 8+16 pixels... which leaves me only 8 pixels to actually project above the floor [sad face] 

Possible solution: allow a "projection offset" on the hologram api [in any axis perhaps, but most importantly Y] , perhaps at a relative power cost pro-rata the pixel offset up to a max limit of, say, 2 blocks from the projector [possibly less, but enough so i can at least hide the projector 1 block below the floor, and put a block on it to hide it completely from view. :)

![2014-09-07_19 59 20](https://cloud.githubusercontent.com/assets/1855688/4179334/8dc39fc2-36c1-11e4-9b06-684fe0a30ead.png)

Kilobyte22:
Sounds like a reasonable idea, and should be doable. Do not count on me though, i am not into the rendering code that much.

ingie:
Definitely doable... i've just changed the relevant code myself - as far as the rendering goes... 

i just changed the Y translation in HologramRenderer.scala to 1.5 as a test - naturally that's not how it needs to be done if it's a hologram component API setting from lua, but i wanted to test the principle first. i haven't even set up a dev env yet, this was just edited in vi and rebuilt :) 

result:
![2014-09-08_08 08 34](https://cloud.githubusercontent.com/assets/1855688/4182045/4bd644c8-3727-11e4-88c2-f00ac34415d7.png)

[...]

PR #548, opened 2014-09-08
Added hologram projection offset

ingie:
Apologies if this is all wrong, forgive my first ever PR :) 

anyway, as described in issue #546 

hopefully this all follows your standards, i was a bit unsure of the var name length for the additional setting, but i thought i'd keep it wordy and clear, than short and possibly obscure.   

hope you approve,
cheers!
ingie.

### DIFF
--- a/src/main/scala/li/cil/oc/client/PacketHandler.scala
+++ b/src/main/scala/li/cil/oc/client/PacketHandler.scala
@@ -44,6 +44,7 @@ object PacketHandler extends CommonPacketHandler {
       case PacketType.HologramPowerChange => onHologramPowerChange(p)
       case PacketType.HologramScale => onHologramScale(p)
       case PacketType.HologramSet => onHologramSet(p)
+      case PacketType.HologramPositionOffsetY => onHologramPositionOffsetY(p)
       case PacketType.PetVisibility => onPetVisibility(p)
       case PacketType.PowerState => onPowerState(p)
       case PacketType.RedstoneState => onRedstoneState(p)
@@ -189,6 +190,13 @@ object PacketHandler extends CommonPacketHandler {
           }
         }
         t.dirty = true
+      case _ => // Invalid packet.
+    }
+
+  def onHologramPositionOffsetY(p: PacketParser) =
+    p.readTileEntity[Hologram]() match {
+      case Some(t) =>
+        t.projectionOffsetY = p.readDouble()
       case _ => // Invalid packet.
     }
 
--- a/src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
+++ b/src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
@@ -85,7 +85,7 @@ object HologramRenderer extends TileEntitySpecialRenderer with Callable[Int] wit
     }
 
     GL11.glScaled(1.001, 1.001, 1.001) // Avoid z-fighting with other blocks.
-    GL11.glTranslated(-1.5 * hologram.scale, 0, -1.5 * hologram.scale)
+    GL11.glTranslated(-1.5 * hologram.scale, hologram.projectionOffsetY, -1.5 * hologram.scale)
 
     // Do a bit of flickering, because that's what holograms do!
     if (Settings.get.hologramFlickerFrequency > 0 && random.nextDouble() < Settings.get.hologramFlickerFrequency) {
--- a/src/main/scala/li/cil/oc/common/PacketType.scala
+++ b/src/main/scala/li/cil/oc/common/PacketType.scala
@@ -16,6 +16,7 @@ object PacketType extends Enumeration {
   HologramPowerChange,
   HologramScale,
   HologramSet,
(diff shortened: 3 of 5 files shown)


ingie:
derp: i note that i left the default offset in common/TE/Hologram.scala at 1.5 which is cool for me, but not best for others... it should be zero of course.

fnuecke:
Looks pretty good! One thing though: you're not saving the offset - see `write/readFromNBT`, so when saving, quitting and loading again, the offset would be reset. And while you add that, also add the offset to the NBT tag in `write/readFromNBTForClient` to properly synchronize it to clients when the chunk with the hologram projector in it is first sent to them.

You can make changes to your RP by simply committing to the same branch again, btw ;-) (when you're done, please squash your commits into one, though - if you don't know how to, google it, you'll find tons of tutorials).

ingie:
ah... re: NBT: of course - i admit i got excited when it "worked" :)

i'll do that. 

as a separate thing, i'm also working on a 3d primitive library for plotting line vectors + rotation transformation... basically i want to rotate cubes in 3d hologram space - i've done this many years ago in flash, so i'm revising my actionscript and converting it to lua initially... my thinking is it'd be nice to be able to say hologram.line( vec1, vec2 )  / hologram.rotate( thetaX, thetaY, thetaZ ) / hologram.transform( vec ) - you get the idea :)

any thoughts on that?

ingie:
@fnuecke i think i've successfully pushed that NBT/zero the default change, and squashed it properly... i hope - it looks ok on my repo now, so i'm guessing it's all good. 

i'm still hacking this in vi, since i've still yet to set up my dev env properly :) 

but i've released the jar to my own server and it's saving the offset correctly and all that jazz... 

fun script to test with it: 
while true do 
  for i = 0,math.pi*2,math.pi/20 do hologram.setOffset( math.sin(i) + 1.5 ) os.sleep(0.125) end 
end 
... bouncy bouncy :)

ingie:
... i think i've found a slight display bug. i think it's a "feature" of what was there before in the culling - but it's difficult to see normally... basically, if one looks above the hologram to a certain amount, then the hologram isn't rendered.  

this is happening normally - i.e. before i changed it, but i can see that it only happens just on the cusp of not being able to see it anyway in the camera, so it's not noticable. 

on a hunch, i'm thinking i need to factor the offset into the sx/sy/sz culling equation/decision - but i'm not 100% sure of what that bit is specifically doing...

it's not a _bad_ bug, but it's a bug... so i'll look at that when i can... if you have any quick ideas on that, let me know [ not asking to be walked through it at all, just in case you recognise the bug enough to say whether i'm on a wild goose chase/ or on the mark with my own hunch ]

fnuecke:
Ah right, I wanted to mention that earlier but forgot while reading through the diff: you'll want to adjust the bounding box of the tile entity. That is offset the AABB returned from `getRenderBoundingBox` (and/or overscale slightly).

ingie:
gotcha. cheers. i'll do that tomorrow sometime... i'm going to play for a bit, now my holo-projection room is darker :) 

p.s. while i'm on - did you see my post on mcf about the OC website not sending me a confirmation, so i can't post there. i'm still not getting one [ and as said, my email addy and all are correct, and it's not being spammed, so ]

ingie:
... fixed that bug, thanks for that pointer - twas obvious once i saw what that was doing.
I'll still push that tomorrow tho... rather than be too eager again and find other things :)

fnuecke:
Cool. As for the forum account, I activated it manually (and just saw another PM on the MCF from someone with the same issue... will have to keep an eye on that).

fnuecke:
Oh, I only just read your post on the issue to this and realized this is for the 1.7 stable branch.
Please re-create this PR for the `master` branch, since that's the dev branch and my porting workflow is to do everything in 1.6 first, then pull into the 1.7 branch. Flipping that now would be massive headache I think. Thanks!

ingie:
will do... makes sense... the main reason i did it in there is that i couldn't build the gradle from the master due to missing dependencies... so i assumed it wasn't the real master, as it were... 

i'll have a look more into why that was... i just skipped the idea at the time because of that, and worked in a branch i could compile off the bat...

fnuecke:
Hmm, maybe it was just bad timing and some server hosting the dependencies was down (had some builds on Jenkins fail because of that, so annoying).

Wuerfel21:
why doesnt it use offline versions of the dependancy as a fallback anyway?

ingie:
@Wuerfel21 well, it would, but the gradle has to download them at least once before they can be offline :)

fnuecke:
That, too, although I assume he means the Jenkins, in which case the answer is: because I run a clean before each build since I had issues with lingering compiled classes causing problems in the past.

ingie:
righty... it _did_ download everything and build this time [ well, once i'd increased permgen for gradle, but that was a different matter ]

i'll get that back in the correct branch asap... i'm doing a music production session at home today so i doubt i'll get chance until tomorrow.

fnuecke:
Ah, cool. No hurry :)

PR #554, opened 2014-09-09
Update robot.names

LizzyTrickster:
Happy now @fnuecke


fnuecke:
I am! :+1:  See, that wasn't so hard ;-)

LizzyTrickster:
Well, i had to delete my copy of the repo but meh

fnuecke:
You shouldn't have had to >_> But whatever works for you.

PR #558, opened 2014-09-10
Added hologram projection offset

ingie:
Hi @fnuecke 

Changes now in master, as requested...

I've tested in 1.6.4 and all looks as it should :)


fnuecke:
All right, looks good, thanks!

PR #572, opened 2014-09-18
Update DebugCard.scala

natedogith1:
Fixed a few of the DebugCard's Callback docs


fnuecke:
Ahh, good old copy-paste. Thanks!

PR #580, opened 2014-09-28
ChunkLoader Upgrade now loads a 3x3 chunk area centered on the robot.

### LINKED ISSUE #579: Lighting issue with Chunk-Loader Upgrade
Minecraft refuses to (re)light a chunk if all eight of the immediately surrounding chunks are not loaded. (See World.updateLightByType() and World.doChunksNearChunkExist())

Since the Chunk-Loader Upgrade only loads a single chunk, any sort of block interaction the robot does, such as placing torches, will not trigger light updates. I propose that the Chunk-Loader Upgrade load a 3x3 area of chunks centered on the robot. This will probably help with any other unforeseen issues caused by unloaded chunk borders, as well.

This is relatively easy to test for. Simply load up a single person server, put a robot on a very low Y-level, and have him continuously move forward while digging forward/down, and placing a torch every 8 or so blocks with a Chunk-Loader Upgrade enabled. Log out for about 5-10 minutes, keep the server running, and log back in. If nothing else was keeping the chunks loaded, you should have lighting issues.

Nibato:
This should resolve #579

### DIFF
--- a/src/main/scala/li/cil/oc/common/event/ChunkloaderUpgradeHandler.scala
+++ b/src/main/scala/li/cil/oc/common/event/ChunkloaderUpgradeHandler.scala
@@ -59,15 +59,21 @@ object ChunkloaderUpgradeHandler extends LoadingCallback {
   }
 
   def updateLoadedChunk(loader: UpgradeChunkloader) {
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks!

PR #589, opened 2014-10-04
Update motd

Wuerfel21:
finall


fnuecke:
Hmm, I don't really think this makes much sense. The free number will fluctuate wildly, depending on when the GC last ran. Also, my real computer only shows me the total memory when booting, not the used memory :P

PR #602, opened 2014-10-17
Fixed doc in Tablet.scala

natedogith1:


fnuecke:
Ah, thanks! Copy-pasta strikes again :-P

PR #624, opened 2014-10-24
Added rc scripts system

rmellema:
Here is a small system that allows user to more easily manage and install daemons and libraries. There is also an example script that doesn't really do anything. To enable it at startup, at it to the `enabled` table.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/etc/rc.d/example.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/etc/rc.d/example.lua
@@ -1,0 +1,13 @@
+local count = 0
+
+function start(msg)
+  print("This script displays a welcome message and counts the number " ..
+  "of times it has been called. The welcome message can be set in the " ..
+  "config file /etc/rc.cfg")
+  print(args)
+  if msg then
+    print(msg)
+  end
+  print(count)
+  count = count +1
+end
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
@@ -1,0 +1,70 @@
+local fs = require('filesystem')
+
+local function loadConfig()
+  local env = {}
+  local fun, reason = loadfile('/etc/rc.cfg', 't', env)
+  if fun then
+    local res, reason = fun()
+    if true then
+      return env
+    else
+      return nil, reason
+    end
+  else
+    return nil, reason
+  end
+end
+
+local loaded = {}
+
+local function load(name, args)
+  if loaded[name] then
+    return loaded[name]
+  end
+  local fileName = fs.concat('/etc/rc.d/', name ..'.lua')
+  local env = setmetatable({args=args}, {__index = _G})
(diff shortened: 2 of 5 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
```diff
@@ -0,0 +1,78 @@
+local fs = require('filesystem')
+
+local function loadConfig()
+  if fs.exists('/etc/rc.cfg') then
```
fnuecke: The `file.exists` check isn't really necessary, is it? `loadfile` will return something like `nil, "file xyz not found"` if that were false, anyway, so this seems a bit redundant to me.

src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
```diff
@@ -0,0 +1,78 @@
...
+end
+
+local loaded = {}
+
+local function load(name, args)
+  if loaded[name] then
+    return loaded[name]
+  end
+  local fileName = fs.concat('/etc/rc.d/', name ..'.lua')
+  if fs.exists(fileName) then
```
fnuecke: Same as with above `file.exists`.

src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
```diff
@@ -0,0 +1,78 @@
...
+    return nil, reason
+  end
+  if env[cmd] then
+    return env[cmd](...)
+  else
+    return nil, "Command '" .. cmd .. "' not found in daemon '" .. name .. "'"
+  end
+end
+
+local function runCommand(cmd, name, ...)
```
fnuecke: Is there a particular reason for this argument order? I feel `name, cmd, ...` would be more intuitive, as it follows the common object -> property -> arguments order (e.g. as in `obj.fun(args)`).

src/main/resources/assets/opencomputers/loot/OpenOS/etc/rc.d/example.lua
```diff
@@ -0,0 +1,13 @@
+local count = 0
+
+function start(msg)
+  print("This script displays a welcome message and counts the number" ..
+  "of times it has been called. The welcome message can be set in the" ..
+  "config file /etc/rc.cfg")
+  print(args[1])
```
fnuecke: Isn't `args` just "Hello World!" in this case? I.e. is the `[1]` really correct here?

src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
rmellema: Yeah, that would make more sense, no real reason to have the comments in that order. Can't remember why I did it that way either.

src/main/resources/assets/opencomputers/loot/OpenOS/etc/rc.d/example.lua
rmellema: I guess it would even error in this case, will fix.

PR #642, opened 2014-11-01
Added Sangar and Notch

isoadam:
Sangar as creator
Notch as god


fnuecke:
While I could sort-of count as contributor, I don't need my name in that list; It'd feel a little too much like patting myself on the back if you know what I mean. I have a pet robot for that :-P But thanks for the thought ;-)

As for Notch, creating MC doesn't count as contributing, IMHO (we'd kind of also have to add the names of everyone who contributed to Forge, then). And I'm pretty sure he's not a robot. Also, after previous experience I'm not inclined to add people's names to the list without their consent.

PR #660, opened 2014-11-10
Fixed recipes for StargateTech2

magik6k:
The recipes were broken for new SGT2 builds causing the cards to disappear from game.

### DIFF
--- a/src/main/resources/assets/opencomputers/recipes/default.recipes
+++ b/src/main/resources/assets/opencomputers/recipes/default.recipes
@@ -84,7 +84,7 @@ openOS {
 }
 
 abstractBusCard {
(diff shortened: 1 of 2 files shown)


LordFokas:
Oh nice. Thanks.
Also, is the StackManager still being used anywhere? I'm considering dropping it and moving the whole system to FML's `getCustomItemStack()`, but if it's still being used I can leave it as a wrapper around FML.

fnuecke:
Awesome, thanks!

@LordFokas well, at least I'm not using it. The recipes in OC work purely on block/item IDs and (unlocalized) names.

LordFokas:
@fnuecke cool. That means I can get rid of it. Thanks.

Issue #536, opened 2014-09-04
Add switch upgrades


Wuerfel21:
These would be similar to robot upgrades. Some ideas:

Intelligence upgrade - make the switch reject dupe messages
Port filter upgrade - filters ports
One way upgrade - what it says on the tin
Advanced filter upgrade - filters addresses, ports and message sizes

Techokami:
What's wrong with using a computer to do these functions?

Wuerfel21:
The Source address is wrong then.
Also computers don't differentiate networks

fnuecke:
Hmm, well, some of this could be quite interesting, I guess. Though I'm not a fan of the extra upgrade types idea. It'd be better to have these unlock via the already existing "upgrade system" I think (e.g. needs T1 CPU for simple filtering, T2 for advanced etc).
This will need a lot of GUI work, which I hate very much :-P Also will need some drafting for nice GUI layouts to properly visualize all this... so yeah, relatively time consuming. Will have to see.

Wuerfel21:
why not expose some api and configure stuff from a computer?(of couse only affter entering a password)

fnuecke:
Expose them as components (as access points already are) maybe. Password? No.

Kilobyte22:
Switches are not intended as security devices, but only to seperate networks. If you want anything with security, you gotta make your own switch using a computer and two network cards. (Actually, i am not sure if that is possible currently)

fnuecke:
It's possible to do manual "routing" using servers (that's the reason the internal mode was added). But that seems somewhat... excessive, if all you want to do are really simple things, like one-directional thinger, so I didn't reject it immediately.

Kilobyte22:
indeed. But, if you want more, grab a server.

Kilobyte22:
> The Source address is wrong then.

Then write your own networking protocol or wait for OpenIP (or whatever i'll call it)

LizzyTrickster:
(This was originally intended to be posted after @Kilobyte22 second comment but was helping little brother with the wii) SKS managed to do something similar to manual routing using computers on my server. I think he had computers/switches in a configuration of: Computer <=> cable <=> switch <=> cable <=> computer (with the sequence repeated a few times). I think it worked by broadcasts or somehing

magik6k:
[...]

PR #676, opened 2014-11-16
Add network loot disk

magik6k:
I worked at simple networking API and it is finally done. It is/will impement features from #536.
Features for now:
- Automatic route discovery, routing
- Drivers for modem and tunnel(Linked cards)
- Multiple names/addresses for host(+ability to define own)
- TCP and UPD like protocol(for connecton and connectionless data sending)
- ifconfig utility that looks pretty much like the original one
- Simple netcat and ping
- API documented in man file(network)

To try it just put floppy and run installNetwork on computers that you want to connect. For now you have to rebot when you change network components of computer(reboot only one host to be clear)

I've also added --fromDir option to install so there is no need to implement --exclude which would be way more complex.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
+++ b/src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
@@ -31,7 +31,7 @@ function network.icmp.ping(addr, payload)
     return pingid
 end
 
-function internal.icmp.hanlde(origin, data)
+function internal.icmp.handle(origin, data)
     if data:sub(2,2) == "P" then
         local matcher = data:sub(3):gmatch("[^:]+")
         local compid = matcher()
@@ -70,7 +70,7 @@ function network.udp.send(addr, port, data)
     driver.send(addr, "D".. string.char(math.floor(port/256))..string.char(port%256)..data)
 end
 
-function internal.udp.hanlde(origin, data)
+function internal.udp.handle(origin, data)
     local port = data:byte(2)*256 + data:byte(3)
     if internal.udp.ports[port] then
         computer.pushSignal("datagram", origin, port, data:sub(4))
@@ -133,7 +133,7 @@ function network.tcp.send(channel, data)
     return false
 end
 
-function internal.tcp.hanlde(origin, data)
+function internal.tcp.handle(origin, data)
     if data:sub(2,2) == "O" then
         local port = data:byte(3)*256 + data:byte(4)
         local rchan = data:byte(5)*256 + data:byte(6)
@@ -195,9 +195,9 @@ end
 
 event.listen("network_message", function(_, origin, data)
     --print("NETMSG/",origin,data)
-    if data:sub(1,1) == "I" then internal.icmp.hanlde(origin, data)
-    elseif data:sub(1,1) == "T" then internal.tcp.hanlde(origin, data)
-    elseif data:sub(1,1) == "D" then internal.udp.hanlde(origin, data) end
+    if data:sub(1,1) == "I" then internal.icmp.handle(origin, data)
+    elseif data:sub(1,1) == "T" then internal.tcp.handle(origin, data)
+    elseif data:sub(1,1) == "D" then internal.udp.handle(origin, data) end
         
 end)
 
@@ -212,11 +212,3 @@ network.info.getInterfaceInfo = function(...)return driver.intstat(...) end
 ------------
 
 return network
-
-
-
-
-
-
-
-
--- a/src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
+++ b/src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
@@ -1,0 +1,222 @@
+local event = require "event"
+local computer = require "computer"
+
+local driver = {}
+
+local network = {}
+local internal = {}
+
+------------
+--Core communication
+network.core = {}
+
+function network.core.setCallback(name, fn)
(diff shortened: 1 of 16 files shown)


src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
```diff
@@ -0,0 +1,222 @@
...
+internal.icmp = {}
+
+local pingid = 0
+function network.icmp.ping(addr, payload)
+    pingid = pingid + 1
+    driver.send(addr, "IP"..computer.address()..":"..tostring(pingid)..":"..payload)
+    return pingid
+end
+
+function internal.icmp.hanlde(origin, data)
```
fnuecke: Even if it's internal, `hanlde` is a typo. Might want to fix that to avoid running into issues in the future when instead trying to reference it with correct spelling.

src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
```diff
@@ -0,0 +1,222 @@
...
+    internal.udp.checkPortRange(port)
+    internal.udp.ports[port] = nil
+end
+
+function network.udp.send(addr, port, data)
+    internal.udp.checkPortRange(port)
+    driver.send(addr, "D".. string.char(math.floor(port/256))..string.char(port%256)..data)
+end
+
+function internal.udp.hanlde(origin, data)
```
fnuecke: Ditto (`hanlde`).

src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
```diff
@@ -0,0 +1,222 @@
...
+
+function network.tcp.send(channel, data)
+    if internal.tcp.channels[channel] and internal.tcp.channels[channel].open then
+        driver.send(internal.tcp.channels[channel].addr, "TD".. string.char(math.floor(internal.tcp.channels[channel].remote/256))..string.char(internal.tcp.channels[channel].remote%256)..data)
+        return true
+    end
+    return false
+end
+
+function internal.tcp.hanlde(origin, data)
```
fnuecke: Ditto (`handle`).

src/main/resources/assets/opencomputers/loot/Network/data/lib/network.lua
```diff
@@ -0,0 +1,222 @@
...
+end
+
+------------
+--Data processing
+
+event.listen("network_message", function(_, origin, data)
+    --print("NETMSG/",origin,data)
+    if data:sub(1,1) == "I" then internal.icmp.hanlde(origin, data)
+    elseif data:sub(1,1) == "T" then internal.tcp.hanlde(origin, data)
+    elseif data:sub(1,1) == "D" then internal.udp.hanlde(origin, data) end
```
fnuecke: Ditto (`hanlde`).

fnuecke:
Cool, looks quite impressive! One question beside the internal typo: I'm assuming the install script isn't called `install` to avoid it calling itself recursively? Did you deliberately choose not to use `/bin/install` instead, and if so, why?

fnuecke:
Oh, and would you mind making the changes/fixes(?) to `install.lua` a separate PR for clarity? Also, I guess it should really be `return "/mnt/" .. origin .. fromDir` now (although you don't use it anyway, from what I can see)?

src/main/resources/assets/opencomputers/loot/Network/installNetwork.lua
```diff
@@ -0,0 +1,8 @@
+local filesystem = require "filesystem"
+local shell = require "shell"
+local process = require "process"
+print("installNetwork")
+
+local status, installDir = shell.execute("install --noboot --nolabelset --name=Network --fromDir=/data/ --from="..filesystem.get(process.running()).address)
```
magik6k, line 6: ^fromDir here

fnuecke, line 6: Well, yes. You assign it to `installDir`, but you don't use `installDir`, do you?

fnuecke, line 6: To clarify, I don't mean the parameter is unused, I'm thinking the return value is unused.

magik6k, line 6: True, it's there because earlier(before I added fromDir) I used this parameter to later remove files that shouldn't be copied(i.e. the script), but after adding autorun I had to do it in some better way -> fromDir

magik6k:
Typos fixed; The install script isn't called install as it wouldn't be invoked when user was running 'install' as it is at end of path. Also it's IMO more intuitive. 

---

Edit
Also I didn't want to make another PR with the install as this change is needed for this PR to be installable(i.e. this PR depends on install change).

fnuecke:
Oh, one would have to call it via `./install` then, I guess? Allright then.

magik6k:
yep, and thus writing installNetwork seems to be better option(even better with tab completion)

src/main/resources/assets/opencomputers/loot/Network/installNetwork.lua
fnuecke, line 6: Right. I'll probably remove the return then, though.

fnuecke:
Will do some quick testing later this evening, and probably make some minute changes (such as it getting installed to `/usr/` instead of `/`), then push the merge.

PR #678, opened 2014-11-18
Update robot.names

MyNameIsKodos:
Added Herpaderpicus because it sounded funny.


fnuecke:
Eh, I think there are enough names so we can stick to actual robot/AI names without needing to pad the list with "fun" names ;-)

PR #720, opened 2014-12-09
Changed origin of names

scj643:
Changed names that just said Portal but were really only in Portal 2.
Added Wolfram to the list of names.

### DIFF
--- a/src/main/resources/assets/opencomputers/robot.names
+++ b/src/main/resources/assets/opencomputers/robot.names
@@ -79,3 +79,4 @@ Watson             # IBM Watson
 Weebo              # Flubber
 Wheatley           # Portal 2
 Wobbo              # Contributor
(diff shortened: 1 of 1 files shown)


Kubuxu:
Wolfram's algorithm for integration. Probably the most commonly used nowadays.

fnuecke:
Thanks! But ;-)

Portal: think of it as the franchise. This was in fact a deliberate choice, it's also the case for a few other entries (Dog, e.g.).

I'm not sure Wolfram qualifies as an actual AI, at least in the vein of other fictional AIs, and quite definitely not as a robot. As mentioned in a previous PR, I think the list is long enough to allow some stricter requirements now - while robots are pretty self-explanatory, let's say AIs should at least have _perceived_ self-awareness. I'll update the readme accordingly, later.

PR #729, opened 2014-12-11
Updated Processor Texture Tiers

darkstorm2150:
Updated Processor Textures 1st Revision


magik6k:
Nah. Colors of processors are meant to indicate their tier, same goes to other items in the mod. Just darker textures are unreadable, and IMHO not so nice.

fnuecke:
I prefer the existing textures, too. Also, for future reference, just changing the PSD isn't sufficient, you'd also need to update the actual PNGs in the assets/opencomputers/textures/... directory.

darkstorm2150:
I completely understand

PR #755, opened 2014-12-23
Add readNumber()/file:read("*n")

SoniEx2:


fnuecke:
Github needs to send periodic reminders about pending PRs >_>

PR #777, opened 2015-01-02
Allow changing the color of robots' lights to an arbitrary color via API...

Hotrian:
... (`component.robot.setLightColor`/`component.robot.getLightColor`).

Designed to match c4227d2.

Changes that allow Robots to change the color of their animated band.
Exposes the `component.robot.setLightColor(value:number)`
and `component.robot.getLightColor()` functions.
All Robots default to a red color that is close (hopefully exact) to the previous red.
Uses the same RGB int value system as the drones.

robotcolorexample

Edit: Updated description


Hotrian:
There's a bug in this that makes the game crash if you are one of the 3 people ( not me :( ) that have a thingy, but I'm fixing that right now.

src/main/scala/li/cil/oc/client/renderer/tileentity/RobotRenderer.scala
```diff
@@ -214,6 +214,15 @@ object RobotRenderer extends TileEntitySpecialRenderer {
           RenderState.disableLighting()
         }
 
+        // Additive blending for the light.
+        GL11.glBlendFunc(GL11.GL_SRC_ALPHA, GL11.GL_ONE)
+        // Light color.
+        var lightColor = if (robot == null) 15872048 else if (robot.info == null) 15872048 else robot.info.lightColor
```
fnuecke: Please use hex values for the colors, makes it much more readable (`0xF23030` in this case, which immediately makes it apparent that it's a red-ish tone).

Also, as Kubuxu said, simplify it to

``` scala
if (robot != null && robot.info != null) robot.info.lightColor else 0xF23030
```

src/main/scala/li/cil/oc/util/ItemUtils.scala
```diff
@@ -225,6 +225,8 @@ object ItemUtils {
 
     var containers = Array.empty[ItemStack]
 
+    var lightColor = 15872048
```
fnuecke: Again, hex value please.

fnuecke:
Looks good! Change the two things noted above and squash the commits, then I'll test and merge it.

Hotrian:
Okay, will do in a few minutes. Sorry, first time contributing on GitHub. Also, that number is the RGB integer value for the default red which can be calculated the same way as for the drones, but I will switch to hex values as mentioned.

fnuecke:
No worries. And yes, it is the same value, but representing it using hexadecimal notation makes it human readable (well, to those familiar with the notation anyway, which is at least everyone who ever dabbled with CSS/HTML colors :P)

Hotrian:
Okay! Sorry, about the delay. Had to run out for a bit, and I had never actually rebased before so that took a minute. The Travis CI build is failing because the CDN for TMechworks went down? Or they removed that version. I tested it before pushing of course and it built fine so it should be good to go.

Kubuxu:
It is because servers are down. Not your fault.

fnuecke:
Yeah, should build fine locally and on my Jenkins where those dependencies are cached.
Will have a look in a bit.

fnuecke:
All seems fine, thanks!

PR #799, opened 2015-01-10
Minor bug in flash.lua

### LINKED ISSUE #796: Drones not turning on
I followed the "primitive" remote control tutorial.
Did exactly what You did.

Drones just ... won't turn on (charged fully, BIOS flashed, all components with wireless card too, tier1/tier2 tested).

Please fix ASAP.

Hotrian:
Fixed eeprom reference in flash.lua

We asked the users to swap the eeprom out here. If they did, the reference to eeprom is invalid and must be redefined. Using `component.eeprom` directly would work, but the following lines reference `eeprom` which, as requested, should have been swapped, and now references an invalid eeprom, causing an error.

This closes #796


fnuecke:
Hrm, I think I actually prefer pulling the actual variable declaration into the individual methods. I'll do that real quick. Thanks anyway!

PR #800, opened 2015-01-10
Fix issues with locale/tonumber() on some machines

SoniEx2:
This fixes tonumber("1.2") not working but tonumber("1,2") working.

### DIFF
--- a/src/main/resources/assets/opencomputers/lua/machine.lua
+++ b/src/main/resources/assets/opencomputers/lua/machine.lua
@@ -1,3 +1,4 @@
+os.setlocale("C")
 local hookInterval = 100
 local deadline = math.huge
 local hitDeadline = false
--- a/src/main/scala/li/cil/oc/util/LuaStateFactory.scala
+++ b/src/main/scala/li/cil/oc/util/LuaStateFactory.scala
@@ -229,8 +229,7 @@ object LuaStateFactory {
         state.pop(8)
 
         // Prepare table for os stuff.
-        state.newTable()
-        state.setGlobal("os")
+        state.openLib(jnlua.LuaState.Library.OS)
 
         // Kill compat entries.
         state.pushNil()


fnuecke:
Uh, did you test this? Because I'm pretty sure that method isn't exposed, so that'd error with a `trying to call a nil value`.

SoniEx2:
uhh it happens outside the sandbox? (but no, I didn't test it)

fnuecke:
Outside the inner sandbox ;-) [No I'm totally not paranoid or anything]

As discussed on IRC, you'll need to load the OS lib [here](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/scala/li/cil/oc/util/LuaStateFactory.scala#L221), which it currently isn't (same for the IO lib, because I wanted to make sure there's no way to access that even if the inner sandbox got compromised).

However, this will need either some serious testing, to make sure this change in locale doesn't change Java's behavior somewhere, or some documentation telling me that it won't (I'd imagine this might be documented in JNI somewhere).

SoniEx2:
It might change stacktraces... I'm not sure how Java does stacktrace localization stuff...

fnuecke:
I did a bit of research, and there's something I'd like to have cleared up: from what I've read C's `setlocale` (which is what this boils down to) isn't necessarily thread-safe, so it'd be best to do this once on initialization, not for every state. Could you please look into this? Thanks.

Providing some test results of whether and how this may affect other parts of Java would also make me a lot more likely to merge this. I do _not_ want to risk messing anything up somewhere, because that'd be hell on earth to track down for those affected.

Also, this call should be made from within the factory itself, after which the `os` table should be replaced with an empty one. I don't want it to be exposed to any scripts.

SoniEx2:
Why can't you expose the full Lua and just sandbox things from Lua? makes it much easier to use a custom Lua (doesn't have to rebuild OC to add custom libs)

fnuecke:
I won't because it is more secure this way. If you want a custom Lua without writing your own libs, mod a custom architecture and link against OC itself reusing the internal classes.

natedogith1:
How much would you be against a modification to OC's custom version of lua (the c code) to make the tonumber (and probably tostring) functions locale independent?

Actually, modifying the c code might not be the best; so which would be a better option, a config option with a scala-side os.setlocale, or a lua-based woraround?

fnuecke:
I'm strongly favoring a call to `os.setlocale` if anything, because I only want to recompile the natives if I really have to.

PR #806, opened 2015-01-12
Fix build.gradle for Gradle 2.x, copy over assets automatically

AtomSponge:
`idea.module.scopes.PROVIDED.plus += configurations.provided` is no longer valid since Groovy 1.8.9, the "object" you want to add has to be a collection too.
Also, let Gradle automatically copy the assets (from build/resources/main to build/classes/main) so we don't have to do that manually anymore.


fnuecke:
Thanks! No more manual copying \o/

PR #809, opened 2015-01-14
Add BloodMagic integration

AtomSponge:
Add basic BloodMagic integration for Blood Altars, Master Ritual Stones and Blood Orbs.


fnuecke:
Thanks!

PR #817, opened 2015-01-15
Added match check to term.read

magik6k:
As simple as possible, yet it helps to keep many programs a bit shorter


fnuecke:
Thanks!

PR #832, opened 2015-01-18
Fixed filesystem.list returning a list with duplicate values

mpmxyz:
The error was caused by increasing the index i after removing its value: The value after the removed one was ignored.
To see an effect you need two directories being neighbors in the list.
Both also have to have a symbolic link or a mounting point inside.
-> That's a rare but possible combination.

Here an example:
-bugdir/
--a/
---link1 -> link somewhere
--b/
---link2 -> link somewhere
# ls bugdir

a  b  b (!)


fnuecke:
Thanks!

PR #848, opened 2015-01-24
Fix InventoryUtils.insertIntoInventory

DirtYiCE:
According to the documentation `inventory_controller.suckFromSlot`, `robot.suck` and friends try to put the sucked item into the selected slot, and only if it's occupied try others. However the implementation first tried to merge the stack with existing stacks in the inventory, and only if it failed put into the selected slot. This commit fixes it.

But only of course if the documented is the intended behaviour. (But I hope so, because otherwise using `robot.suck` is a mess since you have no idea what you just picked up at where it ended, unless you have a completely empty inventory...)

### DIFF
--- a/src/main/scala/li/cil/oc/util/InventoryUtils.scala
+++ b/src/main/scala/li/cil/oc/util/InventoryUtils.scala
@@ -155,6 +155,12 @@ object InventoryUtils {
       var remaining = limit
       val range = slots.getOrElse(0 until inventory.getSizeInventory)
 
+      val stackSize = stack.stackSize
+      if ((inventory.getStackInSlot(range.head) == null) && insertIntoInventorySlot(stack, inventory, side, range.head, remaining, simulate)) {
+        remaining -= stackSize - stack.stackSize
+        success = true
+      }
+
       val shouldTryMerge = !stack.isItemStackDamageable && stack.getMaxStackSize > 1 && inventory.getInventoryStackLimit > 1
       if (shouldTryMerge) {
         for (slot <- range) {
@@ -166,7 +172,7 @@ object InventoryUtils {
         }
       }
 
-      for (slot <- range) {
+      for (slot <- range.tail) {
         val stackSize = stack.stackSize
         if ((inventory.getStackInSlot(slot) == null) && insertIntoInventorySlot(stack, inventory, side, slot, remaining, simulate)) {
           remaining -= stackSize - stack.stackSize


DirtYiCE:
Sorry, there was a bug in the first version.

fnuecke:
Hmm, I find the fact that merging takes precedence to be quite intuitive. Amongst other things because it's also what happens for a player's inventory.

The main issue I see with this change is that it can very subtly break existing scripts, making me _almost_ classify it as a major breaking change.

If this goes in I think this should get its own method, though (that, after doing the added check, calls the existing `insertIntoInventory`). For one because that method is also used by other parts of the mod where this behavior would be very unintuitive (debug card inserting into inventories, where there is no "selected" slot e.g.), but also because I would likely forget about this special case and use the method wrongly myself at some future point ;-)

fnuecke:
Oh, and re

> you have no idea what you just picked up at where it ended

This might be alleviated by `suck` (and family) returning a second value, the slot whatever was picked up got put into? This would be non-breaking but also resolve your gripe with the current behavior, right?

DirtYiCE:
Yes, that second return value should also solve my problem. And actually it would probably use the robot's inventory more efficiently than my current hacked together script :)

But back to the inventory management: I think there's still place for a more "strict" variant of these inventory management functions. For example a machine that has multiple imputs where it's significant what you place in which slot. (But actually unless you have to put the same item in multiple slots the current implementation also works, provided nobody else uses the machine...)

DirtYiCE:
Actually there is a problem with that second return value: what if the itemstack ends up in multiple slots? Return all of them?

SoniEx2:
> Hmm, I find the fact that merging takes precedence to be quite intuitive. Amongst other things because it's also what happens for a player's inventory.

That implies shift-clicking, which is not the case here?

fnuecke:
> For example a machine that has multiple imputs where it's significant what you place in which slot.

There's `inventory_controller.dropIntoSlot`, which doesn't merge or do overflow stuff.

> Actually there is a problem with that second return value: what if the itemstack ends up in multiple slots? Return all of them?

Hmm, valid point. I... guess? A table seems like overkill, since it'd _usually_ only hold one entry?

> That implies shift-clicking, which is not the case here?

How does picking up an item from the ground involve shift-clicking? ;-)

SoniEx2:
> How does picking up an item from the ground involve shift-clicking? ;-)

If we're talking about that... how does it go into the selected slot? :P

fnuecke:
Had a look at what'd be necessary to return the slot something was inserted to... and it'd be relatively messy. Unless I'd just do what you could do in Lua anyway - check all slots before and after. So I'll probably merge this in OC 1.5, where there'll be other breaking-ish stuff anyway.

PR #889, opened 2015-02-04
Create yes.lua

hunator:
Lua implementation of functionality provided by yes command from GNU coreutils (in Linux). However it doesn't use any of GNU code, so it is not forcing GPLv3+ license.

It has full functionality but options had to been changed to single letter (like -h instead of original --help in GNU). Main use case requires piping, but works without it.

I have also created man page, but I have to send it in different pull request (don't know how to do it in github webpage).

yes
yes man

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/yes.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/yes.lua
@@ -1,0 +1,43 @@
+--[[Lua implementation of the UN*X yes command--]]
+local shell = require("shell")
+
+local args, options = shell.parse(...)
+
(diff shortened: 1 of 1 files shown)


Vexatos:
Why did the options have to be changed? "--help" would work in OC.
Also, to add a file to this Pull Request, just go to https://github.com/hunator/OpenComputers/tree/patch-1/src/main/resources/assets/opencomputers/loot/OpenOS and add a new file there.

hunator:
So that is how to do it... Thanks
With that options, I've got distracted by -- stop processing options at -- in shell.lua file.

hunator:
Now with --version and --help options like in GNU coreutils.

PR #897, opened 2015-02-08
Allow "flattened" converted values to be recursively converted

benblank:
Currently, converters which use the `oc:flatten` special key aren't given the benefit of recursive conversion.  In other words, the following code (for example) in a converter produces the "Tried to push an unsupported value" warning rather than a converted item stack.

```
output.put("oc:flatten", itemStack);
```

This change simply waits to flatten until _after_ recursive conversion has occurred.


fnuecke:
Thanks!

PR #901, opened 2015-02-08
Add converters for World and WorldProvider objects

benblank:
The rationale for dividing this into two converters is that a) it saves the effort of looking up a `World` reference when one already has a `WorldProvider` reference; and b) while the information is _actually_ contained within a `WorldProvider` object, it _logically_ belongs to a "world", which simply happens to be represented by multiple objects.

`id` is exposed so that worlds/dimensions can be correctly distinguished even if they happen to have the same name.  `name` is, of course, exposed for player-friendly display.

(Note also that this change is dependent on PR #897 in order to function correctly.)

### DIFF
--- a/src/main/scala/li/cil/oc/integration/vanilla/ConverterWorld.scala
+++ b/src/main/scala/li/cil/oc/integration/vanilla/ConverterWorld.scala
@@ -1,0 +1,17 @@
+package li.cil.oc.integration.vanilla
+
+import java.util
+
+import li.cil.oc.api
+import net.minecraft.world
+
+import scala.collection.convert.WrapAsScala._
+
+object ConverterWorld extends api.driver.Converter {
+  override def convert(value: AnyRef, output: util.Map[AnyRef, AnyRef]) =
+    value match {
+      case world: world.World =>
+        output += "oc:flatten" -> world.provider
+      case _ =>
+    }
+}
--- a/src/main/scala/li/cil/oc/integration/vanilla/ConverterWorldProvider.scala
+++ b/src/main/scala/li/cil/oc/integration/vanilla/ConverterWorldProvider.scala
@@ -1,6 +1,9 @@
 package li.cil.oc.integration.vanilla
 
+import java.nio.ByteBuffer
+import java.security.MessageDigest
 import java.util
+import java.util.UUID
 
 import li.cil.oc.api
 import net.minecraft.world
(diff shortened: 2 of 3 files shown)


fnuecke:
Hrm, I'm a little uncomfortable with exposing the `id` of a world; that feels a little too much like an implementation detail to me. What would a program need it for, anyway?

benblank:
I understand your concern!  Unfortunately, dimension names aren't really suitable for comparison.  They aren't guaranteed to be unique (e.g. AE spatial IO dimensions), nor are they even guaranteed not to change, I believe.

World `id`s may be a bit of a grungy detail, but they're the only real way to track dimension identity.

As a use case, I'm currently working on another pull request which expands Blood Magic support to include a converter for teleposition foci, which are bound to a set of coordinates in a particular dimension.  An OC program for managing a library of foci may well need to compare worlds, for which their name alone wouldn't be sufficient.

Vexatos:
I wouldn't want the `id` as well, because those are numbers that shouldn't be shown outside admin commands and the F3 screen.

benblank:
@Vexatos — What about a hash of the dimension ID?  It would still allow high-confidence comparison, but wouldn't expose the actual dimension number.

fnuecke:
How about making it a hash of id plus seed? Or maybe of just the seed? I'm actually not sure if two dimensions on the same server with the same seed will look exactly the same... if that's the case that'd be the only reason to even include the id in the hash, I think?

benblank:
I'm not sure what's to be gained by including the seed?  Even if two dimensions have identical seeds and terrains, they're still two distinct places (changes in one won't affect the other, after all).  Seeds also aren't _necessarily_ constant, though I'm not sure that actually happens outside of a server crash or some such.

For reference, under normal circumstances, the seed for the overworld, the Nether, and the End are all the same.

fnuecke:
The original thought was to use the seed as salt when hashing the ID. I didn't think it could change over the lifetime of a world / save.

benblank:
I'm not certain that it _does_ change outside error conditions (usually, `level.dat` becoming corrupt and needing deleted).  I'm just pretty sure it _can_.  And that raises the possibility of a mod out there which does exactly that.  :-)

On the other hand, I'm not aware of much else that would make a good salt.  For now, I'll keep poking and see if I can find something likely.

benblank:
After some digging, I really don't think there's anything else appropriate to use as a salt — and seeds shouldn't change outside of unusual circumstances, anyway.  I've gone ahead and updated the PR.

fnuecke:
OK, cool, thanks for looking into it! The only thing I'll probably change (or you can do that yourself before I pull it later today) is to replace the hashing with the Guava one; since Forge provides that anyway, and it's a lot more readable IMHO. It'd become

``` Java
Hashing.md5().newHasher().
          putLong(provider.getSeed).
          putInt(provider.dimensionId).
          hash().asBytes()
```

PR #912, opened 2015-02-13
Clean io mode for sh.lua

magik6k:
This way there is nice way to implement basic remote telnet-like thing without any ugly hacks

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -253,6 +253,23 @@ if #args == 0 and (io.input() == io.stdin or options.i) and not options.c then
       end
     end
   end
+elseif #args == 0 and (io.input() ~= io.stdin) then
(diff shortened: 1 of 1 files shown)


fnuecke:
Looks good, thanks!

src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
```diff
@@ -253,6 +253,23 @@ if #args == 0 and (io.input() == io.stdin or options.i) and not options.c then
       end
     end
   end
+elseif #args == 0 and (io.input() ~= io.stdin) then
+  while true do
+    io.write(expand(os.getenv("PS1") or "$ "))
+    local command = io.read("*l")
+    if not command then
+      io:write("exit\n")
```
SoniEx2, line 261: "io:write"?

fnuecke, line 261: Good catch.

PR #915, opened 2015-02-16
Event view program for OpenOS

magik6k:
This is IMO one of the most useful utilities OpenOS was lacking. Name is probably not super-accurate but it seems good enough.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
@@ -1,14 +1,16 @@
 local event = require "event"
 local component = require "component"
+local keyboard = require "keyboard"
 
 local interactive = io.output() == io.stdout
 local color, isPal, evt
 if interactive then
   color, isPal = component.gpu.getForeground()
 end
+io.write("Press 'q' to exit\n")
 pcall(function()
   repeat
-    evt = {event.pull()}
+    evt = table.pack(event.pull())
     if interactive then component.gpu.setForeground(0xCC2200) end
     io.write("[" .. os.date("%T") .. "] ")
     if interactive then component.gpu.setForeground(0x44CC00) end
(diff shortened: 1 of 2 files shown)


Vexatos:
How about a `Press <key> to stop` print in the beginning?

src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
```diff
@@ -0,0 +1,31 @@
...
+local component = require "component"
+
+local interactive = io.output() == io.stdout
+local color, isPal, evt
+if interactive then
+  color, isPal = component.gpu.getForeground()
+end
+pcall(function()
+  repeat
+    evt = {event.pull()}
```
fnuecke: Better to use `table.pack` here, to avoid dropping parameters after a `nil` one.

src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
```diff
@@ -0,0 +1,31 @@
...
+    if interactive then component.gpu.setForeground(0xCC2200) end
+    io.write("[" .. os.date("%T") .. "] ")
+    if interactive then component.gpu.setForeground(0x44CC00) end
+    io.write(tostring(evt[1]) .. string.rep(" ", math.max(10 - #tostring(evt[1]), 0) + 1))
+    if interactive then component.gpu.setForeground(0xB0B00F) end
+    io.write(tostring(evt[2]) .. string.rep(" ", 37 - #tostring(evt[2])))
+    if interactive then component.gpu.setForeground(0xFFFFFF) end
+    if #evt > 2 then
+      for i = 3, #evt do
```
fnuecke: Then use `evt.n` here (from `table.pack`).

src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
```diff
@@ -0,0 +1,31 @@
...
+        io.write("  " .. tostring(evt[i]))
+      end
+    end
+    
+    io.write("\n")
+  until evt[1] == "key_down" and evt[3] == 113
```
fnuecke: Maybe better `evt[2] == keyboard.keys.q` here, or at least a comment. Magic numbers are annoying when you have to look them up ;-)

fnuecke:
Also, I agree with Vexatos' suggestion of adding a message on startup to let users know how to quit (without ctrl+alt+c, which could potentially screw up colors).

PR #923, opened 2015-02-18
Create pt_BR.lang

ghost:


fnuecke:
Thanks. Do you plan on completing the translation (e.g. the tooltips) or should I merge this as-is?

ghost:
Sorry, unfortunately I have no time to finish the translation and review it ingame.
I'm working 15 hours a day and lately I don't have much spare time to Minecraft.
[]s

LordFokas:
It bothers me a bit that Sangar is credited for the creation of the pt_PT file in the pt_BR one, given I am the author (although sangar committed a lot of small changes to it). It also confuses me, why doesn't my commit show up the pt_PT's history? A git server is not the kind of thing that forgets were stuff came from.

I should also fix the pt-PT file, when I have the time. It just happens I often don't, and when I do patching lang files isn't the first thing that comes to my mind :(

Also, if someone could put the original en_US key/value pairs that are missing in the pt_PT file, it would ease my life a lot in terms of attempting to translate it, which would also ease @al-myr's life patching the pt_BR one as he could simply pull stuff from pt_PT instead of checking for missing entries.

fnuecke:
@LordFokas The reason it doesn't show up seems to be that Github doesn't track that info across file moves/renames. If you look at [the history starting at the commit before the rename](http://git.io/p6Nl) (which was for gradle, I so do not miss the old days) all is as it should be.

fnuecke:
Closing this due to inactivity.
If someone wants to finish this, please create a new PR with the updated files, thanks.

PR #924, opened 2015-02-19
Added basic hostname mechanism to OpenOS

magik6k:
It gives some simple 'standard' way of naming machines, useful for networked applications. Also, the network floppy is compatibile with that.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/hostname.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/hostname.lua
@@ -1,14 +1,20 @@
 local args = {...}
 if args[1] then
-  local file = io.open("/etc/hostname", "w")
-  file:write(args[1])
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/bin/hostname.lua
```diff
@@ -0,0 +1,14 @@
+local args = {...}
+if args[1] then
+  local file = io.open("/etc/hostname", "w")
+  file:write(args[1])
```
fnuecke: Better a `file, reason = ...` with a `if not file then io.stderr:write(reason .. "\n") else ...` here to avoid comparatively cryptic nil errors on read-only file systems / when out of disk space?

magik6k:
Done

fnuecke:
Looks good, thanks!

PR #939, opened 2015-02-25
Improved rc script system

magik6k:
Tested, this time it wont break OpenOS, I promise ;p. All changes are backwards compatible.
- Implicit restart when start&stop is present
- disable/enable commands for easier management
- Listing commands in service
- Documentation!

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
@@ -1,4 +1,5 @@
 local fs = require('filesystem')
+local serialization = require('serialization')
 
 -- Keeps track of loaded scripts to retain local values between invocation
 -- of their command callbacks.
@@ -16,6 +17,19 @@ local function loadConfig()
     end
   end
   return nil, reason
+end
+
+local function saveConfig(conf)
+  local file, reason = io.open('/etc/rc.cfg', 'w')
+  if not file then
+    return nil, reason
+  end
+  for key, value in pairs(conf) do
+    file:write(tostring(key) .. " = " .. serialization.serialize(value) .. "\n")
+  end
+  
+  file:close()
+  return true
 end
 
 function rc.load(name, args)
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
```diff
@@ -39,14 +53,47 @@ function rc.unload(name)
...
+          io.output():write(tostring(command) .. " ")
+        end
+      end
+      return true
+    elseif type(result[cmd]) == "function" then
+      res, what = xpcall(result[cmd], debug.traceback, ...)
+      if res then
         return true
       end
+    elseif cmd == "restart" and type(result["stop"]) == "function" and type(result["start"]) == "function" then
```
rmellema, line 72: Maybe also test for the non-existance of a `restart` command here? Maybe a user wants to have control over restarting that is different from simply stopping and starting again.

rmellema:
I may be a little late to the party, but I still wanted to comment on this.

First of all, its nice to see someone give the rc system some love! If I had time I should have done it myself. 

Second thing is about the printing by the library. I think this is kinda ugly, since this takes control away from programmers and is likely to break GUIs. I think this is better done by making this into a separate function for the rc lib. Then the binary can notice the missing of the command and print the help itself. This function could then also be used by other programmers to get information about a script.

Third thing is about the enabling and disabling commands. I feel like this could better be done by `/bin/rc.lua` using flags, so like:

```
# rc -e example
```

To enable example and:

```
# rc -d example
```

To disable example. This is just personal preference though.

src/main/resources/assets/opencomputers/loot/OpenOS/lib/rc.lua
Kubuxu, line 72: Look at L67. If there is restart command, script will choose that branch.

rmellema, line 72: Oh, derp, missed that. Never mind then.

PR #983, opened 2015-03-30
Mandatory OPPM Update

Vexatos:
Not sure why you would want this as OPPM auto-installs itself from the latest version anyway, unless _you know what you are doing_...


fnuecke:
Thanks!

PR #991, opened 2015-03-31
cat.lua now closes opened files properly

mpmxyz:
The title says everything.
It caused a lot of Out of Memory-Errors when I tested a custom file system. (-> many unclosed streams)


fnuecke:
Thanks!

PR #995, opened 2015-04-01
Update .gitignore

xbony2:
This is so mac users like me can send pull requests without the .DS_Store file that are auto-generatored by OS-X, since most people don't like those. See: http://en.wikipedia.org/wiki/.DS_Store

(I'm surprised this hasn't be done yet)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@
 # Debugging
 /eclipse
 /run
+
+# OS X
+.DS_Store


rmellema:
You can use a global gitignore file and put the .DS_Store file in there. Then it will never get included and you don't have to exclude it in each and every repo you have.

fnuecke:
Can't hurt to have it either, thanks!

xbony2:
Thanks :3

PR #1003, opened 2015-04-02
Cortana added to robots.name

KJA1582:
Now also includes mod integration drivers for Flaxbeard's Steam Power.

NOTE: You might have to look into the dependencies, I couldn't figure it out. This why there is a Flaxbeard.jar (it is version 0.28.7, latest at the time of writing) in the main folder. I used the compile dependency in build.gradle.


build.gradle
```diff
@@ -178,6 +178,9 @@ dependencies {
 
     compile 'com.google.code.findbugs:jsr305:1.3.9' // Annotations used by google libs.
 
+    //EDITS for Flaxbeard's Steam Power
+    compile files('Flaxbeard.jar')
```
fnuecke: Please add the dependency in a way it can be autodownloaded. If FSP has a maven, that'd be ideal, otherwise add it the the CoFH stuff is referenced, e.g. (i.e. via Curseforge).

build.properties
```diff
@@ -37,4 +37,7 @@ re.version=3.0.0.342
 waila.version=1.5.8a
 wrcbe.version=1.4.1.2
 
+//EDIT_for_Flaxbeard's_Steam_Power_integration
+fsp.version=0.28.7
```
fnuecke: Just insert it where it belongs in alphabetical order, no needs for the "EDIT"s, that's what we have Git for ;-)

src/main/scala/li/cil/oc/integration/Mods.scala
```diff
@@ -72,6 +71,9 @@ object Mods {
   val WirelessRedstoneCBE = new SimpleMod(IDs.WirelessRedstoneCBE)
   val WirelessRedstoneSVE = new SimpleMod(IDs.WirelessRedstoneSV)
 
+  //EDIT Flaxbeard's Steam Power, povidesPower = false because it doesn't have any RF Gen.
+  val FlaxbeardsSteamPower = new SimpleMod(IDs.FlaxbeardsSteamPower, providesPower = false)
```
fnuecke, line 75: Ditto, alphabetical order.

src/main/scala/li/cil/oc/integration/Mods.scala
```diff
@@ -111,7 +113,10 @@ object Mods {
 
     // We go last to ensure all other mod integration is done, e.g. to
     // allow properly checking if wireless redstone is present.
-    integration.opencomputers.ModOpenComputers
+    integration.opencomputers.ModOpenComputers,
+
+    //EDIT for Flaxbeard's Steam Power
+    integration.fsp.ModFSP
```
fnuecke, line 119: Ditto, alphabetical order.

src/main/scala/li/cil/oc/integration/Mods.scala
```diff
@@ -179,6 +184,9 @@ object Mods {
     final val Waila = "Waila"
     final val WirelessRedstoneCBE = "WR-CBE|Core"
     final val WirelessRedstoneSV = "WirelessRedstoneCore"
+
+    //EDIT for Flaxbeard's Steam Power
+    final val FlaxbeardsSteamPower = "Steamcraft"
```
fnuecke, line 189: Ditto, alphabetical order.

src/main/scala/li/cil/oc/integration/fsp/DriverSteamGauge.scala
```diff
@@ -0,0 +1,32 @@
...
+  class Environment(tileEntity: TileEntitySteamGauge) extends ManagedTileEntityEnvironment[TileEntitySteamGauge](tileEntity, "steamGauge") with NamedBlock {
+    override def preferredName = "steamGauge"
+
+    override def priority = 0
+
+    @Callback(doc = "function():number -- Returns the valves pressure.")
+    def getPressure(context: Context, args: Arguments): Array[AnyRef] = result(tileEntity.getPressure)
+
+    @Callback(doc = "function():boolean -- Returns the comparator output of the gauge.")
+    def getComparatorValue(context: Context, args: Arguments): Array[AnyRef] = result(tileEntity.getComparatorOutput)
```
fnuecke, line 29: `getComparatorValue` is the vanilla one, isn't it? In that case, no, there's the comparator block for that.

src/main/scala/li/cil/oc/integration/fsp/DriverSteamBoiler.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def createEnvironment(world: World, x: Int, y: Int, z: Int): ManagedEnvironment = new Environment(world.getTileEntity(x, y, z).asInstanceOf[TileEntityBoiler])
+
+  class Environment(tileEntity: TileEntityBoiler) extends ManagedTileEntityEnvironment[TileEntityBoiler](tileEntity, "boiler") with NamedBlock {
+    override def preferredName = "boiler"
+
+    override def priority = 0
+
+    @Callback(doc = "function():number -- Returns the boilers pressure.")
+    def getPressure(context: Context, args: Arguments): Array[AnyRef] = result(tileEntity.getPressure)
```
fnuecke, line 26: Seeing as `getPressure` is in all three drivers, I have to wonder: is there no interface that provides this you could make a driver for?

fnuecke:
Aside from the above, the drivers look OK! When done, please squash your commits.

fnuecke:
Aaand I just read your not. Derp. As I said above, look into how the CoFH ones are done. You get the numbers and by looking at the download URL when you get the file from Curseforge.

KJA1582:
What is squashing my commits? I don't use git that offen/much

KJA1582:
By looking at the CoFH, you do mean the ones in build.gradle? Cause I don't quit understand how they are defined as such.

KJA1582:
Ok. Got the autodownloady bit figured out, but now it fails on the javadoc because fsp is called Flaxbeard'sSteamPower and it doesn't like the ' in it. How can I exclude it?

src/main/scala/li/cil/oc/integration/fsp/DriverSteamBoiler.scala
lordjoda, line 26: `getPressure`is from ISteamTransporter.java

Vexatos:
I hope you know that Computronics already has support for all of these so it's actually pretty unnecessary...

KJA1582:
CC has, but unless CC is installed with OC, you can't use it.

Vexatos:
Of course you can, Computronics adds FSP integration to ComputerCraft and OpenComputers.

KJA1582:
I couldn't access it while only having OC installed. in RR3, i did have access to all CC methods.

Vexatos:
Well, [is definitely is there](https://github.com/asiekierka/Computronics/blob/master/src/main/java/pl/asie/computronics/integration/fsp/DriverSteamTransporter.java), and it also [definitely is being registered](https://github.com/asiekierka/Computronics/blob/master/src/main/java/pl/asie/computronics/oc/IntegrationOpenComputers.java#L131-L135).

KJA1582:
Nice, but it isn't OC native.

Vexatos:
...

Isn't that what compat addons are for? Adding integration with other mods?

KJA1582:
Native is like BootCamp. Computronics is more like Parallel Desktops. Nothing wrong with that, but thats the way I see it.

LizzyTrickster:
Also having loads of mod Compatibility in OC itself puts more strain on Sangar because he has to check if any of their APIs have breaking change and potentially support more than one version of the API in some rare cases..

Just my two pennies

**Edit spelling -.-

KJA1582:
Thats a good point.

vifino:
I agree with @LizzyTrickster, and I think using compat addons is just fine.

KJA1582:
There is nothing wrong with compat add ons, don't get me wrong. I just wanted a more native implementation, thats all. After all, thats why open software exists, to build on it and expand it.

fnuecke:
Muh inbax.

So here's the thing: Computronics is, AFAIK, in every modpack that has OC in it anyway (for good reason), and soon OpenPeripheral, too, will have OC support. So there's that. I understand your desire to have the integration bundled with OC itself, as it'd make it more "batteries included", but as @LizzyTrickster pointed out, in the long run maintaining the integration is something _I_ will have to do, either because whoever added it isn't available anymore, or has no time, or it has to get fixed quickly. And it piles up.

That doesn't mean I want no integration code in OC, on the contrary. But adding new integration only makes sense to me, if it's not already present in one (or possibly two!) other mods, each of which are quite popular. And seeing as this is already present in Computronics, I don't think it makes much sense to duplicate the driver. And it doesn't make much sense to move that code from Computronics to OC, either, as it supports other mods, too, which is a huge argument in favor of having separate integration mods.

Thanks a lot for your work and interest nonetheless! If you find any other things that might make sense to add integration for - ideally without requiring JAR dependencies, but that's admittedly rare - please do suggest and discuss that!

KJA1582:
Will do, I actually didn't know there already was integration for this, so this one is my mistake.

Anyway, I still have that failing javadoc on the build, how do I solve it? Using the exclude statement didn't do anything, I probably set it up wrong somehow.

LizzyTrickster:
@Vexatos did point out the intergration....

fnuecke:
No idea what that error is you're getting there, sorry.

He did, but only after the PR was made, unless I missed something?

KJA1582:
AFTERWARDS, I didn't know this mod even exists/did it.

KJA1582:
It can't find the packages sSteamPower, which occurs since in the name there is an ', which also is the string tag

fnuecke:
No clue, but that's more of a Gradle question, and there are places for that. I'll be locking this before it gets any more off topic...

PR #1008, opened 2015-04-02
Label some stuff

ghost:
For developers.


magik6k:
Not really useful imho

fnuecke:
I agree, this doesn't really add anything that wouldn't be obvious from the context.

ghost:
Good point.

PR #1009, opened 2015-04-02
Update README.md

xbony2:
This gives a little bit more information for those of us who do not use the IDEA ways.


fnuecke:
Thanks! `then` -> `than` (both of them), and I'll merge it ;-)

xbony2:
Sorry about that >.>

PR #1036, opened 2015-04-09
In-game documentation files

rashdanml:
Made as many changes to the in-game documentation as I felt comfortable making. Added links where necessary. Also added /item/worldSensorUpgrade.md since Sangar forgotted it (in all fairness, I forgot to create pages for it earlier on ocdoc XD). Tried to be as consistent as possible, but a few little bits may have slipped through. 

We should probably include component API documentation, would be easier to access than typing commands in a lua environment in a computer.


fnuecke:
Awesome, thanks a _lot_! I'll have to go through the list once more... probably forgot some other things, too >_>

For reference, [this is pretty much the complete list](http://git.io/ve9OB).

fnuecke:
Update: I think I got all of the ones that make sense now. That the network debugger item doesn't have one is intentional.

PR #1040, opened 2015-04-09
OC doc crafting pics, should be merged soon

KJA1582:
@fnuecke  

^Title. Added pics to the Documentation regarding how to craft the item. _Should_ be complete.


fnuecke:
I really appreciate the effort, which is why it honestly pains me to turn this down. For one, the images are too verbose, if anything it should be the crafting grid, and the crafting grid only. For two, if at all, this will be added as dynamically generated content, because it's much too easy to get out of sync, either from updates or modpacks using custom recipes as the two most obvious examples.
An intermediary solution will probably to allow pressing 'r' while on a page to open NEI with a recipe handler for the item of the currently shown page, or something like that.

Something that does need doing, and AFAIK noone is working on right now would be a block index, i.e. the equivalent to item/index.md for the blocks, if it's OK to steer your motivation a little ;-) [as one single commit, preferably]

KJA1582:
I already thoughtnthat The pics are to large, oh well. 

I will Look into The block Index.

Single commits are rauher hard when The upstream updaten 10 Times while working on something. I just like to be up to date.

fnuecke:
Doesn't mean you can't be! Ask your favorite search engine for `git squash commits`.

KJA1582:
So my network decided to denie Internet access, and I can't restart it because I'm doing a 300gb backup :/

This means, that I'm stuck with outdated sources (no items/index.md for example) and I also cannot push anything.

If it resolves itself I will proceed with a block index, otherwise I'll have to wait for ca. 15hrs till that backup is done.

KJA1582:
Somewhat fixed, checking layout of the new index file, expect PR soon.

PR #1057, opened 2015-04-11
Added text to index.md, added lua.md and openOS.md

rashdanml:
Title. All changes as one commit without breaking any other files, yay. As far as I know, only these files used the new formatting for code blocks, so the remaining files should be good.


fnuecke:
Thanks a lot!

PR #1092, opened 2015-04-19
Basic self-update program for OpenOS

dgelessus:
New `--ignore` flag for `install` to specify destination directories where files should not be overwritten, and `update` command to update OpenOS to a newer version from a floppy. Also, man pages for `install` and `update`.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/install.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/install.lua
@@ -6,8 +6,8 @@ local shell = require("shell")
 local unicode = require("unicode")
 
 -- Modified version of recurse from /bin/cp.lua
-local function rcopy(fromPath, toPath, ignore)
-  ignore = ignore or {}
+local function rcopy(fromPath, toPath, update)
+  update = update or false
   os.sleep(0) -- allow interrupting
   if fs.isDirectory(fromPath) then
     if fs.canonical(fromPath) == fs.canonical(fs.path(toPath)) then
@@ -20,7 +20,7 @@ local function rcopy(fromPath, toPath, ignore)
     print(fromPath .. " -> " .. toPath)
     fs.makeDirectory(toPath)
     for file in fs.list(fromPath) do
-      local result, reason = rcopy(fs.concat(fromPath, file), fs.concat(toPath, file), ignore)
+      local result, reason = rcopy(fs.concat(fromPath, file), fs.concat(toPath, file), update)
       if not result then
         return nil, reason
       end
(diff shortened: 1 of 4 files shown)


fnuecke:
Hmm, having the code duplicated from `cp` is somewhat... meh. How about we also add something like `--ignore` to `cp`? Then just pass it along?

As for the defaults, I think it'd be better to explicitly mention the files to not copy, since the examples in /etc would make sense to always keep update (with, as mentioned on the forums, rc.cfg renamed to rc.cfg.example; then, if that ever needs to change, the example can be updated). So the default stuff to not update would probably only be... `/etc/motd`?

Aaactually... hmm. OK, slap me if this is overkill, but as alternative approach: what about an `--upgrade` option for `install`, and if that's set, compute a hash for each file, compare those, if the file changed ask user whether to overwrite (with [Y/n]), else ignore the file?

Kubuxu:
@fnuecke This is great idea.
The only thing we need more is pipes and `yes` program. Just joking.

dgelessus:
Yeah, I agree the copied code from `cp` isn't a great solution... but it was the best way I could think of. `--ignore` seems out of place on `cp` IMO. The hash-and-confirm-when-different option sounds like a good idea though.

@Kubuxu Ever tried the `besh` disk?

fnuecke:
@dgelessus Yeah, I agree the ignore arg for cp is kinda weird. Let's agree on the upgrade parameter then, even if that'll involve the copied code after all. Will be enough of a change to justify it, though, I guess. Do you want to have a go at that? The main problem will be that it'll be kinda slow, so some kind of status indication is a must (just printing the currently operated on file like for the normal installation would be fine I think).

@Kubuxu [yes](http://git.io/vfv7M) ;-) Joking aside, might have another look at piping an bring that to the default shell... besh is still around, ofc, but I don't think anyone's actually using it?

Kubuxu:
@fnuecke Yeah. I think everyone misses it. Maybe make it build-in into OpenOS and allow to switch shell program from default to besh, as IIRC it isn't default becouse of RAM usage.

natedogith1:
For the upgrade option, could lastModified be used in some way to detect what files have been changed?

dgelessus:
@natedogith1 That would require knowledge of when the mod was last updated. I would be surprised if the `openos` loot disk properly emulated file metadata.

natedogith1:
The loot disk might not (and actually probably wouldn't, what with it being a zip file and all), but I was more thinking some marker file on the drive you're installing to, any file on the drive older than that file wouldn't have been modified and could be safely overridden. Or perhaps a better idea would be to have that file store when the last installation took place, that way re-saving the file probably wouldn't break things, but then there'd be issues if you copied the file.

fnuecke:
While leaving a "date of installation" marker file could work, I think it's a little too unreliable, since it'll rely on the underlying real file system while unloaded. So moving the files outside of the game, for example, could invalidate that, leading to nothing being updated. Which could be highly... unexpected.

The most reliable way would obviously be to keep a list of hashes of the previous version(s), but that's way too much bookkeeping.

dgelessus:
Okay, I replaced `--ignore` with `--update`. At the moment this checks for file equality by reading 4 KB chunks and comparing those using normal Lua string comparison. This is probably far from efficient, no idea how Lua does string equality checks internally. If anyone can come up with a better way... feel free to PR after this has been merged I guess?

lordjoda:
Hmm could you do a size check before you do the equality check? If the size is not equal you know that something must have changed, so you only need a "backup" check for equality if the size remains the same. Speaking of size: there might be cases when only a 0 is changed to a 1 or so, but in general, the file size should not really ever be the same after a change should it? So of course this makes it a "bit" unsafe but is a size check not basically enough?

Kubuxu:
IMHO size check is not enough but it can be form of an optimization.

dgelessus:
Right, of course if the sizes are different, the files must be different as well. That'll at least speed up the difference checking for most files that changed.

dgelessus:
@fnuecke Ping?

fnuecke:
Ah, this is done-ish? Will have a look over it when I can!

fnuecke:
All right, so. I really dislike the duplicate functionality from `cp`, after all. So here's an idea: adding a bastardized version of `-u` to `cp`: copy only if the files _differ_ or target doesn't exist. Combine that with `-i`, and the user will be queried for changed files as desired, and we can reuse `cp` by simply adding another option. Sound good?

If noone objects/points me toward something where this doesn't work, which I overlooked, I'll be working on that. As in, I'll try to get it done by tomorrow, because that's when I'd kinda like to release 1.5.9-rc.1.

dgelessus:
That's also an option. (The real `cp` also has an `-u`/`--update` flag, which works based on modification time. The problems with that have already been discussed though, checking for true equality is a better option in this case.)

fnuecke:
Yep, hence "bastardized" :P

PR #1095, opened 2015-04-20
Fix: standard io streams being closable

mpmxyz:
To be exact: Not the streams themselves but the buffers on top of them can be closed.
I noticed that because "ls" didn't display anything after some experiments involving an "stdout" file.


fnuecke:
Ahh, good point. I'd have preferred to handle this in the buffer itself, but since `close` will pretty much always return `nil`, determining whether a close was successful would require checking if the second return value isn't... which seems rather... unreliable.

PR #1111, opened 2015-04-29
Implement event.pullMultiple and event.pullFiltered

Kubuxu:
Awaiting any constructive critic. 
`event.pullMultiple` - allows you to specify multiple events names.
`event.pullFiltered` - allows you to specify filtering function.

Also implement soft interrupt which activates on Ctrl-C.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
@@ -2,15 +2,21 @@ local event = require "event"
 local component = require "component"
 local keyboard = require "keyboard"
 
+local args = {...}
+
 local interactive = io.output() == io.stdout
 local color, isPal, evt
 if interactive then
   color, isPal = component.gpu.getForeground()
 end
-io.write("Press 'q' to exit\n")
+io.write("Press 'Ctrl-C' to exit\n")
 pcall(function()
   repeat
-    evt = table.pack(event.pull())
+    if #args > 0 then
+      evt = table.pack(event.pullMultiple("interrupted", table.unpack(args)))
+    else
+      evt = table.pack(event.pull())
+    end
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/opencomputers/loot/OpenOS/lib/event.lua
```diff
@@ -118,28 +144,50 @@ end
...
+    checkArg(1, args[1], "number", "nil")
+    checkArg(2, args[2], "function", "nil")
+    seconds = args[1]
+    filter = args[2]
   end
 
   local deadline = seconds and
                    (computer.uptime() + seconds) or
-                   (hasFilter and math.huge or 0)
+                   (filter and math.huge or 0)
```
fnuecke, line 190: There's no way of having no filter in here anymore, so that could always be `math.huge`, unless I'm missing something? Better yet, just set seconds to `math.huge` in the non-timeout if branch, I guess?

Kubuxu, line 190: No. Filter can be nil, then there is no filter. I mimicked previous functioning.

src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
```diff
@@ -25,8 +31,8 @@ pcall(function()
     end
     
     io.write("\n")
-  until evt[1] == "key_down" and evt[4] == keyboard.keys.q
-end)
+  until evt[1] == "interrupted"
+end
```
Kubuxu: God. My mistake. Fixing it.

src/main/resources/assets/opencomputers/loot/OpenOS/lib/event.lua
fnuecke, line 190: Ah, OK.

PR #1152, opened 2015-05-15
Fixed Interface named as me_controller when chanels disabled

DrummerMC:


fnuecke:
Hrm, I guess things will currently get incompatible either way, with this being the more intuitive approach... by which I mean programs written to use an `me_controller` won't work in worlds with channels disabled without adjustments and vice versa.

Actually, I think it'd make most sense to just allow what's currently the `me_controller` available to the controller block _and_ interface block, even when channels are enabled? It _is_ an interface, after all :P (although originally that means for items, but hey)
If there are no objections to that, I'll do that before merging this.

gjgfuj:
I have disliked how the me controller is required to be able to interface
with an me system. All a controller does is provide channels. Its basically
a switch. That's not really a good thing to control. I would definitely
vote for the interface being what's wrapped, since its an interface to the
network.

fnuecke:
With this there won't be a change in behavior, so that'll be fine. For the record, the idea of having the adapter consume an additional channel was brought up, but I think the interface consuming a channel is sufficient.

I'll probably entirely remove the controller driver in OC 1.6.

For the record: the reason it _was_ mainly via the controller until now was because that was simply carried over from AE1, where it made much more sense :P

PR #1153, opened 2015-05-16
Updated index.md and added endstone in russian translation

cyber01:
Updated index.md and added endstone in russian translation.

P.S. And why do we need fake endstone?


fnuecke:
Thanks.

Fake end stone: because craftable end stone apparently messes with vanilla balance or something...

PR #1158, opened 2015-05-17
Removed duplicate Ctrl+C-Handling from term.read

mpmxyz:
The event library already takes care about that. (returning an "interrupted" event)
I noticed that because I couldn't exit command lines reliably. (The Ctrl+C-Handling in the event library replaces some events.)


fnuecke:
Thanks!

Kubuxu:
Thanks too as I was who broke it.

PR #1168, opened 2015-05-23
Allow fake endstone in hoverUpgrade2 recipe

dgelessus:


fnuecke:
Ah, yes, oversight. Thanks!

PR #1190, opened 2015-05-30
Add Data Card.

Kubuxu:
Current Data Card provides following functions:
- SHA2-256
- MD5
- CRC-32
- base64 coding/decoding
- DEFLATE compression

It also provides following programs:
- deflate
- inflate
- sha256sum
- md5sum
- base64

Deflate compression reduces Lua file size by ratio about 2-3.

TODO:
- [x] base64 program
- [x] sha256sum program
- [x] md5sum program
- [ ] mans
- [x] recipe (I need help with that) - Thanks Sangar.
- [x] texture (I need help with that)
- [x] limits
- [x] energy costs
- [x] md5
- [x] crc-32

Any feature requests?


fnuecke:
Looking good so far!

If no one else steps up (but beware, I'm picky :P), I'll make a texture when I can.

As for the recipe, maybe something like

| ironNugget | alu | chip1 |
| --- | --- | --- |
| - | cardBase | - |
| - | - | - |

(not the correct names, but you get the idea)

General remark on the formatting: I have set my formatter to keep imports split up, because it makes it easier when merging stuff (in particular when porting to different Minecraft versions). I should really upload my settings... though that's really the only difference to IntelliJ defaults, I think.

Maybe also md5? Just because md5sum still not quite dead yet. And maybe CRC? (And if just to compare it to hashes of EEPROMs, because those use CRC.)

Also, a general point to discuss: limits. I'm fine with keeping the `direct=true`, but some limit should be put in. Something around 32 for the hashing stuff, 4-8 for the inflate/deflate maybe? Also, possibly some input size limits? Sure, generally there's a limit based on the RAM installed in computers, but there's also the fallback and - at least in the long run - potentially other architectures. Configurable, I suppose, with a getter - similar to the network card's packet size limit? Then there could be a Lua side program that splits inputs in chunks of the max size if it's larger than that?

Kubuxu:
I forgot to add it into TODO but limits and energy costs are going to be applied. Any suggestions on energy costs?

asiekierka:
Overpowered.

On a serious note, we should consider making more stuff available as cards. Still waiting for the native regex library.

EDIT: The programs should not be bundled with the card. What are we, an Option ROM?
EDIT2: If we want functionality on cards, we need more card slots in cases. Option 2 - add upgrades to CPU.

Kubuxu:
@asiekierka ROMs are already bundled with internet card, robot itself and generator upgrade.

I can't see how it is overpowered, everything that is in the data card can be done in Lua. The difference is that with the card you can use that in 1. real-time applications (hashing and deflating is incredibly slow in pure Lua), 2. environments with limited memory.

asiekierka:
I was joking with the Overpowered.

dgelessus:
@asiekierka

> Still waiting for the native regex library.

Wouldn't that need to be rewritten in Lua though? Unlike the codec/hashing functions, pattern matching time is not directly related to the pattern/input length and cannot be easily limited that way. Even Lua's wannabe regex matching apparently "[can take a looong time for certain "evil" patterns](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/resources/assets/opencomputers/lua/machine.lua#L35)", with real regexes you could probably do even worse things.

> The programs should not be bundled with the card. What are we, an Option ROM?

What @Kubuxu said - the Internet Card already does this, and IIRC the robot component as well (for the `robot.lua` lib).

fnuecke:
The file systems attached to certain components are kind of a relic of the olden days, really (from before loot disks iirc). It's something I'm not entirely happy with, but removing the existing ones now would cause too much confusion I'm afraid. So we could just as well go ahead and add more, in particular if the programs are tightly coupled to the component in question.

As for more cards for more functionality, I generally don't disagree. I'm not yet convinced more slots will be necessary, though. If anything, this'll encourage use of servers/multiple computers with networking more (i.e. virtual/remote components).

I agree with @dgelessus concerns about regex. This is the reason OC comes with a plain Lua port of Lua's pattern matching, because those, too, can be abused to lock up a program for a long time - this way the normal deadline enforcing functionality works seamlessly.

As for energy costs, I'd say let's orient ourselves on the (more expensive) GPU operations?

Kubuxu:
I was wondering about binding those programs to OpenOS itself and somehow making them available only when card is. They are almost OpenOS specific but I can't think of a simple way to make such lib injection possible from outside mode.

dgelessus:
The libs and programs could always be included with OpenOS, but fail to work when the required components are not installed. That would be the most logical solution IMO, the magical self-installing component filesystems are indeed a bit unusual. (They also don't work when autorun is disabled.)

Kubuxu:
Creating something like mergefs isn't a problem. We would need a way to make this scheme extensible for components outside of OC itself without PR into OpenOS.

dgelessus:
Those either provide (possibly craftable) loot disks, or put the libraries up for download somewhere on the internet. That's how you would get the drivers/libraries for a nonstandard piece of hardware IRL.

fnuecke:
Including them in the OS would be OK for built-in stuff, sure (since there's the package lib... I believe the separation was from before that? Possibly? To save memory or something.)

As for third-party libs... a compound FS could work (with third party stuff in subdirectories in /lib)... though that'd require special cases for the OpenOS loot disk again (which there currently are not, which I'm quite happy with). Then again, it'd be up to them to change it, if they wanted (i.e. migrate from auto attached FS to custom loot disk - which is now possible via the API, and hasn't been that long, after all).

natedogith1:
Could I request maybe adding AES/CFB8 to this card?

Kubuxu:
@natedogith1 I would have to ask for @Vexatos permission as he is providing simplified version of AES already (unfortunately it can't be used in OC <-> real world communication).

Vexatos:
@Kubuxu The Cipher Block AES is still something special as it's based on ItemStacks in the Block's inventory. Having AES in this card would not be an issue because it's AES, rather I can see it be an issue because this card already has so many features it can easily remove the need for the Advanced Cipher Block. Also I don't think encryption in a card is a good idea, it should stay a block...

Kubuxu:
Yeah. This is why I was somehow sceptical for idea of cripters in data card from beginning. 

I don't plan implementing RSA into it.

Vexatos:
@Kubuxu This card already is extremely powerful (too many features in one item, in my opinion, also it's still far too cheap, should be tier 2 or 3 in my opinion, too). I don't think having any more features would be good.

Kubuxu:
Point is that in many cases you will use this card for only one of the features and this card alone gives you almost nothing.

Vexatos:
Indeed, and it does make sense to have things like *sum on a card. But I don't think encryption should be on it.

fnuecke:
Am working on merging it now, made a texture (not sure I like it yet, though) and added soft and hard limits, as well as energy costs.

Kubuxu:
Oh, thanks Sangar. I started working on NOVA more and forgot a bit about this.

PR #1194, opened 2015-06-01
Made the SimpleComponent injection message more descriptive.

skyem123:
_(Disclaimer: @Victorious3 told me to do this, please don't shout at me, shout at him!)_
This makes the error message more descriptive so that the person seeing the error has a better idea of what mod caused their game to crash.

I should mention is that I only changed the English language file because I only know English, so people who have chosen another language won't get the extra detail.

Also, sorry if my coding is bad, I don't really know Scala.


fnuecke:
While in principle this looks good, IIRC displaying the string will fail this way for the other languages, because I think it'll fail if there is no pattern to match the given arguments for formatting the string. If I'm wrong on this, then fine, I guess.

In general I'd like to say that this was actually quite intentionally kept short, because it should really be fixed by devs _before_ they make a release, i.e. I'd expect them to see this in their dev-env; and I generally expect devs to be able to read logs :P

Victorious3:
Side note: This is so that I can do the following: 

```
Class transformerClazz = Class.forName("li.cil.oc.common.asm.ClassTransformer");
Method simpleComponentErrors = transformerClazz.getMethod("simpleComponentErrors");
Object buffer = simpleComponentErrors.invoke(null);
buffer.getClass().getMethod("$minus$eq", Object.class).invoke(buffer, "moe.nightfall.vic.integratedcircuits.compat.gateio.GPOpenComputers");
```

(Don't mind the reflection, I don't write fancy scala)
My proposal for adding a backdoor without having to worry about anything ;)

fnuecke:
Ah. Was thinking you'd gotten that under control with your own ASM black magic :P Well, in that case no, I'll rather add a new marker interface/annotation that can be used to suppress the warning. I'm assuming it still "works" because you're merging the 'template' class that implements that interface into actual tile entities before OC's transformer runs?

Victorious3:
@fnuecke Well, I'd still think its an improvement to know which mod was causing the issue in the first place without looking at the logs. Having a marker interface/annotation is cleaner but I personally wouldn't think its worth the effort, as what I'm doing here is clearly not supported behavior ^^
And yes, I'm injecting the interfaces before OC's transformer runs so its going to transform the TE class nevertheless.

fnuecke:
I think I still prefer the marker interface/annotation approach, in particular because it'd also suppress the warning in the log files. As for the naming, how about `SkipInjection`? Since that's basically what it'd have to do, not try to inject the logic in the template class, right? Probably an interface, because it can be stripped via `@Optional.Interface` (same reasoning `SimpleComponent` is an interface).

Victorious3:
If so, I'd make it an annotation.

fnuecke:
Ehh, above reasoning still makes me not favor that approach... then again, I suppose this would pretty much only be used for scenarios such as yours, meaning a class transformer is already involved, meaning it won't really matter anyway. Allrighty then.

skyem123:
Is there any way I can improve this pull request?

fnuecke:
Thanks, but it's fine, added the annotation so that should handle this. Needs testing, not sure if the desc string is correct (should be, but didn't actually test it myself).

Victorious3:
@fnuecke Thanks, I'll try it~

skyem123:
@fnuecke, well Victorious3/Integrated-Circuits@63e016e0b425332e1406eb3ef48345ab679698c6  works for disabling the message!
However, is there any way to make it require the correct opencomputers|core version?

Victorious3:
I tested it, disables the message but instead I get this:

```
[17:32:27] [Client thread/WARN] [OpenComputers]: Something went wrong!
java.lang.NullPointerException
    at scala.collection.convert.Wrappers$JListWrapper.iterator(Wrappers.scala:88) ~[Wrappers$JListWrapper.class:?]
    at scala.collection.IterableLike$class.exists(IterableLike.scala:77) ~[IterableLike$class.class:?]
    at scala.collection.AbstractIterable.exists(Iterable.scala:54) ~[AbstractIterable.class:?]
    at li.cil.oc.common.asm.ClassTransformer.transform(ClassTransformer.scala:137) [OpenComputers-MC1.7.10-1.5.13.662-dev-dev.jar:?]
    at net.minecraft.launchwrapper.LaunchClassLoader.runTransformers(LaunchClassLoader.java:279) [launchwrapper-1.11.jar:?]
    at net.minecraft.launchwrapper.LaunchClassLoader.findClass(LaunchClassLoader.java:176) [launchwrapper-1.11.jar:?]
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424) [?:1.8.0_25]
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357) [?:1.8.0_25]
    at java.lang.Class.getDeclaredConstructors0(Native Method) [?:1.8.0_25]
    at java.lang.Class.privateGetDeclaredConstructors(Class.java:2663) [?:1.8.0_25]
    at java.lang.Class.getConstructor0(Class.java:3067) [?:1.8.0_25]
    at java.lang.Class.newInstance(Class.java:408) [?:1.8.0_25]
    at moe.nightfall.vic.integratedcircuits.IntegratedCircuits.init(IntegratedCircuits.java:176) [IntegratedCircuits.class:?]
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_25]
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_25]
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_25]
...
```

(Yes I have to instance my own class with reflection in order to prevent the class loader from running too early...)

fnuecke:
@skyem123 Just require OC 1.5.13, since it's not released yet, that's good enough.

@Victorious3 Ah, fudge, I wanted to check if that can return `null` but forgot. Will fix.

fnuecke:
Done, should be fine now.

PR #1208, opened 2015-06-07
Grammar + Format

megamandos:
Minor grammar and format fixes for the English version of the manual.

### DIFF
--- a/src/main/resources/assets/opencomputers/doc/en_US/block/capacitor.md
+++ b/src/main/resources/assets/opencomputers/doc/en_US/block/capacitor.md
@@ -2,7 +2,7 @@
 
 ![It's over 9000.](oredict:oc:capacitor)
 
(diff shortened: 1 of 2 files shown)


megamandos:
Ah, when you look at it in-game, the "d" is on the next line. So it looked really weird. I'll fix this, give me a sec.

megamandos:
Fixed, the "bugger" is not intentional, right?

dgelessus:
> Fixed, the "bugger" is not intentional, right?

Yes, that's clearly a typo. ;)

> Ah, when you look at it in-game, the "d" is on the next line.

@fnuecke sounds like a bug, text shouldn't linewrap at a formatting change.

fnuecke:
Eh, the thing with the going to a new line is unfortunate indeed; adding a forced newline before it will be needed. It's not really a bug, since the formatting changes it's actually intended to allow the renderer to wrap there (fwiw, Github's Markdown rendering also would wrap that), e.g.
.....................................................................................................................................................`require`d.
^ The d is wrapped there for me, no space between the two. Anyway, adding a manual newline should solve this. If you'd like you could add that, then please squash the commits and I'll merge it.

megamandos:
Fixed and squashed.

fnuecke:
Thanks!

PR #1214, opened 2015-06-10
Handle a string parameter for keyboard.isKeyDown()

bawr:
keyboard.isKeyDown() accepts a string, but doesn't do anything sensible in the case a string is passed.

keyboard.pressedChars[] is always indexed by the unicode codepoint (maps to ASCII in the simple case), so passing a string currently always returns nil.

The proposed tweak changes that. Alternatively, we shouldn't be allowing strings in the first place, I guess.


fnuecke:
Hrm, the only things this change would "break" would have been broken in the first place, unless I'm missing something? (Since strings were indeed not handled, seems I had a brainderp when I wrote this)

The proposed change is kinda what it should be - for 5.2 anyway. For 5.3 we can use the utf8 lib. So I'd say change it to
`return keyboard.pressedChars[utf8 and utf8.codepoint(charOrCode) or charOrCode:byte()]`.

PR #1223, opened 2015-06-16
Added EC Integration

DrummerMC:


DrummerMC:
When you test this, pleace use the latest build from the buildserver http://shadowcity.net:8080/job/ecrv2/lastSuccessfulBuild/

In the latest version only the gas storage cells and the gas import bus are work

PR #1235, opened 2015-06-19
Plan9k loot disk

magik6k:
Here it is!
Plan9k is fully multitasking Unix like system. Here are some major features:
- Multitasking and multi-threading
- /dev and /proc filesystems
- Tape support, you can do `dd if=/dev/random of=/dev/tape1 bs=10000`
- EEPROM support, much like tapes
- Support for multiple screens with multiple GPUs
- Built-in networking
- Shell with pipe support(`cat foo.bar | wc -l`)
- vt100 like shell
- More

Note that some things are incompatible with OpenOS:
- You shouldn't use `component.gpu` anywhere, use vt100 codes instead
- Term lib is on top of io
- Autostart scripts are handled only by rc script system
- Few functions are still missing from libs, mostly from shell lib

And yes, some code is borrowed from OpenOS, with slight modifications to add neat features. The system bootstrap is more complex than OpenOS, First thing that is loaded is Openloader which loads (nano)kernel from /boot/kernel/pipes which then loads its modules from /lib/modules/base and starts them. Modules provide filesystem, userspace, IPC, multitasking, etc., after modules are started(start routine reaches threading), `/bin/init.lua` is executed. The kernel can be used standalone, that's why it's called 'pipes', 'Plan9k' is userspace name. The userspace init assigns screens to GPUs, starts shells on them and generally manages the system. All running processes have separate stdin/out/err settings, 90% of time they point to their tty(terminal/screen) that is handled by 'getty.lua' and 'readkey.lua' which provide vt100 like i/o(That's why term lib is built on top of io). 

Here are some features I plan to add in future:
- ssh/telnet(I have actually working version of it but it's unstable)
- Chroot
- Remote filesystem
- Network block devices
- Mount program(yes, with /dev/blockdevice support, including tapes)
- `top` with load metrics, etc.
- Text editor(I have no motivation so start working on this, I thought of compiling real `nano` to lua)
- Docs / man program


Vexatos:
How about checking for system requirements in `install.lua`, something like a warning when too little RAM is detected... I can imagine this using quite a bit of RAM, Tier 1 or 1.5 surely isn't enough.

Vexatos:
More requests: Add `_PROMPT` for `lua.lua` and `term.getCursor` for more OpenOS compatibility :palm_tree:

fnuecke:
Cool! Will give it a quick test run before merging it when I get home.

fnuecke:
Oh, right: maybe it could have a recipe? Dunno what would best fit. Manual is taken for OpenOS, Interweb for OPPM. Maybe crafting OpenOS plus [something] -> Plan9k, crafting Plan9k with [something] -> OpenOS (and expand that cycle as/if more OSes are added)? Something being... idk, something cheap. Redstone or so.

asiekierka:
Too OP, pls nerf

Vexatos:
@fnuecke and set the dungeon loot weight to 0 too, then.

Techokami:
Maybe the recipe could be OpenOS disk plus Tier 3 microchip?
This OS sounds exciting, can't wait to try it!

Sent from my Windows Phone

Vexatos:
The shell still needs a lot of work, though...

fnuecke:
> Too OP, pls nerf

Today's Modded Minecraft Drama (:tm:) - asie hating on OpenPeripheral! ;-)

> and set the dungeon loot weight to 0 too, then.

Of course.

> OpenOS disk plus Tier 3 microchip

That sounds way too expensive. The idea was to have all OSes at a somewhat samey cost, to avoid playing favorites ;-) Ideally it'd be crafting the OS disk on its own, but that's already reserved for wiping loot disks, so...

Vexatos:
@fnuecke The OS isn't close to done yet and still lacking a lot of convenience features OpenOS has (also it's still lacking edit, you can only move the cursor with the arrow keys in shell/lua interpreter etc.); there should be some note that this may or may not be for more advanced users.

Techokami:
> That sounds way too expensive. The idea was to have all OSes at a somewhat samey cost, to avoid playing favorites ;-) Ideally it'd be crafting the OS disk on its own, but that's already reserved for wiping loot disks, so...

Okay, I had an idea. We add another manual-type book about general Lua programming, for newbies. They can read the book to learn a good deal of Lua basics, which will make developing programs for drones and microcontrollers less spooky. To craft it, you craft an OpenComputers Manual by itself (which is how you can craft the other manuals in Tinker's Construct, fyi).
Then you just craft this new book with a blank floppy to get a Plan9k disk.

gjgfuj:
A lua tutorial book would be very good, I think. although it should probably just be part of the existing manual, the system is already there.

Techokami:
> although it should probably just be part of the existing manual, the system is already there.

Well, making it a separate book was to solve the crafting recipe problem for the Plan9k disk while not making one disk more expensive than the other...

Another potential idea: craft blank floppy with Lua BIOS EEPROM, but do not consume the EEPROM itself. You kind of need to make one, anyway...

gjgfuj:
Yeah, but that would only work for one os. What if we have more oses that
are distributed. More books?

Vexatos:
I think we should at least wait for the OS to get a proper editor before this gets merged.

fnuecke:
Just gave this a shot, and... whenever I try to run `ls`, for example, it just stops and sits there. Cursor blinking on a new line to the left, but not accepting input, no output either. When I try to run a program that doesn't exist (tired `pwd`) I get the error message, but then the same hang. `install` worked, though! Buuut same issues with it sitting on an r/w fs.

And yeah, some kind of editor, even if it's just a line editor, seems... kinda obligatory. Just using `echo`, `cat` and co to edit files is somewhat painful :P

As for the whole "new book" idea... aside from not really liking the idea, to be honest (I also think that should just go into the normal manual), who's supposed to write that? :P

Techokami:
How about... crouch + use while holding an OS disk will cycle it through a list of OS disks? It's really hard to tackle the issue of making OSes equal in crafting costs...

Vexatos:
Sneak+right click wipes the disk, IIRC.

Techokami:
OpenComputers Manual + Paper = OS Source Code
It's a stack of papers that are confusing to read (and trying to use the item as-is gives the player Nausea, maybe?) but sneak + use switches the item to a different sub-type, representing a different OS that can be crafted.
OS Source Code + Blank Floppy = OS Disk for that OS

I think this hits all the criteria: It creates a common cost for making all OS disks, it's not too expensive (adds one paper to the existing recipe, which is a cheap and renewable item), and it can be expanded upon for more than two OSes.

fnuecke:
Doesn't wipe anymore, so that could be an option. Just sucks that it'd need a new system to mark OS floppies, unlike reusing the existing crafting/recipe system.

Vexatos:
I don't think this should have a crafting recipe at all, to be honest. It's still far from finished, and even once it is, it will be available through mpt (so indirectly via OPPM) AND as a loot disk. It's a fairly complex OS and I am sure providing an easy way to acquire it would confuse a lot of people mixing it up with the "normal" OS.

fnuecke:
I can't judge how complicated to work with it could be, since, as noted above, I can't even get `ls` to run, but anyway. I'll wait a bit more with merging, to allow some more testing and working out potential issues. Still think having a recipe or such for it would be nice, to offer some alternative to the default that is OpenOS.

Vexatos:
It is fairly complex compared to OpenOS, simply due to all the additional features.

Techokami:
Played with it a bit. So far so good, but it needs work still. For starters, you forgot the shutdown program.

fnuecke:
`ls` (or pretty much anything else) still doesn't work for me:
http://gfycat.com/SilverImpossibleLaughingthrush
Anything I can provide you with that'd make debugging that easier? FWIW, I did some hammering on the keyboard at the end there, to no avail. It's just dead.

Techokami:
I managed to get `ls` working on my test system by installing the OS first. But it not working on a bootable floppy _IS_ an issue.
But your version of `edit` does crash when I try to edit a nonexistent file. It brings up the following error:

```
[dmesg] ERROR IN THREAD 8: /bin/edit.lua
[dmesg] /bin/edit.lua:40: attempt to perform arithmetic on a nil value (local 'h')
[dmesg] stack traceback:
  lib/modules/base/20_threading.lua:62: in function <lib/modules/base/20_threading.lua:59>
  (...tail calls...)
  /bin/edit.lua:40: in upvalue 'exec'
  lib/modules/base/20_threading.lua:58: in function <lib/modules/base/20_threading.lua:57>
  [C]: in function 'xpcall'
  machine:738: in global 'xpcall'
  lib/modules/base/20_threading.lua:57: in function <lib/modules/base/20_threading.lua:55>
```

Is it because I was trying to edit a nonexistent file? If so, you should make `edit` error if trying to access a nonexistent file, and add `touch` to create new files.
But you really should copy over `shutdown` to your new OS so I can gracefully turn the system off!

magik6k:
the ls issue seems to be related to read only rootfs, trying to fix that now. Also looking at edit issue

fnuecke:
Allrighty, works well enough for me, and it can always be updated via the interwebs ;-) (Or pull requests :P)

Will see about a recipe sometime. Let's get it out there for now!

Issue #1265, opened 2015-07-05
quiet preceding commands in pipe never resume subsequent commands


payonel:
This is similar to issue #1263 which was closed by commit b60c784
In this case consider: p1 | p2
where p1 does a read, and no output, and p2 has output and then a read
Well, because p1 gets to a non-piped read, it waits for user input and p2 never is resumed

I actually have a fix...which also would have resolved issue #1263 with less code!
I'll make the PR soon and reference this issue

fnuecke:
Closed via 6683a5bd493297eb8eb978e6c33331c5e5dbdede. Thanks again!

PR #1266, opened 2015-07-05
fix piped quiet commands from stalling the pipe chain

payonel:
appending an empty string to the memory stream buffer is safe because any subsequent calls to read will yield correctly. Also, this ensures that quiet commands pass the next command argument pack (no longer needing to use memoryStream.empty)

Also, this resolves issue:  #1265

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -13,9 +13,6 @@ local unicode = require("unicode")
 local memoryStream = {}
 
 function memoryStream:close()
-  if (self.empty) then
-    self:write('')
-  end
   self.closed = true
 end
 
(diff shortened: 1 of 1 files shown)


payonel:
i've tested this more, I feel good about this change. This even causes piped commands to roll and unroll in the same order as in bash.

Also, reads still close early identically to bash
consider something like:
file: p1
print(os.getenv()._ .. " pre-read")
local v = io.read('*l')
print(os.getenv()._ .. " post-read: " .. tostring(v))

and then cp p1.lua to p2.lua and p3.lua
then: ./p1 | ./p2 | ./p3
/home/p3 pre-read
/home/p2 pre-read
/home/p1 pre-read
text
/home/p1 post-read: text
/home/p2 post-read: nil
/home/p3 post-read: nil

and this is perfect

payonel:
combined with another PR, this one now closed

Issue #1267, opened 2015-07-05
piped commands that end early break preceding commands from further output


payonel:
p1 | p2
where p1 outputs multiple strings but p2 reads only 1 and ends (coroutine dead) causes p1 to set its memoryStream.closed = true
subsequent calls from p1 to io.write will error that you should not write to an closed stream
however, p1 should be able to write all it wants - and the values should be thrown away (not buffered)

fnuecke:
Closed via 6683a5bd493297eb8eb978e6c33331c5e5dbdede. Thanks again!

PR #1268, opened 2015-07-05
ignore all writes to a closed pipe stream if next is dead

payonel:
resolves issue:  #1267


Techokami:
Uh, can you squash everything into a single PR?

payonel:
I combined two PR's into one, this one is no longer needed.

PR #1269, opened 2015-07-05
ignore closed pipe writes to dead coroutines and fix stalled pipes

payonel:
ignore all writes to a closed pipe stream if next is dead and fix piped quiet commands from stalling the pipe chain

resolves issue: #1267
resolves issue: #1265

this is a merge(a squish) of two previous PR's:
https://github.com/MightyPirates/OpenComputers/pull/1266
https://github.com/MightyPirates/OpenComputers/pull/1268


fnuecke:
I'll just believe you that this works :3 Thanks!

PR #1272, opened 2015-07-06
Corrected 2 small mistakes

Ionaru:
"will contain of the components" -> "will contain all of the components" on line 7
Added a comma on line 13, action -> result.


fnuecke:
Thanks!

PR #1290, opened 2015-07-11
Fix ProjectRed dep and make hollow covers close on cables.

Kubuxu:
It looks like this:
Hollow covers

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,7 +106,10 @@ repositories {
         name 'DVS1 Maven FS'
         url 'http://dvs1.progwml6.com/files/maven'
     }
-
+    maven {
+        name 'ProjectRed'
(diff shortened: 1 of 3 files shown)


Kubuxu:
Hmm, it is building on my machine.

makkarpov:
The issue of Travis build is that ProjectRed Maven domain is `files.projectredwiki.com`, not `projectredwiki.com` anymore.

Kubuxu:
It is changed. PR requires CoFHCore which is cached on my machine.
Fixing it now.

Kubuxu:
Might have fixed that, I have to wait for Travis build as cache won't allow to check me on my machine.

fnuecke:
Thanks!

PR #1308, opened 2015-07-18
Given a smaller buffer to output

RyanSquared:
I accidentally tried to write 15KB of text to the terminal. This would at least somewhat fix it.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cat.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cat.lua
@@ -19,6 +19,7 @@ else
       local blob = file:read(1024)
       if blob then
         io.write(blob)
+        os.sleep( 0 )
       end
     until not blob
     file:close()
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cat.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cat.lua
@@ -16,11 +16,11 @@ else
       os.exit(false)
     end
     repeat
-      local line = file:read("*L")
-      if line then
-        io.write(line)
+      local blob = file:read(1024)
+      if blob then
+        io.write(blob)
       end
-    until not line
+    until not blob
     file:close()
   end
 end


fnuecke:
Hmm, the problem I see with that, is that it can break unicode chars apart. It's also just fighting the symptoms at a very high level.

It'd probably be better to have the term split output apart using unicode.sub before calling whatever hangs (probably text.detab or text.wrap used in term.write). Also fighting symptoms (that being that the matching reimplemented in Lua - to avoid evil patterns locking down a worker thread - is quite slow), but slightly more generic.

dgelessus:
As for incomplete UTF-8 sequences:
1. Use `utf8.charpattern` to ensure that the last char is valid UTF-8.
2. Not all files are text, and not all stdout is the screen. There are cases where splitting UTF-8 sequences doesn't matter, or where the data is not UTF-8 text.

fnuecke:
1. `utf8` is only available in 5.3 though.
2. So you agree it'd be best to handle this in the term/text libs, since they are meant to display/work with text?

dgelessus:
1. The charpattern is just a normal string, there should be no issue with something like this:

``` lua
local charpattern = utf8 and utf8.charpattern or "..."
```
1. I don't think this is something `term` could handle alone - what should it do when it gets half a UTF-8 sequence? Wait for the next string? And what happens with invalid UTF-8?

fnuecke:
1. Eh, sure. But I don't quite see the desire to manually check that when there's `unicode.len` anyway.
2. I think it'd be fair to expect callers of `term.write` not to pass corrupt strings in.

(Woah. `0.` is auto numeration. TIL)

RyanSquared:
We could read everything up to the last character (something like `:match("^(.-)<CHARPATTERN>$")`) and then have the next read append to that?

RyanSquared:
I think the major problem with this, especially with `cat`, is that it tried to read() too much. Is there any kind of error handling now for this?

fnuecke:
@payonel will the reworked cat also have this problem / has your rework of OpenOS touched buffers to possibly fix it?

payonel:
essentially no, there are no yields added that would resolve this by my rework

payonel:
I'll fix this with another update to term to check uptime and sleep on long writes.

PR #1312, opened 2015-07-19
Advanced data cards (Tier 2 & 3)

makkarpov:
Tier 2 additional functions:
- MD5 and SHA256 can act as HMAC's if you pass them key parameter (call limit = 32 as in Tier1, energy consumption at _simple_ level)
- SecureRandom data generation (call limit = 2, energy consumption at _simple_ level)
- AES-CBC-PKCS5Padding encryption and decryption (call limit = 2, energy consumption at _complex_ level)

Tier 3 additional functions: (energy consumption at _complex_ level)
- EC key pair generation (256 or 384 bits) (call limit = 1)
- Key exporting and importing (call limit = 4)
- ECDH key agreement support (call limit = 1)
- ECDSA signature support (call limit = 1)

**What is to do**
- ~~Textures (I'm not very good at it)~~
- German translation for Tier 2 & 3 tooltips (now these tooltips have identical strings)
- ~~Crafting recipes. I have no idea about components these card should consist of.~~

**What is not implemented**
- Streaming support due to fact that Java ciphers/hashes/deflaters/signatures cannot be persisted.


Vexatos:
Well, @fnuecke definitely needs to apply the balancing we were talking about, if this PR gets merged.

makkarpov:
Maybe you should just write these balancing concepts? I din't talk in IRC, so I just do not know it.

Vexatos:
@makkarpov Don't worry, if it gets merged I am sure Sangar will do it himself.

Kubuxu:
@Vexatos if you could write it here it would be great. Everybody would like to participate in this discussion. Not only 3 people that were on IRC on that time.

Vexatos:
I don't know how it would be applied to other stuff, but the idea was for AES to consume 0.1 OC units per character in the string to encrypt/decrypt plus an additional 16 OC plain usage; also, the methods would not be direct but synchronized to take 0.05 seconds.

makkarpov:
@fnuecke said that these methods will be direct, but with limit = 1 that is effectively same as synchronized. 

I think that 0.1 OC/char + 16 plain is too big cost. Deflate costs 4 + no data cost, and runs slower than encryption (at my benchmarks - 3x slower).

Kubuxu:
Sangar also said that it was too high.
In comparison Linked Card is `0.03125` per byte.

makkarpov:
Took a [benchmark](https://gist.github.com/makkarpov/0358c99e43909d9ea960) of all algorithms used in data card. So here is my proposal of limits and costs:

| Algorithm | Call limit | Base cost | Per-byte cost |
| --- | --- | --- | --- |
| Base64 | 32 | 0.2 | 0.005 |
| CRC32 | 32 | 0.2 | 0.005 |
| AES, MD5 | 8 | 1 | 0.01 |
| EC key importing | 8 | 1 | — |
| Deflate, SHA256 | 4 | 6 | 0.1 |
| SecureRandom | 1 | 6 | — |
| EC key generation | 1 | 10 | — |
| ECDSA | 1 | 10 | 0.1 |
| ECDH | 1 | 10 | — |

makkarpov:
So at present time limits in the table and in the code matches. Any comments or suggestions?

Kubuxu:
Those are my recipes proposals:

```
dataCard2 {
  input: [[nuggetGold, "oc:materialALU", "oc:circuitChip3"]
          ["", "oc:dataCard", ""]]
}
dataCard3 {
  input: [[diamond, "oc:cpu2", "oc:circuitChip3"]
          ["", "oc:dataCard2", ""]]
}
```

makkarpov:
Now this card looks good. I can write manual page about it's API, but I don't sure that it is necessary.

Kubuxu:
Just now, me and @magik6k understood how awesome is EC's shared key with AES.

Expect StarCoin soon.

makkarpov:
I think that now this card is completed, maybe with only one possible improvement - I'm not good at texture making. @fnuecke what do you think about it?

fnuecke:
I'll look into it when I can. In the meantime, please squash those commits :-)

makkarpov:
Done, squashed.

Kubuxu:
`data.random(128)` causes too long without yelding. Probably due to missing entropy, you should probably use local instance SecureRandom.

makkarpov:
Switched to `ThreadLocal` SHA1PRNG instance. It also works much faster than `getSeed` one (as benchmarks shows 0.004ms per 128 bytes instead of 1.463 ms) thus allowing me to enforce weaker limits.

But if there is not enough entropy, it will still block at initialization, but, I believe, this condition is rare at least on Linux systems.

Kubuxu:
Also better error in case of wrong ec type during de-serialization would be good.

I've spent 20min scratching my head over `ec_public` instead of `ec-public`.

fnuecke:
All right, all looking good I think. I did rename the T1 card internally (dataCard -> dataCard1) for consistency, and existing cards _should_ get converted seamlessly (did in my testing), but some more testing/verification of that not acting up would be appreciated.

PR #1336, opened 2015-07-26
Enable Travis caching feature

makkarpov:
... because currently builds run for 11 minutes just to download all dependencies.

[Travis docs](http://docs.travis-ci.com/user/caching/)


fnuecke:
My lack of travis know-how shows :P Thanks.

makkarpov:
Unfortunately I was unable to test it because Travis seems not to run on forks. But I think that it will not break anything.

PR #1347, opened 2015-08-02
Translate unmanaged/managed modes

makkarpov:


makkarpov:
Also, Russian version refers these modes as 'file mode' and 'block mode'. Maybe this naming is more descriptive than 'managed'/'unmanaged' one.

cyber01:
Опередил :(

fnuecke:
Thanks, as for the names... we'll see. I kinda like them like that, but maybe I'll make a poll before release.

PR #1360, opened 2015-08-09
Add parameter to disable resolution reset

makkarpov:
I have setup with one GPU and five screens. I use `bind` method to switch between screens and draw on them. But when I switch GPU to screen, screen parameters are reset, which causes noticeable flickering before I set them to correct values.


skyem123:
This seems useful!

fnuecke:
Hmm, originally this was mostly to discourage using one GPU to control _all the screens_, and partially because stuff usually won't know what's being displayed on the newly connected screen, so it'd need to be cleared anyway, IIRC; but given that rebinding isn't that fast anyway... I guess why not.

PR #1370, opened 2015-08-14
Change usage of gitref in versioning

Kubuxu:
As git ref hash can't be used to order versions it should be only used to denote differences between them.
By using plus sign instead of hyphen many version compression scripts and similar will stop confusing custom build versions.
It is also then compatible with SemVer version 2 ordering rules as for points 10 and 11.
It also was moved at the end so it is more obvious that it a metadata.
http://semver.org

It won't break any of dep configurations as it is only used in unofficial builds.


fnuecke:
Ah, right. I actually knew that :X Thanks!

PR #1394, opened 2015-08-30
Update plan9k

magik6k:
Changes:
-Fixed df human readable crash
-Fixes to require
-Fixes to tape drive driver

-Added delete key functionality to edit
-Added ifconfig human readable output
-Added the more program
-Added /dev/klog, some other fixes
-Added /proc[procid]/limits, status
-Added /dev/urandom using data card
-Added unmanaged disk driver!
-Added some datacard tools

-Better kernel log management

I thought of adding fat12/16 driver by gamax92 and some utilities around that, but it still doesn't seem 'stable enough' for me. If you want it I may add it tho, it won't break things out of the box


Vexatos:
Is it a tape drive driver or just a tape driver? :suspect:

Techokami:
> I thought of adding fat12/16 driver by gamax92 and some utilities around that, but it still doesn't seem 'stable enough' for me. If you want it I may add it tho, it won't break things out of the box

Add it now, update it later?

fnuecke:
Oh, missed that last sentence :X Hrm, those that'd use it probably will be able to update to experimental without the loot disk, so adding it when it's stable sounds better to me.

magik6k:
Yup, installing experimental can be done from the floppy+internet card with `mpt -S plan9k mpt --root=/mnt/[disk]`, fat driver is in plan9k-filesystems package IIRC

Issue #1402, opened 2015-09-04
Not Enough Keys API is undownloadable


chaoskagami:
The new source for their build tracker/maven is apparently here: http://ci.dmodoomsirius.me/maven/notenoughkeys/NeK/ 

See here for another issue on this: https://github.com/BuiltBrokenModding/ICBM/issues/107

This is preventing me from actually building OC from source. I'm the kind of person who likes to build things myself if source is available, so this is kinda an issue. I don't care if I have to fix it and submit a PR, but this should be resolved sooner than later. Referring to non-existent URLs is bad.

Y'know, I'll just submit that PR since it takes literally one sed to fix it.

dmodoomsirius:
check out http://docs.dmodoomsirius.me/v1.0/docs/notenoughkeys on further information @fnuecke

chaoskagami:
We're apparently behind by a build now. Stop moving the build server around, @dmodoomsirius 
mmk? ;P Thanks, though.

Anyways, I'm making another fix to repoint to the proper build server (again.) @fnuecke, when I finish in a bit I'll submit another PR. I'm going to make sure it builds first.

dmodoomsirius:
Thats why you apply the apply from: 'https://raw.githubusercontent.com/BuiltBrokenModding/BuiltBrokenScripts/buildScripts/scripts/both/repos.gradle'
That gives you access to a bunch of repos as it is.

As i will always keep that one updated @chaoskagami

chaoskagami:
Opencomputers specifies it's own dependencies. Chances are there's a good reason for that. Usually, there is.

Anyways, I'm submitting a PR with the new new repo api.dmodoomsirius.me. Things had to be expanded a bit due to the maven manifest being located down a few directories in a package way, but otherwise assuming the repo isn't changed again we should be good. Compiles cleanly, all is good now.

dmodoomsirius:
Repos.gradle is nothing more then a massive repository for maven repos.

dmodoomsirius:
and the api.dmodoomsirus.me if i change the url i will change the API subdomain ip so it won't break.

chaoskagami:
It's a good move to separate the repos from the actual build script.

I'm just not too familiar with gradle and maven as build systems. I'm trying to keep my changes as noninvasive as possible.

dmodoomsirius:
@chaoskagami Are you on irc at all? if so pm dmod lets talk there so we stop spamming sangar.

PR #1403, opened 2015-09-05
Fix for NeK maven being removed

chaoskagami:
NotEnoughKeys' current maven at ci.builtbroken.com is gone. Bug #1402 which I just posted covers it. This was an issue (or at least similar) on the ICBM repo over here: https://github.com/BuiltBrokenModding/ICBM/issues/107

This commit replaces the repo with the new one and changes the version code since b35 is not on the new one - it appears to have restarted build numbers.

I have no idea if this actually builds, but considering it doesn't build without NeK, it's the same issue either way. It'll still build with the jar cached obviously, so this probably isn't immediately apparent unless you nuke ~/.gradle, this repo and resync.


fnuecke:
Tested and works, thanks!

PR #1407, opened 2015-09-08
Update to the proper final repo for NeK. Or so I hope, at least.

chaoskagami:
@dmodoomsirius commented on the previous bug report #1402  which was closed - apparently api.dmodoomsirius.me is the new new maven, and should be what we use. The one in my previous PR was a stopgap repo while stuff was moved, I think.

The version of NeK on the proper maven is two ahead now, b2 -> b4, not sure how much has changed but the version number is what was in use before the previous builtbroken maven died. OC should build without issue.


fnuecke:
Well, travis certainly seems to still build, so here goes... :P 

Thank again.

chaoskagami:
No problem. Just helping out a mod that I find really fun to use. ;P

PR #1419, opened 2015-09-12
Add password-protected debug cards

makkarpov:
- Config property `enableDebugCard` was renamed to `debugCardAccess` (does not renames value in existing configs yet)
- `true` and `false` are treated like before - completely enables/disables debug card. String values are treated as a password that must be passed to `debug.unlock(string)` to enable all other debug card functions.

Change is intended for servers where people may cheat in items but admins need debug card.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1358,9 +1358,10 @@ opencomputers {
     # Enable debug card functionality. This may also be of use for custom
     # maps, so it is enabled by default. If you run a server where people
     # may cheat in items but should not have op/admin-like rights, you may
-    # want to set this to false. This will *not* remove the card, it will
-    # just make all functions it provides error out.
-    enableDebugCard: true
+    # want to set this to false. Set this to string if you want to allow
+    # debug card usage to those who knows the password. This will *not*
+    # remove the card, it will just make all functions it provides error out.
+    debugCardAccess: true
 
     # Whether to always register the LuaJ architecture - even if the native
     # library is available. In that case it is possible to switch between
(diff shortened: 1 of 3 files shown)


fnuecke:
Interesting idea. I'll look over it when I can.

One idea to ponder though: I'm not super-keen on actually security-relevant passwords in-game. Maybe prefer having a whitelist of players (name or UUID) and require the machine to only have registered users in that whitelist (but at least one). To get that list, `context.asInstanceOf[Machine].users` should work, I think (better as a match to avoid errors when it doesn't, but in principle).

makkarpov:
1) You write malicious program that use debug card
2) You write script that adds admins using `computer` API and runs program

All conditions are fulfilled - only admins are registered for this computer, but you have malicious script running on it. You do not have access to this computer anymore, but you have access to debug card and can, for example, listen network for incoming commands and execute them.

---

Another approach is to somehow _activate_ debug card with shift-rightclicking by admin with card in hand. But with this approach it's hard to revoke already activated cards. (But it can be integrated into existing password system - right-clicking by admin will fill in correct password into the debug card state)

fnuecke:
Ah, fair point, that'd work.

I like the idea of using the binding system. Basically allow using bound debug cards, if the owner is in the whitelist. Though losing such a card would be an issue, true. If we could make that somewhat less problematic, I'd really like this approach.

makkarpov:
To revoke bound cards we can attach some `nonce` field to each whitelist entry, so card is active only if owner is whitelisted and nonce matches. So revoking a card == changing a nonce for specific user. But it will require some other file than config.

makkarpov:
Will rework and create separate PR for new code.

PR #1422, opened 2015-09-14
Sh patch

payonel:
This changes tab completion for search items in the sh shell. 

For file name completion, the path as entered by the user is unaltered visually
e.g.
if pwd has a file called re.lua and read.lua
./re{tab}
resolves to
./re.lua
and then {tab} again resolves to
./read.lua

previously (i.e. without this change) the {tab} would complete to:
/path/to/pwd/re.lua
and
/path/to/pwd/read.lua
respectively

The goal here is to preserve the path string (which may contain . or ..) the user has entered and not replace it with the results of shell.resolve (visually)

Also, single match completions will append a space after the file name, indicating to the user that they have correctly spelled the name of a file, uniquely

e.g. same pwd as above
./read{tab}
resolves to
./read.lua{space}

But not for directories, to allow the user to continue typing a path (as before)

Additionally for both program search and path search, if the path contains magic characters previously the tab completion pattern matching would fail. Thus, this patch contains a method to escape all magic characters before pattern matching is run on user input.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/sh.lua
@@ -384,6 +384,10 @@ end
 local args, options = shell.parse(...)
 local history = {}
 
+local function escapeMagic(text)
+  return text:gsub('[%(%)%.%%%+%-%*%?%[%^%$]', '%%%1')
+end
+
 local function getMatchingPrograms(baseName)
   local result = {}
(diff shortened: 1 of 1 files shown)


fnuecke:
Sound good (finally someone got annoyed enough to take a look :P). I'll have a look in the evening.

If you could squash the commits, that'd be great.

fnuecke:
Took me a bit longer than promised, sorry. Looking good, thanks! :)

PR #1424, opened 2015-09-14
Added transposer to manual index.

Vexatos:


fnuecke:
I always forget this. _Always_.

PR #1432, opened 2015-09-17
Change EnderIO wrench integration to latest EnderIO API.

mrammy:


Vexatos:
Which version of EnderIO is this? If it is later than 2.2.8, it shouldn't be supported yet due to being considered _beta_.

mrammy:
It is beta. I guess I could make a version check for each one separately or this could just get pulled later. I thought the master branch was for "latest" or "unstable" features. If I was incorrect, my apologies.

Vexatos:
This PR should be accepted once EnderIO 2.3 is out of beta.

payonel:
@Vexatos looks like it is out of beta for 2.3 - should we still merge this?

Vexatos:
This is for the 1.7.10 branch, not 1.10.2. However, the 1.7.10 version of EnderIO is not being worked on anymore and won't ever get out of beta. The 1.10.2 branch already has this change, so I guess it is up to @fnuecke.

payonel:
@mrammy I've tested with and without this change and I can use enderio's wrench in place of oc scrench on oc blocks, and i can use oc scrench in place of enderio wrench on enderio blocks. I'll close this PR for now, but reopen if you know of a workflow broken without this PR

PR #1435, opened 2015-09-18
Update en_US.lang

Vexatos:
> Vexatos's nanomachines went out of control.

Not that good of a death message.


vifino:
Boo.
I like the one that's here right now. :3

fnuecke:
It'd have been more uniform, but this is clearly better because it can't be wrong :P

PR #1437, opened 2015-09-18
Made OC not crash on startup

Vexatos:


fnuecke:
Pfft, not crashing is overrated :P

dgelessus:
What, source files that don't end in a newline crash the game now? Rightfully so, I say! `:P`

Vexatos:
Blame GitHub's in-browser editor for that one.

PR #1455, opened 2015-09-22
Add Rotarycraft Integration (item stack converters for jetpacks and …

Voidi:
…springpowered pumps)


fnuecke:
Looking good, will test soon (probably tomorrow), thanks!

I'm assuming you're using `GameRegistry.findItem` because RC doesn't have a more specific way of looking up items in it its own API?

Voidi:
Have you updated the API? Last time i was looking it was still the old version with Powerthings only.
And for the pump item there is no API anyway.

fnuecke:
I haven't had a look at all, was just wondering blindly ^^

fnuecke:
Thanks; feel free to give the next dev build with this in it a shot, to make sure it still works :)

PR #1461, opened 2015-09-25
fix tab complete bug when using ./ as path

payonel:
Sangar - this fix is pretty important. In my tab complete improvement there was case I missed that causes a script failure in sh.lua - that is, when tab completing ./
I've tested many variations of tab completion. I feel good about the state of things with this.


fnuecke:
Thanks!

PR #1466, opened 2015-09-29
make cd.lua nearly POSIX compliant

payonel:
This submission omits support for -L and -P options for cd which specify whether to resolve sym links logically or physically, respectively. We rely on the openos filesystem for symlinks, which is only partially supported (e.g. not-persistent) and thus I feel we should omit explicit support for these symlink related options to cd - and just allow shell.resolve and shell.setWorkingDirectory to handle the path resolution - which from testing appears to follow all symlinks logically (which is logical from the user perspective - so a good default).

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cd.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cd.lua
@@ -1,11 +1,105 @@
 local shell = require("shell")
+local fs = require("filesystem")
 
-local args = shell.parse(...)
+local shell_name = '-' .. fs.name(os.getenv("SHELL"))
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks! I'll have a read and test in the evening :)

payonel:
I'll be doing about 1 binutil a night. I remembered after I submitted this PR that you like to squash commits. I'm happy to make a new dev branch named binutils and submit a new PR when I've done the bulk of my planned binutil changes. Or if you want, you can take them one at a time.

fnuecke:
My rule-of-thumb for squashing is if it's a functional/logical unit, squash it. So having the programs come in as separate (squashed ;)) PRs sounds good.

PR #1473, opened 2015-10-04
alias change to use equal sign and allow list of input

payonel:
support examples
alias k=v w x y a=b
would:
1. create/overwrite alias k to equal v
2. print alias values of w, x,  and y (if any don't exist, print error)
3. create/overwrite alias a to equal b
also, just `alias` prints all current alias keys and their values


fnuecke:
Looking good; thanks! I think this'll affect few enough users to merge it in now (plus those that use the aliasing should be able to easily revert to the old implementation :P).

PR #1477, opened 2015-10-08
Add backtick quoting to text.tokenize

dimitriye98:


Vexatos:
Where would anyone use this type of quoting? It's certainly not valid Lua...

dimitriye98:
It's valid shell script though, so it's useful in tokenizing shell input

fnuecke:
Aye, and that tokenizer is mainly intended to be used by the shell, so that makes sense I think.
Since I can't test right now, do backtick "quoted" strings actually ignore backslash escaping usually?

Will test the PR when I can!

dimitriye98:
I'm closing this pull request as I've submitted one with more complete logic.

dgelessus:
@Vexatos There are some people (mostly manpage writers) who are still stuck in the 70s and think that the proper way to write quotes is `like this'. `;)`

dimitriye98:
@dgelessus Actually, it won't parse that. I added it because ``commands here`` is shell script for a command substitution, wherein the commands inside are evaluated with their output being substituted in its place.

PR #1478, opened 2015-10-08
Add more complex tokenization logic

dimitriye98:
Now operators will be parsed closer to POSIX spec, that is, any valid operator will be tokenized regardless of being delimited by spaces. I've left out the & operator as as far as I can tell, background processes are impossible.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/text.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/text.lua
@@ -81,7 +81,7 @@ local quotes = {
 --["$(("] = "))"
 }
 
-local function tokenize(value)
+function text.tokenize(value)
   checkArg(1, value, "string")
 
   local i = 0
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/text.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/text.lua
@@ -70,85 +70,137 @@ end
 
 -------------------------------------------------------------------------------
 
-local operators = {";", "&&", "||", "|"}
-local function checkOp(string)
-  for _, v in pairs(operators) do
-    if unicode.sub(v, 1, unicode.len(string)) == string then
-      return true
-    end
-  end
-  return false
-end
+local quotes = {
+  -- array part must be ordered from long to short (irrelevant for now)
+  '"', "'", "`";
+  ['"']   = '"',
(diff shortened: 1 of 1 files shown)


fnuecke:
Hrm, operators seem to have a higher priority than quotes now?
E.g. `text.tokenize('a | "b|c"')` now generates `a, |, "b, |, c"`, the current `a, |, "b|c"` seems more correct (and looks like what my real shell does, too, since it complains about not finding `b|c`).

dimitriye98:
Ah, sorry, this code is actually a snippet from another project, and I forgot to fix that in this version, that'll be resolved shortly.

payonel:
@dimitriye98 Actually, the new /lib/text.lua in my PR (I've been overhauling openos) has a lot of text parsing upgrades (I also support new delimiters for special purposes, like ||, &&, ;, and correctly escaped quotes (as per POSIX again). And good news, spaces are not required around delimiters with it

payonel:
@fnuecke do we WANT tokenize to return more tokens? My new tokenize method takes extra params, one in particular which asks doNotNormalize. 

Legacy calls to tokenize will see the SAME results. With the new tokenize, text.tokenize([[a|"b|c"]]) is the same as before, i.e. { [[a|"b|c"]] }

But text.internal.tokenize([[a|"b|c"]]) (equivalent to calling text.tokenize(value, true), would tell more complex story (which sh parsing needs, i need to know which parts were quoted, and which parts were not -- such as for correct globbing
The tokenize result in this example is (with or without whitespace outside of quotes):
{{{txt="a"}},{{txt="|"}},{{qr{"'","'",true},txt="b|c"}}}

@dimitriye98 The nonnormalized tokenized table result I give above is complex (I know), but it must be this way because I continue to use that metadata all throughout sh parsing. For example: ech"o" is one word, and resolves to echo, and likewise ech"|" is not two parts: "ech" and a pipe. Rather, it is the WORD ech|. In bash you'd get: ech|: command not found. I don't really intend for users to use this complex result directly (I consider it internal api to text and sh), but tokenize was the best place to share the same logic.

I could adjust the text.internal.normalize method and keep delimiters separated. sh parsing does not use normalize, it is there to scrub the metadata before returning tokens to legacy calls.

payonel:
We can close this now, openos 1.6 has this functionality
@dimitriye98 thanks for your interest in this area. feel free to do some manual testing in this area in the latest 1.6 builds.

PR #1481, opened 2015-10-12
disk usage bin utility

payonel:
just providing basic options: version, help, summary, and human readable


fnuecke:
I'm just going to believe you that it works ;-)

PR #1503, opened 2015-10-25
Able to control RTM(Real Train Mod) signals with OC

PotatoOMG39:
Is that possible? (Im tired of pulling every lever to change the signal.) I can make the siganls automatic with OC.


Vexatos:
Uhm.....

makkarpov:
The most useful PR ever.

fnuecke:
I'm not sure what this is, but if you want to suggest a feature, open an issue, don't make a (seemingly pointless) PR. Also give a better explanation of the feature you're suggesting, I'm don't quite understand what you mean. Because it sounds like you want to automate something that uses levers, which indicates you could just use redstone signals, which is already quite possible.

PR #1532, opened 2015-11-18
Added term windows

mpmxyz:
Added API methods:
- term.setWindow(newWindow) -> oldWindow
- term.getWindow() -> currentWindow
- term.newWindow(x, y, width, height, gpu) -> window
- term.setArea(x, y, width, height) -> oldX, oldY, oldW, oldH
- term.getArea() -> x, y, width, height
- term.setGPU(newGPU) -> oldGPU
- term.getGPU() -> (gpu), (gpu or component.gpu)
- term.getGlobalArea() -> x, y, width, height
- term.toLocal(x, y) -> x, y
- term.toGlobal(x, y) -> x, y

Just to document some parts of the code:
- general: All original term functions have been converted to window methods; all term functions except newWindow, setWindow and getWindow forward to the current window.
- line 9: The table "blinking" is needed to store windows for the toggle timer without making them uncollectable for the garbage collector.
- line 59/line 567: term.read now only listens to keyboards connected to the used screen. But since screen.getKeyboards is a very very costly function it is cached by the table knownKeyboards.
- line 93: Window objects only contain a hidden __state value. The methods are added by a metatable.
- line 198: term.setCursor is now properly working with wide characters.
- line 632: This part generates the term functions that redirect to the current Window.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/term.lua
@@ -6,105 +6,257 @@ local text = require("text")
 local unicode = require("unicode")
 
 local term = {}
-local cursorX, cursorY = 1, 1
-local cursorBlink = nil
+local blinking = setmetatable({}, {__mode = "v"}) --allows collection of windows with enabled cursor blink
+local methods = {}
+local metatable = {
+  __index = methods,
+  __gc = function(self)
+    self:setCursorBlink(false)
+  end,
+  __pairs = function(self)
+    return function(_, k)
+      return next(methods, k)
+    end, self, nil
+  end,
+}
 
-local function toggleBlink()
-  if term.isAvailable() then
+local function gpu(self)
+  local state = self.__state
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks! Will give it a bit of testing probably tomorrow-ish.

mpmxyz:
I'm implementing a crude split screen dual shell to make implementing the focus feature possible.
While doing that I noticed a few bugs with my current code. (like "wwidth" instead of "width")
Please wait with merging until I fixed that and implemented the focus!
Edit: I'm now testing a first version of the focus feature.

fnuecke:
Allrighty, let me know when it's good to go and I'll have a look :)

mpmxyz:
RL got in the way for some days. I'm now rewriting the keyboard library a bit to remember pressed buttons per keyboard. That's needed to allow independent usage of multiple screens in a future "multithreading" upgrade. Since I'm not really experienced in using git I'll create a new pull request when this is done. (with a single commit for several files)
Apart from that it's working quite well.
https://github.com/mpmxyz/OpenComputers

PR #1539, opened 2015-11-28
Added term windows and multiscreen support to keyboard lib

mpmxyz:
A crude but working test program:
`wget 'https://raw.githubusercontent.com/mpmxyz/ocprograms/dev/home/bin/split.lua'`
Some programs have to be modified to support term windows properly. (e.g. to use term.getGlobalArea instead of gpu.getResolution)
I'd do that in a next step.

You can change the focus between term windows by using Ctrl [+ Shift] + Tab
Screenshots of the change:
https://raw.githubusercontent.com/mpmxyz/ocprograms/dev/pics/multiscreen1.png
https://raw.githubusercontent.com/mpmxyz/ocprograms/dev/pics/multiscreen2.png

API changes
- keyboard
  - changed
    - isAltDown, isControlDown, isKeyDown, isShiftDown
      - added optional 'address' parameter to specify the keyboard being used
      - defaults to the address of the primary keyboard
- term
  - added
    - term.newWindow([x:number, y:number, width:number, height:number, gpu:table]) -> window:table
      - creates a new term window object
      - x, y, width and height define the position and size of the window
        - negative values indicate a position relative to the lower right part of the screen
          - x = -10 -> Window begins 10 characters from the right border of the screen.
          - height = -2 -> Window extends to the second last line of the screen.
        - default: 1, 1, -1, -1
      - gpu == nil -> always use primary gpu
    - term.setWindow(window:table) -> oldWindow:table
      - sets the window used by the current process
      - e.g. term.read is now redirected to this object
    - term.getWindow() -> window:table
    - The following functions are available in both the library and in the window object. ("win:func()" vs. "term.func()")
      - setArea(x:number, y:number, width:number, height:number) -> oldX:number, oldY:number, oldWidth:number, oldHeight:number
      - getArea() -> x:number, y:number, width:number, height:number
      - setGPU([gpu:table]) -> oldGPU:table
      - getGPU() -> gpuBeingSet:table, gpuBeingUsed:table
        - gpuBeingUsed = gpuBeingSet or component.gpu
      - getGlobalArea() -> x:number, y:number, width:number, height:number
      - toGlobal(x:number, y:number) -> x:number, y:number
        - transforms window coordinates to screen coordinates
      - toLocal(x:number, y:number) -> x:number, y:number
        - transforms screen coordinates to window coordinates
      - focus()
        - changes the focus to this window
        - Focus changes are only applied after the next "term_focus" signal which is queued after a change.
      - unfocus()
        - removes the focus on this window and changes it to the next known one
        - Focus changes are only applied after the next "term_focus" signal which is queued after a change.
      - focusNext()
        - changes the focus to the next window.
        - Focus changes are only applied after the next "term_focus" signal which is queued after a change.
      - focusPrevious()
        - changes the focus to the previous window
        - Focus changes are only applied after the next "term_focus" signal which is queued after a change.
      - hasFocus() -> isInFocus:boolean, screenAddress:string, keyboardAddress:string
        - returns true if this window is in focus
        - additionally returns the screen and window addresses that are used for event processing

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
@@ -1,13 +1,16 @@
 local event = require "event"
-local component = require "component"
+local term = require "term"
 local keyboard = require "keyboard"
 
 local args = {...}
 
 local interactive = io.output() == io.stdout
+local function gpu()
+  return select(2, term.getGPU())
+end
 local color, isPal, evt
 if interactive then
-  color, isPal = component.gpu.getForeground()
+  color, isPal = gpu().getForeground()
 end
 io.write("Press 'Ctrl-C' to exit\n")
 pcall(function()
@@ -17,13 +20,13 @@ pcall(function()
     else
       evt = table.pack(event.pull())
     end
-    if interactive then component.gpu.setForeground(0xCC2200) end
+    if interactive then gpu().setForeground(0xCC2200) end
     io.write("[" .. os.date("%T") .. "] ")
-    if interactive then component.gpu.setForeground(0x44CC00) end
+    if interactive then gpu().setForeground(0x44CC00) end
     io.write(tostring(evt[1]) .. string.rep(" ", math.max(10 - #tostring(evt[1]), 0) + 1))
-    if interactive then component.gpu.setForeground(0xB0B00F) end
+    if interactive then gpu().setForeground(0xB0B00F) end
     io.write(tostring(evt[2]) .. string.rep(" ", 37 - #tostring(evt[2])))
-    if interactive then component.gpu.setForeground(0xFFFFFF) end
+    if interactive then gpu().setForeground(0xFFFFFF) end
     if evt.n > 2 then
       for i = 3, evt.n do
         io.write("  " .. tostring(evt[i]))
@@ -34,6 +37,6 @@ pcall(function()
   until evt[1] == "interrupted"
 end)
 if interactive then
-  component.gpu.setForeground(color, isPal)
+  gpu().setForeground(color, isPal)
 end
 
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/edit.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/edit.lua
@@ -1,4 +1,3 @@
-local component = require("component")
 local event = require("event")
 local fs = require("filesystem")
 local keyboard = require("keyboard")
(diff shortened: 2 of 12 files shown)


fnuecke:
Sounds great! Will give it a spin in a bit. Thanks again.

mpmxyz:
Since I'm already working on it, I found a part in event.lua that I don't understand:

```
  local deadline = seconds and
                   (computer.uptime() + seconds) or
                   (filter and math.huge or 0)
```

Why is the default deadline 0 if you haven't got a filter?
I assume that this is not intended because calling event.pull() without arguments causes it to only return nil when a timer ticks.
Small Update: edit.lua is taking a bit more time than anticipated because I'm also implementing proper wide character support.

mpmxyz:
I updated all programs that are included in OpenOS and added wide character support to edit.lua.
After I fixed an annoying out of memory error I also squashed all commits.

fnuecke:
Just to let you know, haven't forgotten about this, just haven't come around to test it yet, sorry!

mpmxyz:
No problem. I've implemented a small workaround for #1545.

fnuecke:
All right, looks good from what I've tested (just had to fix the workaround). Thanks again!
Pinging @payonel since this may impact some things in your coreutils rewrite.

payonel:
@mpmxyz this work is pretty cool. The features you've built here don't conflict with anything I'm doing (phew!) but definitely I'll have to merge some of my code. My concern is that I would struggle to test my merge sufficiently; you know your code intent better than I would. We should work together on this. Could I merge this with one of my branches and then PR to your repo and then you retest your features?

mpmxyz:
That would be okay. I'd just have to improve my "git fu". ;-)

mpmxyz:
I've been busy for some time. But initial tests look good. I also found out that I forgot to upload a file to try my example program.

`wget 'https://raw.githubusercontent.com/mpmxyz/ocprograms/dev/home/bin/split.lua'`
`wget 'https://raw.githubusercontent.com/mpmxyz/ocprograms/dev/home/bin/newwin.lua'`
`split`
Expect to see two shells per screen after that. You can change the focus by using Ctrl [+ Shift] + Tab or by using your mouse.
Remember: split.lua is just a hack to demonstrate what is possible with "multitasking"! I'm still waiting for a "real" implementation. (It's going to be a big rewrite to the event/signal system.)

payonel:
@mpmxyz I'd like to alter term a bit - dialog at length in this github comment section might get clunky, and your github email isn't real. Can you join #oc so we can chat?

mpmxyz:
I'm on #oc now.

PR #1556, opened 2015-12-12
Added support for HTTP request headers

invliD:


magik6k:
I guess that some would like to have an option to disable this in config

fnuecke:
Looks good! But indeed, please add a config to allow disabling it (I'm fine with it being enabled by default), then I'll merge it!

invliD:
@fnuecke It's done.

fnuecke:
Allright, thanks!

PR #1562, opened 2015-12-15
[1.6] Fixed network card message target for servers

magik6k:
Passing messages to main node of server by all installed network cards is confusing, can cause duplicate packets as well as multiple hard to fix issues with routing stacks, so I guess messages should be only sent to card's node, as connecting card node to the same(or other ofc) rail main node is connected to is [way] more intuitive than getting duplicate packets.


fnuecke:
Ah, that's also a possibility of implementing it. Tad special casey but everything else would probably be overengineering. Thanks!

Issue #1563, opened 2015-12-16
InputStreamFileSystem (and possibly others) do not set filename when throwing FileNotFoundException


Pwootage:
I noticed this while implementing the filesystem for my OS (made it very tricky to track down which file it wasn't finding).

https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/scala/li/cil/oc/server/fs/InputStreamFileSystem.scala#L39-L41

``` scala
  override def open(path: String, mode: Mode) = this.synchronized(if (mode == Mode.Read && exists(path) && !isDirectory(path)) {
    val handle = Iterator.continually((Math.random() * Int.MaxValue).toInt + 1).filterNot(handles.contains).next()
    openInputChannel(path) match {
      case Some(channel) =>
        handles += handle -> new Handle(this, handle, path, channel)
        handle
      case _ => throw new FileNotFoundException()
    }
  } else throw new FileNotFoundException())
```

I can look for the rest of these and open a pull request, if you like (after finals tomorrow :P)

fnuecke:
Ah, you mean throwing something saying "too many open handles" instead? Sure, if there are other instances (can't remember just how re-used the code in the FS stuff is), gladly!

See if just adding a message to the exception does the trick so stuff depending on it being a FNFE don't break, if it doesn't oh well.

Pwootage:
Fixed by #1570

PR #1570, opened 2015-12-16
Add filename to FileNotFoundExceptions thrown in various filesystem implementations

Pwootage:
For #1563

### DIFF
--- a/src/main/scala/li/cil/oc/server/fs/FileInputStreamFileSystem.scala
+++ b/src/main/scala/li/cil/oc/server/fs/FileInputStreamFileSystem.scala
@@ -38,7 +38,7 @@ trait FileInputStreamFileSystem extends InputStreamFileSystem {
     case file if file.exists() && file.isFile => Array(file.getName)
     case directory if directory.exists() && directory.isDirectory && directory.list() != null =>
       directory.listFiles().map(file => if (file.isDirectory) file.getName + "/" else file.getName)
-    case _ => throw new io.FileNotFoundException("no such file or directory")
+    case _ => throw new io.FileNotFoundException("No such file or directory: " + path)
   }
 
   // ----------------------------------------------------------------------- //
--- a/src/main/scala/li/cil/oc/server/component/FileSystem.scala
+++ b/src/main/scala/li/cil/oc/server/component/FileSystem.scala
@@ -301,7 +301,7 @@ class FileSystem(val fileSystem: IFileSystem, var label: Label, val host: Option
 
   private def clean(path: String) = {
     val result = com.google.common.io.Files.simplifyPath(path)
-    if (result.startsWith("../")) throw new FileNotFoundException()
+    if (result.startsWith("../")) throw new FileNotFoundException(path)
     if (result == "/" || result == ".") ""
     else result
   }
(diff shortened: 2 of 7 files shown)


src/main/scala/li/cil/oc/server/fs/FileInputStreamFileSystem.scala
```diff
@@ -38,7 +38,7 @@ trait FileInputStreamFileSystem extends InputStreamFileSystem {
     case file if file.exists() && file.isFile => Array(file.getName)
     case directory if directory.exists() && directory.isDirectory && directory.list() != null =>
       directory.listFiles().map(file => if (file.isDirectory) file.getName + "/" else file.getName)
-    case _ => throw new io.FileNotFoundException("no such file or directory")
+    case _ => throw new io.FileNotFoundException(path + " (no such file or directory)")
```
dgelessus: How about changing the format to `"no such file or directory: " + path`? That way you can see the failure reason right away even if the path is very long, which I find more useful than putting the error message after the path (where it's easy to miss too). That's how Python does it for example:

``` python
>>> open("/usr/bin/nope")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/nope'
```

Same goes for `"read-only filesystem"` down below.

(Or are these exceptions that stay inside Scala and this is an established convention there? In that case feel free to ignore this.)

fnuecke: I second that, makes sense and for more readable errors. Other than that looking good!

Pwootage:
Updated pull request with new messages as requested

fnuecke:
Thanks!

src/main/scala/li/cil/oc/server/fs/FileInputStreamFileSystem.scala
```diff
@@ -38,7 +38,7 @@ trait FileInputStreamFileSystem extends InputStreamFileSystem {
     case file if file.exists() && file.isFile => Array(file.getName)
     case directory if directory.exists() && directory.isDirectory && directory.list() != null =>
       directory.listFiles().map(file => if (file.isDirectory) file.getName + "/" else file.getName)
-    case _ => throw new io.FileNotFoundException("no such file or directory")
+    case _ => throw new io.FileNotFoundException("No such file or directory: " + path)
```
makkarpov, line 41: In Lua (as this exception will propagate to it) error messages usually start with lower case letter.

fnuecke, line 41: Yeah, I'll change it back to lowercase later :P

Pwootage, line 41: Poor non-lua developers seeing non-standard lower-case starting error
messages ;)

On Sun, Dec 20, 2015 at 3:37 PM, fnuecke notifications@github.com wrote:

> In src/main/scala/li/cil/oc/server/fs/FileInputStreamFileSystem.scala
> https://github.com/MightyPirates/OpenComputers/pull/1570#discussion_r48109340
> :
> 
> > @@ -38,7 +38,7 @@ trait FileInputStreamFileSystem extends InputStreamFileSystem {
> >      case file if file.exists() && file.isFile => Array(file.getName)
> >      case directory if directory.exists() && directory.isDirectory && directory.list() != null =>
> >        directory.listFiles().map(file => if (file.isDirectory) file.getName + "/" else file.getName)
> > -    case _ => throw new io.FileNotFoundException("no such file or directory")
> > -    case _ => throw new io.FileNotFoundException("No such file or directory: " + path)
> 
> Yeah, I'll change it back to lowercase later :P
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/MightyPirates/OpenComputers/pull/1570/files#r48109340
> .

## 

Sé onr sverdar sitja hvass!
-Pwootage

PR #1572, opened 2015-12-17
Network activity LED for servers

magik6k:
Here's how it it looks(except for blue-ish LED color): https://assets.magik6k.net/octhing.webm


fnuecke:
Looks great, thanks a lot!

PR #1573, opened 2015-12-18
Fixed embedSlot in GuiType

### LINKED ISSUE #1571: Can only access first server's contents via right clicking front face
Using build 815 from the jenkins, only a server in the first slot of a rack has its contents accessible via right clicking the front face of the rack, all other slots require you to remove the blade and access it via right clicking, as was the case prior to 1.6

magik6k:
Fixes #1571


fnuecke:
Goddammit, how did that even happen >_>
Thanks!

PR #1591, opened 2015-12-31
Ninjafix unbootable plan9k floppy

magik6k:


magik6k:
Or wait, it doesn't work too

Issue #829, opened 2015-01-17
Trading Upgrade


AtomSponge:
An upgrade for Robots and/or Drones, which comes with two tiers:
1. Allows Robots/Drones to view all offers of a Villager
2. An extended version of the first tier, which additionally allows trading with the Villager (providing that the Robot/Drone has an Inventory Upgrade installed)

_Basic idea by @ZeekDaGeek, I just thought of the different tiers_

Sbotkin:
I think your version first tier is pretty pointless. How about making it view and trade, but only second tier?

fnuecke:
I admit I also don't see a use-case for the T1 version, but the general idea is interesting.
Making it T2 sounds about right to me.

ZeekDaGeek:
Possibly for balance. If you would prefer people on your server to find a good villager, leech it and bring it to a user for trading, which would require a users involvement and interaction then disable T2 upgrade.

Also it's a pretty powerful upgrade for just 1 point on  drone.

Otherwise only ever craft the T2 upgrade and pretend the T1 doesn't exist.

fnuecke:
I think it'd be fine to have this as T2, considering it kind of balances itself a bit by the fact that to actually _find_ a villager on its own it'd need a second upgrade (like Computronics' radar).

Inari-Whitebear:
Hm, wondering how this would work.
Would it just be
e.g. trading component
- trading.listOffers() -- returns table, listing trading offers (with info on items, if the villager wants to do that trade, etc)
- trading.trade(tradeID) -- tradeID taken from the list of offers, returns false if the trade failed (e.g. the drone/robot didnt have the right items, invalid ID, villager doesn't want to trade that)

and operating on the closest villager in a small range?

fnuecke:
My idea would have been something along these lines:
- `trading.getTrades()` returns a table of userdata objects, each of which represents a trade offered by a villager in range (i.e. the list contains _all_ trades offered by _all_ villagers in range, the upgrade doesn't really need to differentiate between individual villagers IMHO).
- `trade.getInput()` returns the stack representation of what the villager wants.
- `trade.getOutput()` returns the stack representation of what the villager offers.
- `trade.trade()` tries to perform the trade, using the stack in the agent's selected inventory slot.

The methods on the userdata objects would fail once the respective villager is out of range or the trade becomes invalid.

Inari-Whitebear:
Hm, yeah, makes sense ^^ Will take a shot at it~

[...]

PR #1595, opened 2016-01-03
Trading Upgrade

Inari-Whitebear:
For #829 

The trading upgrade is for drones and robots, it has one function: getTrades(), which just returns the trades in range as Userdata.

Then calling getInput/getOutput/trade/isEnabled on those Userdata table/objects does the expected things, as described more detailed by the docstrings.

From my testing all seemed to work, unless i missed something... and if so, hopefully nothing stupidly obvious in hindsight. ^^"

Code can be cleaned up/changed/fixed of course <.< The trade function isn't very clean, but the best I got while trying to work with InventoryUtils and maybe I don't even need the extra-class.
Also I'm by no means a Scala expert :D

Icon is mostly provided by @NegiAD, i just photoshopped the OC Upgrade arrow and the MC emerald onto the thing.


src/main/scala/li/cil/oc/integration/opencomputers/DriverUpgradeTrading.scala
```diff
@@ -0,0 +1,37 @@
...
+import li.cil.oc.server.component
+
+object DriverUpgradeTrading extends Item with HostAware {
+  override def worksWith(stack: ItemStack) = isOneOf(stack,
+    api.Items.get(Constants.ItemName.TradingUpgrade))
+
+  override def createEnvironment(stack: ItemStack, host: EnvironmentHost) =
+    if (host.world.isRemote) null
+    else host match {
+      case host: EnvironmentHost with Robot => new UpgradeTrading(host)
```
fnuecke, line 22: I'd say the `with Robot` and `with Drone` here should go, since that will be handled via the blacklist (otherwise addon robots, should they be created, could not use this driver otherwise). I'll make the adjustment so you don't have to rebase again, if that's allright.

Inari-Whitebear, line 22: Sure, I think when I wrote that part I was still under the impression that createEnvironment or worksWith decide the Blacklist, somehow didn't occur to me to change it <.<

src/main/scala/li/cil/oc/server/component/Trade.scala
```diff
@@ -0,0 +1,256 @@
...
+    //Tell info to save to nbt
+    info.save(nbt)
+  }
+
+  def inventory = info.inventory
+
+  @Callback(doc="function():table, table -- returns the items the villager wants for this trade")
+  def getInput(context: Context, arguments: Arguments): Array[AnyRef] = {
+    Array(info.recipe.getItemToBuy.copy(),
+      info.recipe.hasSecondItemToBuy match {
```
fnuecke, line 46: Ah yes, Scala, where pattern matching is indeed used non-sparingly ;-)

For future reference, while there's no ternary operator in Scala, you can just do:
`if (info.recipe.hasSecondItemToBuy) info.recipe.getSecondItemToBuy.copy() else null`

Which I generally prefer where possible.

src/main/scala/li/cil/oc/server/component/Trade.scala
```diff
@@ -0,0 +1,256 @@
...
+  @Callback(doc = "function():table -- returns the item the villager offers for this trade")
+  def getOutput(context: Context, arguments: Arguments): Array[AnyRef] = {
+    Array(info.recipe.getItemToSell.copy())
+  }
+
+  @Callback(doc="function():boolean -- returns whether the villager currently wants to trade this")
+  def isEnabled(context: Context, arguments: Arguments): Array[AnyRef] = {
+    Array(info.villager.exists((villager: EntityVillager) => // Make sure villager is neither dead/gone nor the recipe
+      !info.recipe.isRecipeDisabled                          // has been disabled
+    ).asInstanceOf[AnyRef])
```
fnuecke, line 61: Also for future reference, [ResultWrapper.result](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/scala/li/cil/oc/util/ResultWrapper.scala) instead of Array, to save yourself the hassle of manually boxing all the time.

src/main/scala/li/cil/oc/server/component/Trade.scala
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  val maxRange = Settings.get.tradingRange
+  def inRange = info.villager.isDefined && distance.exists((distance: Double) => distance < maxRange)
+  def distance = info.villager match {
+    case Some(villager: EntityVillager) =>
+      info.host match {
+        case Some(h: EnvironmentHost) => Some(Math.sqrt(Math.pow(villager.posX - h.xPosition, 2) + Math.pow(villager.posY - h.yPosition, 2) + Math.pow(villager.posZ - h.zPosition, 2)))
```
fnuecke, line 69: Minor nitpick (which I tend to forget myself), while in this case this is perfectly safe, I _try_ to always use `Option(x)` for creating optionals, because that's always guaranteed to be safe, even if `x` is `null` - for `Some` that'd create an optional with a `null` inside, which you generally don't want.

src/main/scala/li/cil/oc/server/component/Trade.scala
```diff
@@ -0,0 +1,256 @@
...
+        //Takes the stack in the slot, extracts up to limit and calls the function in the first argument
+        //We don't actually consume anything, we just count that we have extracted as much as we need
+        InventoryUtils.extractFromInventorySlot((stack: ItemStack) => extracting -= stack.stackSize, inventory, ForgeDirection.UNKNOWN, slot, extracting)
+    }
+    //If we had enough, the left-over amount will be 0
+    if (extracting != 0)
+      return result(false, "not enough items to trade")
+
+    //Do the same with the second item if there is one
+    if (secondItem.isDefined) {
```
Inari-Whitebear, line 117: right, I just realized this would probably be neater as a pattern-match <.<

src/main/scala/li/cil/oc/server/component/UpgradeTrading.scala
```diff
@@ -0,0 +1,47 @@
...
+  def getTrades(context: Context, args: Arguments): Array[AnyRef] = {
+
+    val boundsLow = position.bounds.offset(-maxRange, -maxRange, -maxRange)
+    val boundsHigh = position.bounds.offset(maxRange, maxRange, maxRange)
+    val bounds = boundsLow.func_111270_a(boundsHigh)
+
+    val trades = ArrayBuffer[Trade]()
+    entitiesInBounds[EntityVillager](bounds).foreach((vil: EntityVillager) => {
+      if (inRange(vil)) {
+        val merchantRecipes = vil.getRecipes(null)
```
fnuecke, line 38: Is there a particular reason you're not using `WorldAware.fakePlayer` here, or did you just not know about it?

Inari-Whitebear, line 38: Hm I had tried something with Agent, but that didn't seem like something I should use, don't think it was WorldAware.fakePlayer though. Looked around some mod-code and something from railcraft I think used null there, and it seems to work fine...

PR #1604, opened 2016-01-07
PetRenderer.scala modernisation

puzzleddev:
I replaced the old way to identify players, using their name with the new way to identify players, using their UUID.
This means that even when the player changes their username.

Also answered the age old (about half a year) question about how to test if the player is in an inventory.


fnuecke:
Looks good I guess - this works in SSP and SMP?

Also I assume the second Izaya is Wobbo? Plus a single small PR like that doesn't quite justify an entry in that list, I'm afraid ;-)

Will test and merge as soon as I can. Squashing would be appreciated.

puzzleddev:
Yes I tested on both and sorry, just left some debug code in it. Will change as soon as possible.

puzzleddev:
First off sorry I didn't reply mainly because the Telekom decided I didn't deserve internet. I feel I need to create a new pull request because something seems to be quite corrupted about it and my local repo.

If there is no problem with it I will create a new pull request using a good branch that actually works and is squashed.

(Really sorry for not replying)

fnuecke:
No worries ;-)

Somethings gotten weird with the PR, yeah... try cleaning it up / squash it, or make a new one.

Also, something I just realized, your new `isForInventory` will probably cause other people's special thingers also to not be rendered while in a GUI (the world's still visible in the background after all). Plus there might be some GUIs which are merely used to capture all input while making players stop moving e.g., stuff like that. But! In principle I like it; fix it by passing along the player in question and check if it's the local player, too, I'd say.

PR #1610, opened 2016-01-13
openos overhaul, bash like shell and more bin utils

payonel:
# OpenOS overhaul

This contains improvements to shell, added bin utils linux shells generally have, and bug fixes found all over the place. This PR also contains memory optimizations for booting openos; OpenOS 1.6 should boot to shell prompt needing about 4k less RAM than OpenOS 1.5. I anticipate future improvements to rc and term will reduce that further.

This work has been significantly unit tested. Over 860 unit tests currently are run on all changes made during development ensuring robustness of new feature and legacy support.
### /bin/sh utilities now in /lib/sh

In upgrading /bin/sh to provide a more bash-like experience for the users, I found that it made a lot of sense to thoroughly test every feature addition and every feature change. Thus most of /bin/sh has been moved to a library in /lib/sh. This library addition has benefits outside of just the interactive shell by providing other libraries the ability to execute commands in very non-standard ways (e.g. popen utilizes /lib/sh). the OpenOS sh shell in 1.6 will provide numerous conveniences including
1. more advanced tab completion
2. boolean operators &&, ||, and ! (a note on this. scripts that return nothing are returning nil. 0, nil, and true represent a successful return. All else is false/failure)
3. multiple commands per input via ;
4. improved parsing, no more whitespace needed, 
5. alias expansion and globbing can occur in multiple statements on a single input
#### In addition to these feature additions to the new shell, various bugs have been fixed including
1. various tab completion issues
2. globbing (e.g. foo.*) general fixes, including magic character in filenames
3. correct quoting rules (as per bash behavior)
4. buffered and piped streams between commands have been improved to new levels of robustness
   I've taken great care to support legacy calls and even support custom user shells. Note that /lib/shell.lua is nearly untouched. One thing of note is that shell environment variables (e.g. set vars and aliases) are copied (like real linux shells). Now processes do not need to change their parent process environments.
### standard io

io.stdin, stdout, and stderr are now represented as file descriptor streams. In the future we may provide an api for scripts or shell to create temporary or custom io file descriptors. print, term.write, io.write, io.stdout:write, and io.output():write are all on the same stream. Further more, piped shell scripts are benefited by this because a user doesn't need to worry which "write" method they use. For users that need to draw text and specifically do not want to write to stdout can use term.drawText. Also note that stderr is now its own stream, piped commands writing to stderr will not pipe their output. Future improvements to support redirecting numbered io is planned.
Also note that due to this change, io.stdout is ALWAYS equal to io.output(). If a user want to check if stdout is on the terminal and not piped, they can check if io.stdout.tty is true
### New bin utils (and greatly improved existing ones)
- time
  - Times the runtime of a command
- sleep
  - Sleep for a defined period of time
- find
  - find a file
- grep
  - grep text from a file, recursive list of files, or stdin
- head
  - Show only part of a file (can take negative numbers for end of file as well)
- rm
  - More bash like, requiring and supporting -r and -f options in cases.
- rmdir
  - For safely removing a directory only if empty
- mktmp
  - Creates a filename with a random guid-like name under /tmp (or whatever $TMPDIR is set to)
- touch
  - Creates and empty or updates the timestamp of a file
- alias
  - Improved, more bash-like
- source
  - Executes a file as a list of commands. Boot now uses /etc/profile to set alias values, and checks for $HOME/.shrc for more sourcing
### popen

io.popen added! popen can take any valid shell command line and executes it as a subroutine allowing the user to read or write to the operation. This is a very short explanation for a feature that honestly took a considerable amount of time to get working just right.


skyem123:
Why do you call it `.bashrc`?

payonel:
habbit, I can rename it. .. ~/.shrc ? we don't really have a name for our openos shell, besides that it is just /bin/sh.lua

skyem123:
yeah, `.shrc` sounds good.

skyem123:
Also, does this still boot on a tier 1.5 computer? :P

payonel:
Yes, 1.5 is fine (though a bit tight, 50k free). My next PR for openos 1.6 (I've split this into two, finishing PR2 currently) will reduce the memory footprint. I chose to postpone memory optimizations until the 2nd PR

skyem123:
That sounds good.

payonel:
@fnuecke you said you wanted a better error message for calling `time` without any arguments. I checked what it SHOULD do and it just runs timing itself so to say. I've changed our time code to do the same, but if you'd rather it give an error like "missing args" or some such, I understand and I'd be happy to change it. Just let me know

fnuecke:
That sounds perfectly fine!

fnuecke:
Thanks again for the great work on this!

PR #1617, opened 2016-01-15
Added `playSoundAt` to debug card.

gamax92:


src/main/scala/li/cil/oc/server/component/DebugCard.scala
```diff
@@ -388,6 +388,16 @@ object DebugCard {
       null
     }
 
+    @Callback(doc = """function(x:number, y:number, z:number, sound:string, range:number) -- Play a sound at the specified coordinates.""")
+    def playSoundAt(context: Context, args: Arguments): Array[AnyRef] = {
+      checkEnabled()
+      val (x, y, z) = (args.checkInteger(0), args.checkInteger(1), args.checkInteger(2))
+      val sound = args.checkString(3)
+      val range = args.checkInteger(4)
+      world.playSoundEffect(x, y, z, sound, range / 15 + 0.5F, 1.0F)
```
fnuecke, line 397: Is the integer division intentional or should that be `range / 15f`?

CaitlynMainer, line 397: This is pretty much the playSoundAt from OpenSecurity, from my testing converting it to a float makes no difference, I don't remember why I didn't in the end though.

fnuecke, line 397: Looking at the code and usage of the method, this seems somewhat... weird. The parameter isn't the range, it's the volume. Which is automatically adjusted for distance by OpenAL / the sound system later down the pipeline. So the range popping up in there is kinda redundant. I'm assuming it "just works" because it'll generally boil down to `0 + 0.5f`, which is a common default volume, also used by chests e.g.

Would suggest changing that parameter to be called `volume` accordingly, taking a value in [0, 1].

CaitlynMainer, line 397: It's "Range" Because if you give it a large number it plays for a large range.  I've used numbers as high as 400 to reach very far away from the source block.  This is handy for playing a sound across an entire world, or atleast a vary large chunk of it.

PR #1632, opened 2016-01-31
Fix __gc and use raw metatable when it's disabled

SoniEx2:
This fixes a small __gc abuse I found, and also uses the raw metatable in a safe way when __gc is disabled. Both were tested.

__gc exploit:

```
Lua 5.3.2  Copyright (C) 1994-2015 Lua.org, PUC-Rio
> mt = {__gc=true} t = setmetatable({}, mt) -- OC only filters __gc if it's a function. The setmetatable marks an object for finalization if there's a __gc set, but it doesn't care what the __gc is. If it's marked, it will be called later. Yes this means you could use __gc if allowGC was false.
> mt.__gc = print
> t = nil
> collectgarbage() -- not available in OC Lua, but you get the idea.
table: 0x74f230
0
> 
```

### DIFF
--- a/src/main/resources/assets/opencomputers/lua/machine.lua
+++ b/src/main/resources/assets/opencomputers/lua/machine.lua
@@ -683,7 +683,7 @@ sandbox = {
     end
     local result = getmetatable(t)
     -- check if we have a wrapped __gc using mt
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks!

PR #1640, opened 2016-02-06
Mystcraft converters look for correct NBT tag for book name

ItsTheKais:
The "agename" tag it was looking for doesn't seem to exist. "DisplayName" contains the book's name, which I assume is what it's looking for.


fnuecke:
Probably changed in some update then, thanks!

PR #1641, opened 2016-02-06
TehBot

lewisakura:
Add a new robot name/

### DIFF
--- a/src/main/resources/assets/opencomputers/robot.names
+++ b/src/main/resources/assets/opencomputers/robot.names
@@ -93,6 +93,7 @@ Tachikoma          # Ghost in the Shell
 TARA UH            # Perry Rhodan
 Terminator         # Terminator
 Twiki              # Buck Rodgers
+TehBot             # A TehMinerz Robot
 Uniblab            # The Jetsons
 Unimate            # First programmable robot.
 Vertigo            # Perry Rhodan


fnuecke:
Source? (If this is an IRC bot or some such thing, no, not going down that rabbit hole :P)

lewisakura:
It was a bot I added in a private modpack that is long gone now. And no, it's not an IRC bot. It would talk to you when you talked to it.

gjgfuj:
Probably has no place in OC then? (Maybe in your modpack or server but).

lewisakura:
I think me and my team would find it funny to see it in OC.

Kubuxu:
They would, you can compile your own version of OC and use it, but inserting it into main OC would set precedence. Then it would be: "He got his name, why I shouldn't?"

EDIT: You don't even have to compile anything, just change file in JAR.

fnuecke:
What @Kubuxu said.

The general rule of thumb is real/historical robots/AIs or ones from more or less popular fiction. In other words, not stuff only you and your circle of friends know about ;)

lewisakura:
And what do you mean about 'there's a file in the jar'? I can't find it anywhere and I've been digging through OC's files.

abedef:
See the file referenced [here](https://github.com/MightyPirates/OpenComputers/blob/bc100426bcce21731b2e0ed3e745d3ab7947df42/src/main/resources/assets/opencomputers/robot.names).

Techokami:
Wait, wouldn't OC complain and fail to work if the JAR was tampered with? I remember having to compile my own JARs to include custom stuff...

But you do know that you can _rename_ your robots with an anvil, right?

LizzyTrickster:
This file is also in the assets folder so you could use a resource pack to
add/replace it (correct me if I am wrong)

fnuecke:
@Techokami that's been disabled for a long time, caused more trouble than it was worth.
@LizzyTrickster the file isn't loaded via MC's resource system (because it's loaded on the server), so you're wrong ;-)

LizzyTrickster:
I thought server side resource packs were possible? I think Cc allowed for
adding floppies that way

PR #1649, opened 2016-02-12
fixing cp -x

payonel:
Hi Sanger!

So, I have figured out the cp fix I initially was trying to introduce and the correct cp behavior that existed before my first change
1. bring back the work you did in https://github.com/MightyPirates/OpenComputers/commit/e028924d612d73a4c127d1ddc9b4db9e65da6783 - whoops - which is to check if the directory is a mount point
2. but respect the given source path and copy it even if it is a mount point (in the source, I call this "origin", origin is the given source path (the fromPath) in the copy argument)

In other words, -x should not traverse other mount points, but if your path argument points directly to a mount point - that's fine, traverse it.

(2) is a good change, it is linux like: create directories of paths given, even if path given is a mount point [but don't traverse it] BUT through testing I found this is incompatible with install: `cp -rvx /* /mnt/###` because copying /\* INCLUDES /tmp in the copy, which is wrong as tmp is a mount point we don't want during install.

At this point I felt it was best to change the install command to support this change.

so...this means to me that install is slightly wrong, ergo a small change to install (remove the *): `cp -rvx / /mnt/###`

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cp.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cp.lua
@@ -59,6 +59,11 @@ local function areEqual(path1, path2)
   return result
 end
 
+local mounts = {}
+for dev,path in fs.mounts() do
+  mounts[fs.canonical(path)] = dev
+end
+
 local function recurse(fromPath, toPath, origin)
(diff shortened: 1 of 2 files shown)


fnuecke:
Looking good, thanks again!

PR #1652, opened 2016-02-13
Add ThaumicEnergistics support

DrummerMC:


Vexatos:
Should be disabled by default, like Thaumcraft integration, I guess.

fnuecke:
Thanks!

PR #1664, opened 2016-02-20
Modernized PetRenderer

puzzleddev:
- Made player checking UUID based
- Made in inventory detection better

(This is the second time, the first time I accidentally used a corrupted Git installation, which worked about as well as it sounds)


fnuecke:
Thanks :)

PR #1675, opened 2016-02-28
new term library, faster, light-weight, better widechars

payonel:
Thanks to the support (feedback, testing, code snippets, encouragement) of members in our #oc irc channel, I've finally been able to craft a new term library.

This new term library significantly improves wide char support, uses vertical scrolling during term.read by default, uses ~80% less memory, walks the dog, and takes out the trash.

**This version allows booting to 1 stick of tier 1 RAM with approximately 30k of free memory**

This new term library drops some API that had been recently added that was better suited for separate and new libraries such as the intent for having multiple terminals. This term remake respects custom "window" size and was designed fully anticipating the potential for a mutli-term library addition later on. In fact, [here ](https://gist.github.com/payonel/187a1281e4191339857e)I demonstrate a prototype of a multi-term, concurrent process sample which depends on this new term library. Note that the sample prototype code does not make any changes to the term api, nor the test programs the two windows run (i.e. they each run a separate shell).

I have reviewed every usage of term in existing scripts found in OpenPrograms - and this new term library is compatible with all of them. I'll gladly resolve any bugs or edge cases not sufficiently tested and found by users.

All the API as described on ocdoc is supported. Some recent and undocumented API has been dropped such as hasFocus().

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/alias.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/alias.lua
@@ -1,7 +1,7 @@
 local shell = require("shell")
 local args, options = shell.parse(...)
 
-local ec = 0
+local ec, error_prefix = 0, "alias:"
 
 if options.help then
   print(string.format("Usage: alias: [name[=value] ... ]", cmd_name))
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/dmesg.lua
@@ -1,16 +1,12 @@
 local event = require "event"
 local term = require "term"
-local keyboard = require "keyboard"
 
 local args = {...}
-
+local gpu = term.gpu()
 local interactive = io.output().tty
-local function gpu()
-  return select(2, term.getGPU())
-end
 local color, isPal, evt
 if interactive then
-  color, isPal = gpu().getForeground()
+  color, isPal = gpu.getForeground()
 end
 io.write("Press 'Ctrl-C' to exit\n")
 pcall(function()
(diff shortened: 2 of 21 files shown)


payonel:
It is worth documenting the other changes included in this PR. They are bug fixes and code cleanup
/bin/alias.lua
- a bug was found that error messages were missing some text

-- the following files were edited for common reasons for the following reasons. These files may be listed again herein for addition edit reasons.
/bin/dmesg.lua
/bin/edit.lua
/bin/ls.lua
/bin/more.lua
/bin/resolution.lua
/bin/sh.lua
- term.getGPU() has been removed in favor of term.gpu() which returns only 1 value, the gpu bound to the terminal
- term.hasFocus() has been removed. This code was multi-term specific. We are developing a better solution for multiple terminals.
- term.keyboard() and term.screen() added which return the keyboard and screen proxies bound to the terminal
- term.setCursorBlink() has been optimized and simplified. There is no longer a need to call it twice now

/bin/lua.lua
/bin/sh.lua
/boot/03_io.lua
- removed terminal history clean up -- the terminal library is responsible for cleaning up history tables

/bin/ls.lua
- had a error message typo due to code refactoring
- incorrectly printed an extra blank new line for empty dirs

/bin/lua.lua
- optimized loading by delaying load of serialization library
- now instructs user to use ^D to close, instead of ^C. NOTE: The correct behavior for ^C should NOT close /bin/lua.lua, but it was felt that users may expect this legacy support due to the fact that the startup message for the lua shell was in fact to use ^C. Thus, both ^C and ^D still close the lua shell, but the instruction is to use ^D. ^C should be considered a deprecated method for exiting lua shell.

/bin/more.lua
- terminal cursor blinking is no longer a registered timer event, but occurs only while pulling events via term.pull() - a new API.

/bin/sh.lua
- an additional term_available check was added before entering the runtime loop of the shell, to ensure the terminal is ready before executing /etc/profile -- which may have significant loading output the user would want see displayed.

/bin/source.lua
- preferred xpcall to pcall to improve debug information that may be hard for the user to troubleshoot given the extra layer of invocation that source will preform

/boot/02_os.lua
/etc/profile
- (as many as were possible) environment variable settings were moved to /etc/profile and removed from /boot/02_os.lua

/boot/91_gpu.lua
- added new gpu_bound signal, used when the gpu has a screen (needed for proper terminal ready)

/boot/93_term.lua
- simplified registration for gpu_bound event. This is single terminal aware - multi-term work can be done via additional libraries.

/etc/motd
- favor term.getViewport() over gpu.getViewport(). gpu viewport is a global viewport but would be incompatible with multiple terminals. In fact, user scripts should not use gpu.getViewport() if they plan to use term.setCursor(), term.write(), etc. gpu.getViewport() gives values only best used with gpu.set, gpu.fill, etc.

/init.lua
- xpcall added for improved crash messages

/lib/sh.lua
- this change/diff looks huge, but really it is just reorganizing the code, moving all delay loaded methods to the end of the file, which simplifies reading stack traces.
- also removed dead code
- there are bug fixes here though. fixed shell variable expansion to respect quotes
- added a bit more code for delay load

/lib/term.lua
- the main purpose of this PR! this is a complete rework (in fact, the 3rd or maybe 4th iteration of reworks). Documentation in ocdoc is planned :)

/lib/text.lua
- added some methods for delay load
- removed unused code

/lib/transforms.lua
- added support for tx.select() to specify keys instead of callback only

fnuecke:
Thanks a lot, this looks great. I feel somewhat sorry for this killing the old multiterm implementation, but from what I can judge this is a superior design, and running the linked test it looks like having multiple terms is now comparatively easy to implement anyway (probably not even necessary in OpenOS itself, if so possibly as something like screen/tmux?).

Issue #1663, opened 2016-02-19
Ctrl+left/right word jumps


SoniEx2:
Currently you can only move one character to the left or one character to the right. I would like to be able to move to the previous/next word using ctrl+left and ctrl+right.

payonel:
will do

payonel:
@SoniEx2 #1679

PR #1679, opened 2016-03-02
fix mouse click cursor placement, add ctrl+move to jump words

payonel:
fix mouse click for placing cursor when prompt has scrolled beyond bottom of screen and also implement ctrl+arrow movement to jump words as per #1663


payonel:
@fnuecke this PR also has a fix for tab completion during the very first term.read -- thanks to @gamax92 for hitting this.

fnuecke:
Thanks!

PR #1681, opened 2016-03-03
fix term ctrl key (regression), fix grep display

payonel:
1. fix filename display in recursive grep results, it is now correctly showing the relative path, no longer the absolute path
2. fix highlighting multiple grep results in a single line
3. fix showing grep results for --match-only where there are multiple matches per line
4. fix crash when using control+key in term.read() This was a regression caused by a recent update to support ctrl+movement for word jumping

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/grep.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/grep.lua
@@ -152,13 +152,13 @@ if ignore_case then
 end
 
 local function getAllFiles(dir, file_list)
-  local spath = shell.resolve(dir)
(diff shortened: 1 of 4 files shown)


payonel:
added commit to fix two more term issues
1. fix crash for tty writes that occur during boot before term isAvailable
2. fix crash in term when resolution is set smaller

payonel:
@fnuecke i just realized this resize fix will only work to shrink the terminal window, not enlarge it. I'll amend the 2nd commit to this PR later.

payonel:
@fnuecke this PR is now in a great state. term crash and resize issues resolved, grep fixed too

fnuecke:
Thanks again!

PR #1705, opened 2016-03-24
Update zh_CN.lang

3TUSK:
Questions: there seems to be no difference between two `en_US.lang` in `master-MC1.7.10` and `master-MC1.8.9` branch. Does it mean that this language update can also apply to 1.8.9 version? If so, should I open another pull request?


fnuecke:
Thanks for the updated translation!

Yeah, since the two branches are pretty much feature-identical the translations work the same (there's one minor change, the EEPROM entry name being uppercase in 1.8.9, but I'll adjust that). No need for another PR, I merge the 1.7.10 branch into 1.8.9 regularly anyway :-)

3TUSK:
Thanks for explanation :)

PR #1709, opened 2016-03-26
Updates for 1.6

cyber01:
Updates for 1.6


fnuecke:
Thanks! There were some conflicts, if you could verify things look correct ingame, that'd be great (via dev build from ci.cil.li).

PR #1711, opened 2016-03-27
make coroutine wrap create tracked coroutines, needed for process data

payonel:
Due to our custom handling of process data and coroutines it is necessary to intercept all coroutine creation. Apparently, coroutine.wrap DOES NOT CALL coroutine.create (though it may in pure [C], it doesn't call back to coroutine.create in the user level). This means we need to also wrap wrap ourselves.

This problem is not a regression with OpenOS 1.6, but has existed for ever (since process data was a thing). ServerFS is an addon to OpenOS which creates a virtual fs over the network, and uses coroutine.wrap -- and fails rather quickly in OpenOS 1.6 due to this.


fnuecke:
Thanks!

PR #1712, opened 2016-03-27
give /bin/time much better output using uptime, big thanks to gamax92

payonel:
It gives a much more interesting representation of process runtime using uptime for real time and clock for cpu time. Thanks gamax for considering this for us.


fnuecke:
And thanks again ;-)

PR #1729, opened 2016-04-01
Typo in WirelessNetworkCard.setStrength?

new-sashok724:


fnuecke:
Herp derp, thanks for spotting that!

new-sashok724:
No problem. I noticed that when i called modem.setStrength(math.huge) and ran out of power :(

Issue #1671, opened 2016-02-25
Lua Interpreter Tab Completion Errors


J0SLAN:
Tab-completion for the stock Lua interpreter on OpenOS errors whenever it finds a number instead of a string, and also works for strings that have non-alphanumeric characters, for which it shouldn't. Just minor bugs.

fnuecke:
Please provide the strings this is reproducible with (i.e. what's in the input when you hit tab), thanks.

payonel:
@J0SLAN 
yes please do. OpenOS 1.6 tab completion has a lot of improvements, I'd want to test your use case against it.

J0SLAN:
@fnuecke @payonel 
Specifically, I'm pulling NBT from blocks with the Debug Card. The tab completion errors whenever I hit a table that's enumerated (like from an array of arrays), as it finds a number where it wanted a string.

Some labels for the tags have non-alphanumeric characters, like a colon, which works just fine until you try to access the value and it errors because syntax doesn't allow non-alphanumeric chars for dot indexing (=blah.foo.bar:baz doesn't work). I believe that it shouldn't even allow tab-completion here for this reason.

payonel:
@J0SLAN I apologize, I'm unsure how to repro and test and fix this. Are you using OpenOS 1.6? If not, please upgrade to latest and retest, I won't be fixing deprecated code. If you do have 1.6, what method is failing? sh.getMatchingFIles? Are you saying filesystem.list() is returning ints and not strings?

Thanks :) I'm quite invested in making this work.

gjgfuj:
@payonel it looks like they're looking at "lua.lua" specifically.

payonel:
@gjgfuj ah yes...also, in the title :) makes more sense now.

payonel:
@J0SLAN would https://git.io/v21k9 work for you?

fnuecke:
Given the massive rework/rewrite of OpenOS/term I'm almost certain this isn't an issue anymore. If it does still happen with OC 1.6/OpenOS 1.6, please let us know.

Forecaster:
Trying to tab-complete sides in the lua prompt causes this:
http://imgur.com/ZyMKUFv

It's apparently related to this : P

payonel:
@fnuecke I'll repro this tonight to confirm, but it sounds legit. If so, we should reopen this and I'll submit a fix for it.

fnuecke:
Let me know how it goes :)

[...]

PR #1737, opened 2016-04-04
term wrapping and sh tab completion fixes

payonel:
term wrapping fixes
Backspacing or deleting during vertical wrap would not update the current end of line correctly what the text was being pulled from the next line

horizontal wrapping deletion was not using dx,dy and instead was removing text from absolute positions on the screen

horizontal wrapping was pushing spaces beyond the windowed areas

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/lua.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/lua.lua
@@ -69,7 +69,7 @@ if #args == 0 or options.i then
   local function findKeys(t, r, prefix, name)
     if type(t) ~= "table" then return end
     for k, v in pairs(t) do
-      if string.match(k, "^"..name) then
+      if type(k) == "string" and string.match(k, "^"..name) then
         local postfix = ""
         if type(v) == "function" then postfix = "()"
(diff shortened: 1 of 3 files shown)


payonel:
closes https://github.com/MightyPirates/OpenComputers/issues/1671

payonel:
The third commit is also important. Found some regressions while writing some samples for Kodos on how to use term. The commit message says more.

payonel:
Turns out the broken tab complete is NOT a regression in this PR, but has been around the whole time with my new term library. It's a bit of a timing issue, but basically, key-up events are clearing the hint cache (dumb!)

Also, I tested lua prompt to have (for examples) "sides" in env all the way back to before my openos overhaul code, and it was still missing. I'll check further back ( 1.5 days ) but for now, this PR seems healthy with the tab complete fix.

payonel:
ok mr sangar
I tested ALL the way back, from the very beginning of /bin/lua's support for tab completion. It did not have a full environment with things such as sides. Sorry :(

fnuecke:
Ohh, thanks a lot for looking into that then. Old and unnoticed bugs are the best bugs I guess >_>
Nevermind that for now then, I suppose. Will re-test and merge in the evening!

PR #1743, opened 2016-04-05
Buffer chomp

payonel:
@fnuecke this is not high priority for 1.6 but for normal consideration. Buffer read allows new line chomping, but doesn't consider \r\n and \r endings correctly. This can make for some unattractive output in shell when too many newlines are printed. For example, cat /etc/motd and you'll see extra lines. This change should correctly detect \r or \r\n pairs, and step over them as a whole.


payonel:
not ready for pull, streams are chunked. will revisit later

PR #1744, opened 2016-04-05
Continue tab complete on single hit results into children

payonel:
This is not high priority, and not super heavily tested. Just for your consideration. If you do like this idea, allow me to give it more testing first. Also, this will merge conflict with https://github.com/MightyPirates/OpenComputers/pull/1737 - so if you do like this idea, I'll update this PR anyways.

Tab complete in the lua prompt shows . on tables, but can give the wrong impression that there are no keys. Consider that the user types =filesyste[tab] and then sees =filesystem. and presses tab again. The hint table has only one entry, "=filesystem." and is cycling through that one entry and not reconsidering children of filesystem .. which is what the prompt makes the user think. This change injects a metamethod lookup in the 2nd index of a single result to repop the hint list with new children.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/lua.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/lua.lua
@@ -98,6 +98,22 @@ if #args == 0 or options.i then
       table.insert(r2, k)
     end
     table.sort(r2)
+    if #r2 == 1 then
+      setmetatable(r2, {
+        __index=function(tbl, key)
+          if key==2 then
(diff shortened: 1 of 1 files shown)


payonel:
Turns out the broken tab complete is NOT a regression in this PR, but has been around the whole time with my new term library. It's a bit of a timing issue, but basically, key-up events are clearing the hint cache (dumb!)

Also, I tested lua prompt to have (for examples) "sides" in env all the way back to before my openos overhaul code, and it was still missing. I'll check further back ( 1.5 days ) but for now, this PR seems healthy with the tab complete fix.

PR #1767, opened 2016-04-16
cp -P added to support preserving symlinks. Also, -r enables -P

### LINKED ISSUE #1763: installer is too instally
Vex informed me that symbolic links are being copied as real files during install - a problem if you have (for example) an internet card in the system which creates a sym link for /bin/wget.lua

Turns out this is natural cp behavior (and in fact not a regression in OpenOS 1.6)
But I agree it would be undesirable behavior for an install.

POSIX cp has a -P param to preserve attributes, and copies sym links as sym links. Also, turns out that -r enabled -P by default.

Thus, I'm adding support for -P (And -r will enable it), and install (because it uses -r) won't need an update.

PR expected.

payonel:
real shell testing shows that -P is enabled by -r
also, whether the target is a dir or a file, if we have -P and it is a symlink, then we create a new symlink and we don't follow the symlink (in case of dir)
install is fixed by this because it uses -r

closes https://github.com/MightyPirates/OpenComputers/issues/1763

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cp.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/bin/cp.lua
@@ -9,10 +9,13 @@ if #args < 2 then
   io.write(" -r: copy directories recursively.\n")
   io.write(" -u: copy only when the SOURCE file differs from the destination\n")
   io.write("     file or when the destination file is missing.\n")
+  io.write(" -P: preserve attributes, e.g. symbolic links.\n")
   io.write(" -v: verbose output.\n")
   io.write(" -x: stay on original source file system.\n")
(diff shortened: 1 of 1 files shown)


Kubuxu:
What if links are absolute but link in scope of copy. Like

```
./dirA/fileA
./dirB/linkA -> $(realpath ./dirA/fileA)
```

I have no idea how real cp will behave in this case so it might be already right.

payonel:
Things to consider about OpenOS sym links
1. they are not persisted
2. they are ALL absolute paths

Secondly, real cp in this case will create an absolute path sym link

fnuecke:
Works as it should for me (tested by running install with an internet card installed). Thanks!

PR #1768, opened 2016-04-16
Chomp \r and \r\n from lines, not just \n

payonel:
This change helps buffer read newlines respecting the different standards, \r, \n, and \r\n

It checks the char following \r in case of \n and treats it as a single newline

As you pointed out in irc, I needed to consider that the stream buffers the data and the \n could be sitting one char beyond the current buffer. I looked at keeping state of the buffer reads between calls, but I feel the most efficient solution to this issue (given all the other types of reads that can occur between calls, such as read number -- in which case I'd have to modify readChunk to eat the next newline under special circumstances) -- the most efficient solution to me seems to just read another  chunk (at most one more).

Given the impact this area could have, I wrote some read line tests here:
https://github.com/OpenPrograms/payonel-Programs/blob/master/payo-tests/var/payo-tests/slow-test.lua#L227
Which pass WITH this change, and fail without (given that it is testing all line ending types).
Following the mixed tests are also same type of tests, but with \n only. These tests pass with and without this buffer change.

Let me know if there is more you'd like tested - or, if you have other thoughts about changing readLine in this way.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/lib/buffer.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/lib/buffer.lua
@@ -240,13 +240,20 @@ function buffer:read(...)
   local function readLine(chop)
     local start = 1
     while true do
-      local l = self.bufferRead:find("\n", start, true)
-      if l then
-        local result = self.bufferRead:sub(1, l + (chop and -1 or 0))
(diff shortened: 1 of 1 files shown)


fnuecke:
Looks good to me, and as it only relates to readLine I agree it's a good idea. Did some (comparatively superficial :P) testing and seems to work as it should. Thanks!

PR #1776, opened 2016-04-25
big awesome update for devfs, command substitution, and io redirection

payonel:
[Edit: for clarification, tests are in a different repo]

This change should be fully compatible with any 1.6 user scripts. This makes no change to api, only improves shell support and makes important process handle cleanup (when pipes and redirection are used - we are not auto-closing user file handles).

This update also reduces RAM costs by another 1300 bytes.

This PR focuses on 3 major feature and various bug fixes found along the way.
## devfs
- devfs mounts a virtual device at /dev. With this release I am only adding /dev/null and /dev/random. The lib/devfs provides a create method to allow user defined dev points.
## stderr
- stderr (io[2]) is now entirely its own io stream, separate from stdout (io[1]). Along with providing a true stderr separation it was necessary to support numbered stdio redirection: #>&# (e.g. >&2 or 2>&1, or even 2>/dev/null)
## command substitution
- Support for command substitution is provide via backticks `. Backtick text is resolved as a sub process and the stdout (io[1]) of the process replaces the text inline. set list=`ls` would set $list to the string of the output of ls

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/OpenOS/boot/03_io.lua
+++ b/src/main/resources/assets/opencomputers/loot/OpenOS/boot/03_io.lua
@@ -33,14 +33,19 @@ function stdoutStream:write(str)
 end
 
 function stderrStream:write(str)
-  local component = require("component")
-  if component.isAvailable("gpu") and component.gpu.getDepth() and component.gpu.getDepth() > 1 then
-    local foreground = component.gpu.setForeground(0xFF0000)
-    term.write(str, true)
-    component.gpu.setForeground(foreground)
-  else
-    term.write(str, true)
+  local gpu = term.gpu()
+  local set_depth = gpu and gpu.getDepth() and gpu.getDepth() > 1
+
+  if set_depth then
+    set_depth = gpu.setForeground(0xFF0000)
   end
+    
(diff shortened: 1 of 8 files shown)


payonel:
I regret naming this PR "big" :) because I don't want to scare people from being excited about this. This PR is quite focused on the following 3 features:
devfs: /dev/null and /dev/random (and /lib/devfs has a create method to create user defined dev points)
stderr true separation from stdout with numbered io redirect support: echo hello >&2
command substitution with backticks

magik6k:
You can look at plan9k devfs drivers from lib/modules/base/17_\* and port some here as they should be sort of portable. I may do this as well, but totally no ETA as I'm away from minecraft/OC.

fnuecke:
Thanks, looking good! I did have to add a few missing methods to devfs to avoid programs expecting them to be there breaking (such as `/bin/df`), please see if the returned values make sense (e.g. should it rather return `true` from `isReadOnly`? Since `/dev/null` e.g. is technically writable, I'm a bit conflicted on that one.)

payonel:
I looked over the added code, thank you for that. I also think isReadOnly is false given the circumstances.

Issue #1775, opened 2016-04-24
Wrong backspace symbol in OpenOS shell


IgorTimofeev:
I'm using OpenComputers-MC1.7.10-1.6.0.3-rc.1-universal.jar mod version and OS X version of Minecraft. When i try to press backspace in shell, i get following:

![2016-04-24 23 51 35](https://cloud.githubusercontent.com/assets/11760002/14770337/46424538-0a80-11e6-9a14-02b02b32bd52.png)

But it edit.lua everything's okay - is seems like shell uses wrong key_down event data. Hope this will be fixed soon, thanks.

payonel:
This looks like a mac-specific problem. I'll look into this promptly.

payonel:
@IgorTimofeev could you help me by testing a couple things?
in-game, start lua prompt. then enter this command:
lua>=event.pull("key_down")

Press enter once (the blinking cursor should go away) and then press the same key you are having problems with (that you opened this ticket for) - which you are calling the "backspace"

You should then see some information printed, and the lua prompt returns. Here is what I see when I press backspace:
"key_down"      "d8a526bb-ad4c-40dc-b721-df7790134e46"   8       14
The long GUID value I do not need, just the 2 digits after that. In my case 8 and 14

Thanks

IgorTimofeev:
@payonel There's no need for such detailed instructions, I know how the event system works, thanks :D 
But here is your event data:

![2016-04-25 9 31 19](https://cloud.githubusercontent.com/assets/11760002/14775564/36e510e2-0ad1-11e6-9763-8c245c2b68ec.png)

The problem is that shell uses **3rd** key_down event argument instead of **4th**.

![2016-04-25 9 34 25](https://cloud.githubusercontent.com/assets/11760002/14775590/7472ace4-0ad1-11e6-8710-53cf39f28048.png)

Personally, I've fixed this problem by adding the following code in the **lib/term.lua** file:

![2016-04-25 9 51 36](https://cloud.githubusercontent.com/assets/11760002/14775925/b5f6c540-0ad3-11e6-81cd-7ca342802635.png)

payonel:
The problem i have with this is that 8 is CORRECT ascii for backspace, 14 is not. 14 and 15 are SO and SI codes (old-school control key). And 127 is actually the delete key. I am not checking the wrong return, your keyboard is sending SO-DEL on backspace it seems. I think this may be a Russian keyboard layout issue.

[...]

PR #1782, opened 2016-04-27
fix onTouch (mouse click) for unicode support in shell

payonel:
as you request, mouse click cursor position placement fixed for unicode


payonel:
plus added the backspace fix, to correctly use the key code as per http://minecraft.gamepedia.com/Key_codes and this closes #1775

payonel:
I meant to Comment, not "Close and comment" -- sorry

fnuecke:
Awesome, thanks, will test this evening!

PR #1786, opened 2016-04-28
devfs should be rw

payonel:
This should be returned in the boot devfs file, not the lib devfs. We already have a method defined in boot/10_devfs, just moving the logic to return false and removing it from /lib/devfs


payonel:
This should be returned in the boot devfs file, not the lib devfs. We already have a method defined in boot/10_devfs, just moving the logic to return false and removing it from /lib/devfs

fnuecke:
Whoops, totally overlooked this one, sorry for the delay!

PR #1819, opened 2016-05-25
Don't use the broken GlStateManager #pushAttrib / #popAttrib.

Lordmau5:
Fixes rendering issues with not just FFS, but potentially even other mods :P
(Also includes me removing every pushAttrib / popAttrib from my mod)

Chatlog regarding this from #simplymodding:
http://pastebin.com/fqTFrGrN

### DIFF
--- a/src/main/scala/li/cil/oc/client/gui/Assembler.scala
+++ b/src/main/scala/li/cil/oc/client/gui/Assembler.scala
@@ -55,7 +55,7 @@ class Assembler(playerInventory: InventoryPlayer, val assembler: tileentity.Asse
   }
 
   override def drawSecondaryForegroundLayer(mouseX: Int, mouseY: Int) = {
-    GlStateManager.pushAttrib()
+    //GlStateManager.pushAttrib()
     if (!inventoryContainer.isAssembling) {
       val message =
         if (!inventoryContainer.getSlot(0).getHasStack) {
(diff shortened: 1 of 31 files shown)


fnuecke:
Sooo this part of GlStateManager is utterly broken to the point of unusability? Why am I not really surprised...

Is that also true for 1.8.9?

Lordmau5:
Glad you brought this up!
It is, indeed, also present in 1.8.9.
I didn't check 1.8.9 though in regards to my OverlayRenderer.

Image

//Ninjaedit:
Funnily enough, I don't have any attrib code in my OverlayRenderer in 1.8.9, but it still happens since it's in your renderers:P

fnuecke:
OK, thanks for checking! I'll probably replace them with a helper method in (OC's) `RenderState` or so, which for now just does nothing. Just to avoid having to track down all the places again should it start working with the `GlStateManager` because they start to push/pop the cache, too :P

fnuecke:
Allrighty, should be fixed via [726aa361 (1.8.9)](https://github.com/MightyPirates/OpenComputers/commit/726aa3616437cf2c2804f57cb013c7cd572473d8) / [5624b881 (1.9.4)](https://github.com/MightyPirates/OpenComputers/commit/5624b88151e6aa03abac10b2f5629579a9122945). Let me know if it isn't.
Thanks again for letting me know!

Lordmau5:
Can confirm, it is fixed! :)
Image

PR #1842, opened 2016-06-02
Faster serialization library with additional features

IgorTimofeev:
Hi. I recently discovered that strings concatenation in Lua does not work as fast as table.concat(), so I decided to write a new library for table serialization and add a few necessary (by my opinion) features - saving serialized tables to files and reading them from a file. There's also possibility of "pretty look" serialization, and user also can specify indentation width, and use a tab character instead of a whitespace:

The screenshot below shows a speed comparison of the old and new algorithms.

Of course, you may have your own requirements for such things, but adding them is not a problem. Here is link to the source code, if you're interested in it:
[https://github.com/IgorTimofeev/OpenComputers/blob/master/lib/serialization.lua](url)


LizzyTrickster:
you appear to have made a PR between two branches in OC's own repo and not from yours to OCs, you might want to try again

IgorTimofeev:
@LizzyTrickster Sorry, will try again.

PR #1843, opened 2016-06-02
Faster serialization library

IgorTimofeev:
Hi. I recently discovered that strings concatenation in Lua does not work as fast as table.concat(), so I decided to write a new library for table serialization and add a few necessary (by my opinion) features - saving serialized tables to files and reading them from a file. There's also possibility of "pretty look" serialization, and user also can specify indentation width, and use a tab character instead of a whitespace:

The screenshot below shows a speed comparison of the old and new algorithms:

Of course, you may have your own requirements for such things, but adding them is not a problem. Gist link: https://gist.github.com/IgorTimofeev/2162dd3db8455a75a27c20c8442844fd


payonel:
I like what you're doing here. Let me test this tonight and then give some feedback.

payonel:
also, feel free to share any other test files you have written for this, e.g. on gist. And remove the comment section "playground" out of this file (i.e. update this PR), and put it in a gist and share a link here, thanks.

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,160 @@
...
+			local file = io.open(path, "r")
+			local data = serialization.unserialize(file:read("*a"))
+			file:close()
+			return data
+		end
+	else
+		error("\"" .. path .. "\" doesn't exists")
+	end
+end
+
```
payonel: remove playground section, please. please feel free to share tests via gists and give links to those gists in the pr discussion.

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,160 @@
...
+-- print("Done.")
+-- print("Saving to file...")
+-- serialization.serializeToFile("serializationTest.lua", sampleTable, true)
+-- print("Done.")
+-- print("Loading from file...")
+-- serialization.unserializeFromFile("serializationTest.lua")
+-- print("Done.")
+
+----------------------------------------------------------------------------------------------------------------------
+
```
payonel: Remove this comment playground section, please. Feel free to share tests via gists and give links to those gists in the pr discussion.

IgorTimofeev:
@payonel No problem, gist link added, comments removed.

payonel:
leaving some comments inline with the code. If we agree on a final PR, I'd like you to squash all commits in this PR into a single commit and then force push to the branch to update this PR. But -- wait on that, let's keep the history for the discussion's sake.

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,93 @@
...
+	table.remove(text, (prettyLook and #text - 1 or #text))
+	table.insert(text, (prettyLook and oldIndentationSymbol or nil))
+	table.insert(text, "}")
+	return text
+end
+
+------------------------------------------------- Public methods -----------------------------------------------------------------
+
+function serialization.serialize(array, prettyLook, indentationWidth, indentUsingTabs)
+	checkArg(1, array, "table")
```
payonel, line 49: it is good to keep the first two args compatible with existing user scripts. That said, let's keep the name of the first argument "value" and I don't see why serialize needs to checkArg that the 1st param is a table, doSerialize doesn't care. Also, consider not using an args list for indentationWidth and indentUsingTabs, consider using a hash of args (typically called options). Why? because when order doesn't matter and all of the args are optional (i.e. can be nil) it is best a hash.

_EDIT_

I have given this more thought, not final, but better

```
function serialization.serialize(value, options)
  if type(options) == "boolean" or options == nil then
    options = { pretty = options and 10 or nil }
  elseif type(options) == "number" then
    options = { pretty = options }
  end
  -- ... 
end
```

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,93 @@
...
-    end
-    if truncate then
-      return result:sub(1, truncate) .. "..."
-    end
-  end
-  return result
+------------------------------------------------- Private methods -----------------------------------------------------------------
+
+local function doSerialize(array, text, prettyLook, indentationSymbol, oldIndentationSymbol, equalsSymbol)
+	text = {"{"}
```
payonel, line 8: we should consider supporting legacy usage of `pretty` which can be a number as well. When a number, it indicated the length of the serialized string output -- truncating and appending ... when reached.

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,93 @@
...
+	return text
+end
+
+------------------------------------------------- Public methods -----------------------------------------------------------------
+
+function serialization.serialize(array, prettyLook, indentationWidth, indentUsingTabs)
+	checkArg(1, array, "table")
+	indentationWidth = indentationWidth or 2
+	local indentationSymbol = indentUsingTabs and "	" or " "
+	indentationSymbol, indentationSymbolHalf = string.rep(indentationSymbol, indentationWidth)
```
payonel, line 52: I'm checking this code via browser, not with an editor, nor have I tested. But -- is that " "(space) or " "(space) ? Is "\t" what you meant? Also, what is indentationSymbolHalf used for here? string.rep only returns 1 value. Also, if the user sends using tabs true and width 4, is that 4 tabs? Is that what we want?

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,93 @@
...
+	return table.concat(doSerialize(array, {}, prettyLook, indentationSymbol, "", prettyLook and " = " or "="))
+end
```
payonel, line 54: doSerialize is a private method, so it's signature is of less concern to me. consider options hash as well, but, I'm less picky with private(local) methods.

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,93 @@
...
+
+function serialization.serializeToFile(path, array, prettyLook, indentationWidth, indentUsingTabs, appendToFile)
+	checkArg(1, path, "string")
+	checkArg(2, array, "table")
+	filesystem.makeDirectory(filesystem.path(path) or "")
+	local file = io.open(path, appendToFile and "a" or "w")
+	file:write(serialization.serialize(array, prettyLook, indentationWidth, indentUsingTabs))
+	file:close()
 end
```
payonel, line 69: I don't think this method is what we want, but a similar feature is not a bad idea.

Path shouldn't be the first argument, but given the optional trailing arguments, I understand the predicament. There are few things to consider with this method. Because you are building the entire string (doSerialize) and  then writing to file, you are not saving the user any memory or performance costs. It would be best to let the filesystem handle the buffering and write one element of the table at a time. Consider having doSerialize return the table of strings, and write to file one segment at a time. With that change, you can have serialize concat the table itself before returning.
Also, this method shouldn't be trying to share feature space with open mode (i.e. you're asking for the open mode). The user should just pass the filehandle.
serialization.serialize(value, options)
...

```
if options.output then
  for _,segment in ipairs(result) do
    options.output:write(segment)
  end
else
  return table.concat(result)
end
```

I need to give this more thought. I'm not sure I have the best solution yet.

src/main/resources/assets/opencomputers/loot/openos/lib/serialization.lua
```diff
@@ -1,134 +1,93 @@
...
+		else
+			local file = io.open(path, "r")
+			local data = serialization.unserialize(file:read("*a"))
+			file:close()
+			return data
+		end
+	else
+		error("\"" .. path .. "\" doesn't exists")
+	end
 end
```
payonel, line 85: like my comment on serializeToFile, we should consider overloading unserialize to take a stream to read from. I will give this more thought and consult with fellow developers smarter than myself.

payonel, line 85: I gave unserialize a bit more thought, i think unserialize should just take a string (current) or a table (any stream object that supports read)

fnuecke:
Thanks for sharing! Looks like a nice speedup.

I'm not a fan of the serialization lib requiring the filesystem though, in that regard I'd support payonel's suggestions of those taking streams instead (or get rid of them altogether).

payonel:
closing in favor of #2514

PR #1857, opened 2016-06-08
MineOS loot disk request

IgorTimofeev:
Hi. I would like to ask you, if it is possible, to add a loot disk with my GUI add-on to OpenOS. It has it's personal app store, multilanguage support, a lot of cool applications and more. Here is screenshot below:

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/mineos/installer.lua
+++ b/src/main/resources/assets/opencomputers/loot/mineos/installer.lua
@@ -1,0 +1,385 @@
+local component = require("component")
+local computer = require("computer")
+local term = require("term")
+local unicode = require("unicode")
+local event = require("event")
+local fs = require("filesystem")
+local seri = require("serialization")
+local shell = require("shell")
+local gpu = component.gpu
+
+-----------------Проверка компа на соответствие сис. требованиям--------------------------
+
+shell.execute("cd ..")
+shell.setWorkingDirectory("")
+
+--Создаем массив говна
+local govno = {}
+
+print(" ")
+print("Analyzing computer for matching system requirements...")
+
+--Проверяем, не планшет ли это
+if component.isAvailable("tablet") then table.insert(govno, "Tablet PC detected - You can't install MineOS on tablet because of primitive GPU and Screen.") end
+
+--Проверяем GPU
+if gpu.maxResolution() < 150 then table.insert(govno, "Bad GPU or Screen - MineOS requires Tier 3 GPU and Tier 3 Screen.") end
+
+--Проверяем оперативку
+if math.floor(computer.totalMemory() / 1024 ) < 1536 then table.insert(govno, "Not enough RAM - MineOS requires at least 1536 KB RAM.") end
+
+if fs.get("bin/edit.lua") == nil or fs.get("bin/edit.lua").isReadOnly() then table.insert(govno, "You can't install MineOS on floppy disk. Run \"install\" in command line and install OpenOS from floppy to HDD first. After that you're be able to install MineOS from Pastebin.") end
+
+--Если нашло какое-то несоответствие сис. требованиям, то написать, что именно не так
+if #govno > 0 then
+  print(" ")
+  for i = 1, #govno do
+    print(govno[i])
+  end
+  print(" ")
+  return
+else
+  print("Done, everything's good. Proceed to downloading.")
+  print(" ")
+end
+
+------------------------------------------------------------------------------------------
+
+local lang
+
+local applications
+
+local padColor = 0x262626
+local installerScale = 1
+
+local timing = 0.2
+
+-----------------------------СТАДИЯ ПОДГОТОВКИ-------------------------------------------
+
+local function request(url)
+  local success, response = pcall(component.internet.request, url)
+  if success then
+    local responseData = ""
+    while true do
+      local data, responseChunk = response.read() 
+      if data then
+        responseData = responseData .. data
+      else
+        if responseChunk then
+          return false, responseChunk
+        else
+          return responseData
+        end
+      end
+    end
+  else
+    return false, reason
+  end
+end
+
+--БЕЗОПАСНАЯ ЗАГРУЗОЧКА
+local function getFromGitHubSafely(url, path)
+  local success, reason = request(url)
+  if success then
+    fs.makeDirectory(fs.path(path) or "")
+    fs.remove(path)
+    local file = io.open(path, "w")
+    file:write(success)
+    file:close()
+    return success
+  else
+    io.stderr:write("Can't download \"" .. url .. "\"!\n")
+    return -1
+  end
+end
+
+local GitHubUserUrl = "https://raw.githubusercontent.com/"
+
+
+--------------------------------- Стадия стартовой загрузки всего необходимого ---------------------------------
+
+
+local preLoadApi = {
+  { paste = "IgorTimofeev/OpenComputers/master/lib/ECSAPI.lua", path = "lib/ECSAPI.lua" },
+  { paste = "IgorTimofeev/OpenComputers/master/lib/colorlib.lua", path = "lib/colorlib.lua" },
+  { paste = "IgorTimofeev/OpenComputers/master/lib/image.lua", path = "lib/image.lua" },
+  { paste = "IgorTimofeev/OpenComputers/master/lib/config.lua", path = "lib/config.lua" },
+  { paste = "IgorTimofeev/OpenComputers/master/MineOS/Icons/Languages.pic", path = "MineOS/System/OS/Icons/Languages.pic" },
+  { paste = "IgorTimofeev/OpenComputers/master/MineOS/Icons/OK.pic", path = "MineOS/System/OS/Icons/OK.pic" },
+  { paste = "IgorTimofeev/OpenComputers/master/MineOS/Icons/Downloading.pic", path = "MineOS/System/OS/Icons/Downloading.pic" },
+  { paste = "IgorTimofeev/OpenComputers/master/MineOS/Icons/OS_Logo.pic", path = "MineOS/System/OS/Icons/OS_Logo.pic" },
+}
+
+print("Downloading file list")
+applications = seri.unserialize(getFromGitHubSafely(GitHubUserUrl .. "IgorTimofeev/OpenComputers/master/Applications.txt", "MineOS/System/OS/Applications.txt"))
+print(" ")
+
+for i = 1, #preLoadApi do
+  print("Downloading \"" .. fs.name(preLoadApi[i].path) .. "\"")
+  getFromGitHubSafely(GitHubUserUrl .. preLoadApi[i].paste, preLoadApi[i].path)
+end
+
+print(" ")
+print("Initialization stage is complete, loading installer")
+print(" ")
+
+package.loaded.ecs = nil
+package.loaded.ECSAPI = nil
+_G.ecs = require("ECSAPI")
+_G.image = require("image")
+_G.config = require("config")
+
+local imageOS = image.load("MineOS/System/OS/Icons/OS_Logo.pic")
+local imageLanguages = image.load("MineOS/System/OS/Icons/Languages.pic")
+local imageDownloading = image.load("MineOS/System/OS/Icons/Downloading.pic")
+local imageOK = image.load("MineOS/System/OS/Icons/OK.pic")
+
+ecs.setScale(installerScale)
+
+local xSize, ySize = gpu.getResolution()
+local windowWidth = 80
+local windowHeight = 2 + 16 + 2 + 3 + 2
+local xWindow, yWindow = math.floor(xSize / 2 - windowWidth / 2), math.ceil(ySize / 2 - windowHeight / 2)
+local xWindowEnd, yWindowEnd = xWindow + windowWidth - 1, yWindow + windowHeight - 1
+
+
+-------------------------------------------------------------------------------------------
+
+local function clear()
+  ecs.blankWindow(xWindow, yWindow, windowWidth, windowHeight)
+end
+
+--ОБЪЕКТЫ
+local obj = {}
+local function newObj(class, name, ...)
+  obj[class] = obj[class] or {}
+  obj[class][name] = {...}
+end
+
+local function drawButton(name, isPressed)
+  local buttonColor = 0x888888
+  if isPressed then buttonColor = ecs.colors.blue end
+  local d = { ecs.drawAdaptiveButton("auto", yWindowEnd - 3, 2, 1, name, buttonColor, 0xffffff) }
+  newObj("buttons", name, d[1], d[2], d[3], d[4])
+end
+
+local function waitForClickOnButton(buttonName)
+  while true do
+    local e = { event.pull() }
+    if e[1] == "touch" then
+      if ecs.clickedAtArea(e[3], e[4], obj["buttons"][buttonName][1], obj["buttons"][buttonName][2], obj["buttons"][buttonName][3], obj["buttons"][buttonName][4]) then
+        drawButton(buttonName, true)
+        os.sleep(timing)
+        break
+      end
+    end
+  end
+end
+
+
+------------------------------ВЫБОР ЯЗЫКА------------------------------------
+
+ecs.prepareToExit()
+
+local downloadWallpapers, showHelpTips, downloadAllApps = false, false, false
+
+do
+
+  clear()
+
+  image.draw(math.ceil(xSize / 2 - 30), yWindow + 2, imageLanguages)
+
+  --кнопа
+  drawButton("Select language",false)
+
+  waitForClickOnButton("Select language")
+
+  local data = ecs.universalWindow("auto", "auto", 36, 0x262626, true,
+    {"EmptyLine"},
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/opencomputers/loot/mineos/installer.lua
```diff
@@ -0,0 +1,385 @@
...
+local shell = require("shell")
+local gpu = component.gpu
+
+-----------------Проверка компа на соответствие сис. требованиям--------------------------
+
+shell.execute("cd ..")
+shell.setWorkingDirectory("")
+
+--Создаем массив говна
+local govno = {}
```
SaphireLattice, line 17: What that table actually does? The name... doesn't really says anything.

MoonlightOwl:
I think, [OpenPrograms](https://github.com/OpenPrograms) repository is the right place for such a kind of stuff.

src/main/resources/assets/opencomputers/loot/mineos/installer.lua
```diff
@@ -0,0 +1,385 @@
...
+  getFromGitHubSafely(GitHubUserUrl .. preLoadApi[i].paste, preLoadApi[i].path)
+end
+
+print(" ")
+print("Initialization stage is complete, loading installer")
+print(" ")
+
+package.loaded.ecs = nil
+package.loaded.ECSAPI = nil
+_G.ecs = require("ECSAPI")
```
SaphireLattice, line 128: Please don't use global table to store things. Make it local.

gjgfuj, line 128: I would say this is absolutely a valid lootdisk addition imo (assuming
other complaints don't exist).

fnuecke:
I'd actually be fine with having another OS on a loot disk, as long as the "multilingual" also includes English. And if it works without having to download additional stuff (keep in mind that loot disks are not bound to the capacity limits normal floppies are).

IgorTimofeev:
@fnuecke It will be extremely difficult to make an standalone-version of MineOS by following reasons:
1) There is an AppMarket application with a dynamic update system. So people can simply run the app and download all updates through the Internet via one button.
2) I am changing the basic system files almost every day, adding new programs, adding localization files and fixing mistakes in existing ones. And, of course, all of these updates becomes immediately available in AppMarket. It is much more convenient than having a static version MineOS on a floppy disk without the possibility to update it if desired. 
3) Online installer has some options like a "Download all apps" or "Download wallpapers", it's pretty helpful if user doesn't want to install unnecessary programs or heavy-weight pictures.

So if I make static version of OS - all of this things will be useless until loot disk upgrade, and if users faces with some errors - I will not be able to help them.

src/main/resources/assets/opencomputers/loot/mineos/installer.lua
IgorTimofeev, line 17: @dangranos This is like an english word "shit". We're creating a table of shit - and if installer faces some shitty RAM modules, video cards, etc - table will fill with information about this shit. And, of course, installer will close and show this information to user.

I hope such a minor thing like name of the variables will not create additional problems, although for some people it is really more important than speed, beauty and quality of the program.

SaphireLattice, line 17: Can't you just name that variable "errors" or "problems"? And i know what that word means, thank you very much for a short explanation of a single russian word.

SaphireLattice:
@IgorTimofeev MAybe then it's time to split the OS into multiple packages? One of the packages being the core OS: desktop and libraries to run it plus some must-have programs like terminal and editor?

skyem123:
It's worth noting that there are other OSes that have online functionality, yet still have loot disks.

You can keep the online utilities, and maybe encourage the user to have an internet card, but is having an outdated version on the loot disk (updated whenever OC updates) really _that_ bad?

IgorTimofeev:
@dangranos I's already splitted to packages. Installer option "Download all apps" is used just for this purposes.

@skyem123 No, it's not really _that_ bad, I just do not understand _why_ you need this. There's nothing wrong that for OS installation user needs an internet connection. I don't understand why I must create difficulties - the Installer works fine, OS installs fine, what's wrong?

skyem123:
@IgorTimofeev, well, the thing is, the point of loot disks is to not need an internet connection to use them.

SaphireLattice:
By putting installer on the loot disk you're basically replacing the step
where it downloads it to a step where user grabs the floppy. It doesn't
makes sense to make a floppy that has installer which requires internet to
work. Because you can easily download the installer if you have inet
anyways.

fnuecke:
I was under the impression p9k did not _require_ an internet connection to install/work. If that's incorrect and there's precedent anyway, bleh. But yeah, the point of the loot disks was/is to _not_ require an internet connection (well, explicit internet programs excluded). If it's totally unrealistic for this OS... meh. Will consider it.

IgorTimofeev:
@skyem123 @fnuecke Of course, I can make a copy of my hard drive and commit files to GitHub - but what about language choose, what about wallpapers and applications installation? The basic concept of this OS - is the automatic downloading all necessary files based on user's choice. You don't need internet connection to work with this OS, just for install - only in this way you can get latest and stable OS version.

skyem123:
@fnuecke: If I recall correctly, Plan9k doesn't require an internet connection, but it has a package manager that it uses to do system updates, similar to modern *nix distributions.

skyem123:
@IgorTimofeev, well, you could have a complete installer on the loot disk, which includes a copy of all the files, including the optional ones, so a user can install it without an internet connection.

However, you can also have an option in the offline installer that downloads the latest version from the internet to install instead. This is how modern Windows and Linux install work.

As you seem concerned about out of date versions, you can put messages in the installer that an internet connection is highly recommended to get the best version of the system

fnuecke:
Well put.

That said, my main concern is not even it requiring an internet connection per-se, it's that I'd like it to work to some capacity still when the backing server is unreachable, be that because it's been abandoned, connectivity trouble or what-not.

IgorTimofeev:
@skyem123 @fnuecke But it's not an standalone OS! It's just an add-on to OpenOS, why make everything so difficult? I can agree with your point if MineOS was autonomous and independent - then the installer and other things would be reasonable. But to make a separate installer for the add-on and rewrite hundreds of files, rewrite basic concept of OS just for... for this? Smells like an masochism! :D

Maybe it's time to make it autonomous, yes. But now we have what we have: a working and stable GUI-based add-on for OpenOS, and I would be very pleased if you add it to loot disks.

fnuecke:
Oh. I totally skipped the 'add-on' part, did not realize that. Apologies.

Hm. In that case I do have to agree that if this needs to be installed from the interwebs anyway, it might fit better for it to be installable via OPPM.

skyem123:
Then don't call it an "Operating System"! :stuck_out_tongue: 
(seriously though, you are contradicting yourself there)

However, making this work without an internet connection shouldn't be too hard, I'd bet. (unless you really hardcoded some stuff, in which case... ugh)

I might be able to help a bit, but I'm busy for the rest of the month, sadly. also, I can't read your Russian comments and variable names.

IgorTimofeev:
@fnuecke @skyem123 This operating system representing a add-on to another operating system, everything's correct :P

But well, let's say I've done it, and you've added then loot disk. What if I want to change the installer file, change the contents of the loot disk, upgrade standard MineOS files? I need to make another pull-request? I'm not very good at github system, could you explain it?

There is also another curious thing: MineOS uses many standard OpenOS libraries, how about this? Copying them will not create troubles? I know you, foreign developers - you care very much about such things (no offense)

LizzyTrickster:
@IgorTimofeev If you want to update the base floppy, you'd just submit another PR. That's now Plan9k works for it's disk.

skyem123:
@IgorTimofeev, yes every time you want to update MineOS, you need to make a new pull request. However, you only need to do it before an OC release.

Also, you are allowed to use OpenOS libraries and other files from it in your OS, because it is under the MIT Licence, which, to paraphrase, basically means that you need to keep a notice saying that you took code from OpenOS, along with a copy of the licence somewhere, this might help: https://tldrlegal.com/license/mit-license.

IgorTimofeev:
@LizzyTrickster Thanks. And what is Plan9k exactly? Console OS? 

@skyem123 Well, it seems like I have to focus on creating adequate static version of the OS. But if after this Sangar won't add it to loot disk - it will be an insulting insult)

skyem123:
@IgorTimofeev, Plan9k is a completely new OS, which is text based. But it is completely different internally and does use a package management system for updates.

Inari-Whitebear:
I mean, I'm not sure how accurate google translate is. But you might want to look over that license again.

https://github.com/IgorTimofeev/OpenComputers/blob/master/MineOS/License/English.txt

Or the russian version google translated https://translate.google.com/translate?hl=en&sl=ru&tl=en&u=https%3A%2F%2Fgithub.com%2FIgorTimofeev%2FOpenComputers%2Fblob%2Fmaster%2FMineOS%2FLicense%2FRussian.txt

IgorTimofeev:
@skyem123 Okay, need to test it.

@Inari-Whitebear It's normal. English agreement has more gentle interpretation. Russian version is more cruel. I will rewrite it when (and if) MineOS comes to adequate release.

Inari-Whitebear:
Even the English one is... questionable to say the least. And not sure how anal something even enters the case in anything.

skyem123:
@IgorTimofeev, @Inari-Whitebear: The licence is a serious worry here. The fact that it's different between languages is not very good, and might be a legal issue.

Secondly, the current licence does not allow OC to distribute MineOS, as they (the licences) are incompatible. 

Thirdly, the fact that your licence lets you collect things like passwords is a serious issue, that is actually illegal in some countries (in the UK, there is a law called the Data Protection Act that must always be followed. and cannot be ignored, even if the user "agrees"), which could be legally risky for OC itself. While OC wouldn't own any servers said data was sent to, it still would be distributing the software.

IgorTimofeev:
@skyem123 It's just a jokes - I was need to fill license space with some text, so there is it. Do not take everything so seriously.

skyem123:
@IgorTimofeev, then if it's a joke, use a real licence, I recommend the MIT licence, so it's compatible with OC, and move the current licence into a subfolder, such as "`jokes`"?, and have a disclaimer that it's a joke in the file, so people know not to take it seriously. :stuck_out_tongue:

IgorTimofeev:
@skyem123 I don't know anything about licenses. In our country, people do not read the licenses, we just spit on all of these "lawyer stuff". It does not apply to this discussion - I'm telling you this only in educational purposes, to explain why it's a joke. And that's the reason why our nations will not ever understand each others.

Of course, if Sangar will agree to add OS on the loot disk, then I would remove that agreement or, at least, would change its content.

skyem123:
@IgorTimofeev, thing is, in this country, people are _too_ happy to use a lawyer if they want to mess something up. Remember what happened to craftbukkit?

Also, for OC, jokes are in places where it's obvious it's a joke.

IgorTimofeev:
@skyem123 Well, if someone has enough brains to take seriously this agreement, then no problem: all the conditions are displayed by black and white letters, and if user disagree with any of them, he must close the installer. So do not like it - do not use it.

But it's really strange to take seriously an obvious joke.

LizzyTrickster:
The joke may be obvious to you, but to the rest of us it isn't (It also doesn't read like a joke).

skyem123:
Trust me here, it's not obviously a joke.

Inari-Whitebear:
@IgorTimofeev it's fine and all that people can choose to disagree and not install. However, we are concerned here about having such license terms present in the OC repository and loot disks in general.

I'm not sure how @fnuecke sees those of course, but it seems that so far the people here don't like the license much.

IgorTimofeev:
@Inari-Whitebear But it is quite another matter. Simply say "we do not want to see such an uncultured agreement in OpenOS repository" and that's it. 

And, as I said, if Sangar will agree to add OS on the loot disk, then I would change agreement's content, it's not a big deal.

Vexatos:
To throw my two cents in: If the thing depends on an Internet connection, I'd also say that OPPM or mpt would be the right place for this, not a floppy. Furthermore, I I'd recomment working on moving the entire software to English (including comments) so it's easier for people to read through.

IgorTimofeev:
@Vexatos Eh, I just wanted to share pretty operating system with other users, so that they can simply run installer from floppy disk that located at their fingertips instead of "pastebin run" install. 

However, I can see already at least three persons, who doesn't like online-installation (although I still do not understand a logical reason, because in any case the most people will get the latest version of the operating system via the Internet, meaning the offline installation will not be really required). Of course, I can, and most likely will do a standalone installer of this OS, but it is extremely unpleasant to do so solely because of "someone's whims".

This is a very good example of how due to human principles some of good projects simply do not pass "face control" for wide use among the people.

Vexatos:
As an example: You can still install Windows using a DVD and run it offline. Of course it will try to get the latest updates as soon as it's got an internet connection, but it would still work fine without one.

skyem123:
No, actually, the logical reason is that it is inconsistent with the point of loot disks. The point of loot disks is to be useful when there's no internet connection!

IgorTimofeev:
@skyem123 For me and lots of people I know loot disks are some sort of "best of best" programs that written for OpenComputers. Something like hall of fame. Online, not online programs - it doesn't matter.

P.S. Minecraft without internet connection? Oh, please...

skyem123:
@IgorTimofeev, they're really not though. In fact... they tend to be worse than software available online.

Vexatos:
Originally, they were just a bunch of programs examples to get people inspired.

This _really_ looks like something that should rather go onto OPPM or mpt.

IgorTimofeev:
@skyem123 @Vexatos Very sadly. It might be a great way to show everyone how interesting and beautiful software can be made in OpenComputers - and get these people inspired.

skyem123:
@IgorTimofeev, inspiring people involves having source code that people can experiment with.
The loot disks are just that. The full code of the programs on them are visible to see, without needing to install.

IgorTimofeev:
@skyem123 I meant inspired by visual part, not by the code. Code is just a letters, letters can not be beautiful. For example, here is such thing that inspires me:

Inari-Whitebear:
Code can be beautiful. In how it implements certain things. How structured and clean it is. Etc.

"Art is just dots. Dots cannot be beautiful"

skyem123:
It's also worth noting that people should be allowed to experiment with code to learn from it.

IgorTimofeev:
@skyem123 Nobody forbids the user to install the OS through installer and experiment with it's code. Or experiment with the installer code, it's also a good experience to see how online-program works.

@Inari-Whitebear When I open the source code of some OpenOS files - I see a separate pot with a special non-stick coating in hell, prepared for some of its authors. Code must be quick, compact and easy understandable if it's a library or framework, everything else, including "beauty", is a bullshit that depends of an programmer and his desires and coding style.

fnuecke:
Before you start rewriting your whole OS/addon to make it work on a loot disk, I'd like to encourage you to more seriously consider making it available via OPPM. I imagine this would be less of a break in how it works now (since OPPM has modules and even dependencies IIRC, anyway).

Please do not at all take this as a "it's not worthy to be put on a loot disk". They are definitely not a hall of fame; as mentioned by others, for the most part they're deliberately simple programs to give people something to start working _on_ rather than _with_.

OpenOS and Plan9k being the exceptions, so to some degree I'll agree this could fall in the same category as those two, but still, seeing as this is already heavily focused on downloading its components and updating itself (from what I understand), OPPM just seems like an obvious match to me.

fnuecke:
Oh, and about the license thing, if this goes on OPPM it's obviously completely up to you, but either way I'd much appreciate an actual license (just use the MIT one if you don't care anyway), and get in the humor in some other way (or for all I care, just in Russian :P). While this may be an utter non-issue for you, please consider how people from other cultural backgrounds perceive it, and that this will get exposure to people who won't see it as a joke (as obvious from the above comments).

IgorTimofeev:
@fnuecke Too many difficulties for creating loot disks, so to hell with them. It's more interesting to chatting people here and to understand their logic. By the way, I do not know much about the OPPM - no one of my friends uses it, and I also didn't try. It seems to be some kind of program for downloading other programs? No thanks, I satisfied with my pastebin installer. I think, there's no lot of difference between them.

And my OS does not update itself and doesn't require internet card to work. It can get updates only if user wants this. Install OS - work with it as long you need. Want to check updates - click check button - install.

payonel:
@IgorTimofeev I try to be open to criticism, feedback, corrections, and advise. PR any improvements to OpenOS that you have. Keep in mind, a LOT of the strangeness in the code for OpenOS was to minimize memory costs.

Also, I pass on a lot of my OpenOS ideas because they do not fit the concept/purpose the "OpenOS lootdisk". My more advanced ideas are planned for oppm work.

I used to think when my ideas for the OpenOS lootdisk were denied or not favorably met that it was because it was not "hall of fame" worthy. But I honestly do not see it that way anymore, I've come to learn a bit more about the simplicity, and entry level item, we are providing to the user with the lootdisk.

IgorTimofeev:
@payonel Well, at least now I'm more or less understanding your point of view and the concept of loot disks, thanks for explanation.

PR #1861, opened 2016-06-09
openos 1.6 release candidate

### LINKED ISSUE #1827: [Suggestion] Make install.lua support installing any connected loot disk.
Just an idea. @fnuecke's suggestion was adding a `.lootprop` file to every loot disk and having `install.lua` check that. Maybe it could contain installation instructions. This would allow installing all this new loot disk software more easily.

### LINKED ISSUE #1831: The network stack disk install is broken.
When I run installNetwork.lua, it puts stuff on my drive, but it's all under the /data directory.

As in, /data/bin/... is created, and /data/usr/..., etc. I needed to copy the files one level higher by hand.

### LINKED ISSUE #1845: [OC 1.6.0.20 - dev] Interrupt doesn't work
Ctrl + C seems to work but Ctrl + Alt + C also just throws a interrupt event and doesn't force the program to stop.

OC 1.6.0.20-dev
MC 1.9.4

payonel:
install: support rework to search for candidates filesystems and support .lootprop feature, new manual (/usr/man/install) file added
closes #1827 

term: fix to not consume(hide) hard interrupt in term.pull, and new method term.scroll for vertical scrolling
closes #1845 

cat: default to stdin with no args

cp: support "contents of" /. paths, correct verbose output, prefer link copy over directory, error on missing source files
closes #1831 

df: support relative paths

guid: correct guid format (8-4-4-4-12)\

head: close stdin

less: new pager option, supports scrollback and mouse scroll

ln: use source name when omitted

mv: report error when target path is read only, and report error when source path is a mount point

profile: remove less alias (we have /bin/less now)

rm: fix issue where links to dirs could not be removed

serialization: inline pairs in case of [C] boundary issue with delay loaded libraries

/lib/sh: support white space in tab completion with back slashes, emulate SIGPIPE on closed pipe writes,

shell: allow calls to getWorkingDirectory during boot before PWD is set (defaults to /)

text: advanced internal tokenizer upgrade to more fully support whitespace metadata needs of the shell

transforms: new api, at(tbl, index), to return key-value pairs by index

other files: fix formatting and typos


payonel:
btw, this last commit made it possible to finish an install on a single T1 ram stick

fnuecke:
As usual, thanks a bunch! I adjusted the old loot disks (software disks?) to work well with the new install (i.e. directory structure and removed autorun adding symlinks).

PR #1876, opened 2016-06-12
regression fix: add check for nil due to split code

payonel:
oops

btw, install still works, even when it crashes: e.g. install --help

It's just that it doesn't close nicely

So, this isn't super high priority :) must have before final release, but not for beta

Maybe just makes the new install look bad :/

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/network/.prop
+++ b/src/main/resources/assets/opencomputers/loot/network/.prop
@@ -1,0 +1,1 @@
+{fromDir="/data/"}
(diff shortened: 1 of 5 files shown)


payonel:
also added some other issues being reported by users:

install was crashing if there were no options. Having no options is a perfectly valid case where install is trying to exit early (e.g. --help, or user cancelled) I forgot the check for nil.

During tests I found that cp was unnecessarily slow due to constant event pulling on every file. I introduced a 4 second gap between checking events. But his caused cp very hard (nearly impossible) to interrupt. Thus, it is preferrable to slow down copy so that the user could interrupt it.

Also, thanks to Inari, we have improved the text of the install prompts. Please review and comment if you feel the text still needs some revising. I'm not a technical writer, so I hope we are making this simple.

Also, if the user tries to run install with the network loot disk, the /data dir will be copied to /, and the user probably wants the data/. dir used. Using .prop's {fromDir="/data/"} fixes  this

fnuecke:
Thanks as usual :)

PR #1900, opened 2016-06-21
changes to pwd and /mnt

payonel:
Functional and objective fixes moved to #1903

filesystem: don't mount tmpfs in /mnt
~~buffer: don't allow reading zero bytes from a stream, and fix libraries that should be allowed to set zero size buffer~~
~~install: report setting boot address, ^d should not accept prompts~~
~~ls: use shell to get working dir, not PWD~~
~~package: remove clutter for library keys and _G~~
~~profile: PS1 to use : instead of #~~
~~rc: add unload to lib methods and fix typo in source~~
~~sh: remove clutter from _G and use zero sized buffer for piping~~
shell: remove trailing slash for pwd

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/boot/90_filesystem.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/boot/90_filesystem.lua
@@ -22,7 +22,7 @@ local function onInit()
 end
 
 local function onComponentAdded(_, address, componentType)
-  if componentType == "filesystem" then
+  if componentType == "filesystem" and require("computer").tmpAddress() ~= address then
     local proxy = component.proxy(address)
     if proxy then
       local name = address:sub(1, 3)
--- a/src/main/resources/assets/opencomputers/loot/openos/etc/profile
+++ b/src/main/resources/assets/opencomputers/loot/openos/etc/profile
@@ -17,7 +17,7 @@ set HOME=/home
 set IFS=\ 
 set MANPATH=/usr/man:.
 set PAGER=/bin/more
-set PS1='$PWD# '
+set PS1='$PWD # '
 set PWD=/
 set SHELL=/bin/sh
 set LS_COLORS="{FILE=0xFFFFFF,DIR=0x66CCFF,LINK=0xFFAA00,['*.lua']=0x00FF00}"
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/shell.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/shell.lua
@@ -127,13 +127,15 @@ function shell.resolveAlias(command, args)
 end
 
 function shell.getWorkingDirectory()
+  -- if no env PWD default to /
   return os.getenv("PWD") or "/"
 end
 
 function shell.setWorkingDirectory(dir)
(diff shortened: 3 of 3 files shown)


skyem123:
Wait... Why should the tmpfs not be in `/mnt`?

payonel:
It is good to have a conversation about these changes, thanks for asking @skyem123 - before I give my opinion I'll also point out that I'm happy to take counsel and feedback about these changes.

/mnt is for user (or system auto)mounted filesystems. Because virtual filesystems are not disks, nor floppies, nor tapes, nor cdroms, etc which are physically added to the system, I feel they don't hold the same promotion.

http://www.linfo.org/mnt.html doesn't say tmpfs breaks this rule, but it does feel unnatural to me to have virtual filesystems that already have real homes (tmpfs on /tmp and devfs on /dev) also in /mnt

skyem123:
In the early days of OC, when everything was in ROM, both the ROM and the tmpfs (aka RAM) were in `/mnt`

xarses:
The # in PS1 was fine just not the trailing slash before it `/home#` is fine, `/home/#` make it look like we are in home's subdirectory `#`. If this is still the case, I'll have the same problem with moving it to `:`

payonel:
i changed the # to : because i just think it looks nicer, but, I'm happy to be voted out on that
the trialing / was fixed separately (to $PWD and pwd) which is what PS1 prints. This matches bash behavior

payonel:
apparently # is default for root and $ is for users, and : is something i must have inherited from ages past on who-knows-what-distro and just always preferred it.

anywho, reverted the PS1 change. still, $PWD fixed still

skyem123:
I prefer `/home/#` to `/home#` because the `something/#` looks neater than `something#` to me.

SaphireLattice:
Proposal: customizable PS1 string? Or is it already customizable using
bash-like tokens?

payonel:
@dangranos PS1 is definitely customizable, it is set in `/etc/profile` and it is recommended to make changes to defaults in `/home/.shrc` -- so one could add `set PS1='$PWD% '` to `/home/.shrc` if they wanted. The slight issue with having or not having the trailing / is that we don't have a simple way to leave that up to the user, `$PWD` and `/bin/pwd` are set by the os. This discussion affects what they return (to trail or not to trial). One could write their own pwd in (for example) /usr/bin/pwd.lua to customize trailing /, and then set PS1 to '`/usr/bin/pwd.lua`# '

@skyem123: linux+bash (my target behavior) does not return / from `$PWD` nor `pwd` - like I said to @dangranos about custom PS1's, maybe the right solution is:

`/usr/bin/pwd.lua`

```
return require("shell").getWorkingDirectory():gsub("([^/])$", "%1/")
```

`/home/.shrc`

```
set PS1='`/usr/bin/pwd`# '
```

fnuecke:
Let me know when you've all reached a consensus and I'll merge it in ;-)

payonel:
If we agree on any form of this PR -- please note it is DEPENDENT on #1903 
I decided  to chain these PRs because this PR on its own does not pass all my unit tests.

xarses:
I'm happy with this combination, it follows standards established in *NIX systems

PR #1906, opened 2016-06-24
Send internet events when ready to read

magik6k:
Now, when TCP connection is ready to be read, card emits 'internet_ready' signal.


fnuecke:
Thanks. Ported it over to the 1.7.10 branch: 422592d84e83c5c293bd0eb5e5f103ab590fa076
(And pulled up into the other ones from there.)

PR #1934, opened 2016-07-04
a possible too-many-screens' solution

### LINKED ISSUE #1933: too many screens
In testing my booth for BTM, I found that openos really doesn't behave nicely with multiple screens and multiple keyboards.
1. If you attached two screens and a keyboard to just one screen - boot the computer, and then switch keytboards to the 2nd screen (Removing it from the first), then trying to use tab complete will crash the shell. This is because the shell is trying to check the keyboard shift key state, and there is no keyboard on the bound screen.
2. if you attach multiple keyboards to a single screen, the terminal handles a key_down even for each keyboard.
3. if you attach mutliple screens to a computer (either all with keyboards, or none) the boot status will render to one screen, but the shell will prompt on a different machine.

payonel:
**Recommended for post 1.6 release**
This PR might be best for post 1.6 release because it is rewriting some rather critical boot level operations (although, completely backwards compatible). In my testing I have found this stable and accurate, but this PR fixes a low priority issue with some risky level code.

This fixes #1933

init: remove primary component intialization from init, and give full component control to component driver (it already had the primary component code, init was trying to accomplish the same logic)

04_component: add screen and keyboard "upgrading" to the component.setPrimary code path. There are cases where a keyboard or a screen should become the new primary component, and other cases where they should be ignored. This logic is based on permutations where a user boot with screens and keyboards and adds/removes screens and keyboards after boot. Keyboard may and may not be attached to screens in all permutations. The experience should be predictable for the user, and always the primary screen should match the primary keyboard as a pair when the primary screen has a keyboard

04_component: add logic to reuse the boot screen as the first primary screen option

term library: cache the screen-attached keyboard and check keyboard events against it
term library: remove cached keyboards from the terminal on component_removed -- this forces the terminal to recache the keyboard
term library: assume keyboard unknown on start - this allows keyboard removals and additions to not interfere with terminal operations

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/boot/04_component.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/boot/04_component.lua
@@ -101,6 +101,7 @@ function component.setPrimary(componentType, address)
     if wasAvailable or wasAdding then
       adding[componentType] = {
         address=address,
+        proxy = primary,
         timer=event.timer(0.1, function()
           adding[componentType] = nil
           primaries[componentType] = primary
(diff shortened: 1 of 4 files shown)


Vexatos:
Actually, OC 1.6 is the major breaking update, so why not put it in there? Noone knows when/if OC 1.7/2.0 is going to happen.

payonel:
I think this fix is good, but in case it is buggy I didn't want to delay 1.6 over something that is an edge case fix (i.e. multiple screens and multiple keyboards). I"m happy with what sangar decides on this. In the meantime, I'm using it at my btm booth and will continue to test it.

payonel:
this isn't ready, found some bugs. i'll revise and make a new pr

PR #1948, opened 2016-07-09
The too-many-screens' solution

### LINKED ISSUE #1933: too many screens

payonel:
boot screen is preserved
screen with keyboard always gets keyboard focus
lost screens automatically switch to available screens
screens without keyboards that are primary screens correctly take keyboard focus if given a keyboard
and closes #1933

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
@@ -430,7 +430,7 @@ local function find()
     setStatus("Find: " .. findText)
 
     local _, address, char, code = term.pull("key_down")
-    if address == term.keyboard().address then
+    if address == term.keyboard() then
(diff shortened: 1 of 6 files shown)


payonel:
separating other (awesome) fixes to a 2nd PR: #1954

fnuecke:
Thanks!

PR #1954, opened 2016-07-11
Needful things

payonel:
This PR is rebased on #1948 -- it is expected you merge #1948 **first**

still finding very helpful fixes

support cp -r . ../path which should copy "contents of" current dir 
rename guid to uuid, and remove useless toHex method
update devfs to support calling handler open and close
move install confirmation before customer .install execution
fix glob that may contain . and .. in path


payonel:
this PR just isn't ready, and just in case @fnuecke comes back and merges this before I'm able to finalize it ... I think it best to close it and reopen it when I feel it is ready for rc

PR #1959, opened 2016-07-13
Fix internet TCP events

magik6k:
It's impossible to add channel SelectionKey back to selector if it was cancelled. The only workaround to his seems to be creating new selector each time. Other way this could be done would be to remove op_read from keys' interest ops, but that broke java select implementation on Linux. This seems to be the only way to get this kind of behavior. With usual usage of TCP in OC it shouldn't use much CPU.


fnuecke:
Thanks!

PR #1961, opened 2016-07-13
The most important fix in the history of this mod.

Vexatos:
Unacceptable!

### DIFF
--- a/src/main/resources/mcmod.info
+++ b/src/main/resources/mcmod.info
@@ -5,7 +5,7 @@
   "version": "${version}",
   "mcversion": "${mcversion}",
   "url": "http://oc.cil.li/",
-  "authorList": ["Florian 'Sangar' Nuecke", "Johannes 'Lord Joda' Lohrer", "Everyone who contributed to the mod on Github - thank you!"],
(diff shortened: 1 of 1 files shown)


fnuecke:
No, in some cases it needed to be UTF-8 to show up correctly, in some CP-1252 or whatever, this is good enough.

gjgfuj:
Wait, are you saying ü isn't utf-8? Unicode text format 8? ???. Do you mean
ASCII?

skyem123:
In Minecraft, it would render the ü in a different font to the rest of the text.

dgelessus:
@skyem123 No, I think most Latin characters with diacritics (`üùúû` etc.) are displayed in the default Minecraft font.

@gjgfuj I think he means that some things decode the text as Latin-1 or Windows-1252 instead of UTF-8, which would mess up the Ã¼.

fnuecke:
What @dgelessus said. I know that in some cases MC decoded it one way and in others the other (though I'm not 100% sure what the cases were, i.e. whether it was built on Jenkins vs. local or Windows vs. Linux and so on). Either way, it's something I can't be bothered with anymore >_>

dgelessus:
This is a JSON file, right? In that case you could use `\u00fc` in place of `ü`.

PR #1962, opened 2016-07-13
adding native lib for sunos derivatives, built and tested on smartos

glasspelican:
should function on all sunos derivatives including illumos, smartos, smart data center, openindiana, and solaris

built and tested on smart data center smartos instance 5.11 joyent_20160627T182738Z using oracle jdk-8u92-solaris and gcc 4.9.3 from pkgsrc

the only changes i needed to make to get the native libs to build where in the build.gradle file, the one i used can be viewed here https://gist.github.com/glasspelican/d65eb2d3a5bd8f327ede58a55819b4eb


technostar98:
Would love to see this happen!

fnuecke:
Cool, thanks! I'll be looking into separating the natives from the main mod JAR soon, possibly before 1.6 (yay, more delays :X), I'd love to merge it into that then (I'd like to avoid blowing up the JAR size until then, I hope you understand).

glasspelican:
anything i can do to get this moving?

asiekierka:
This should be closed, OpenComputers provides a way to force-feed a native library name and SunOS is not an officially supported Minecraft platform, so I don't think we should go out of our way to support it either.

Deep apologies, but the mod isn't doing well (low developer count) and the plate is plenty full already.

PR #1964, opened 2016-07-15
Fixed Navigation Upgrade recipe.

Vexatos:
As it turns out, has always had the wrong recipe... ever since it was created in 2013.


fnuecke:
Kewl.

PR #1965, opened 2016-07-15
Allow properly launching OC in a dev env again.

Vexatos:


fnuecke:
Is there no less ugly way of doing this? :/ Not that I can think of one myself right now... some way to make IDEA also do the replacements on build maybe?

fnuecke:
Bah, whatever :X

PR #1967, opened 2016-07-16
improvements to guid, devfs, cat, cp, and system error clarity

payonel:
renaming guid to uuid and removing useless method toHex
adding text read and writer, compare to c++ stdlib stringstream. needed for advanced devfs
fix globbing with relative paths and . and .. in path
moving install confirmation before calling .install
fix cp: cp -r . /another/path (should copy contents of current dir)
fix cat when stdin is missing or closed
improved devfs supporting custom dir handlers for dynamic file listing

other non functional changes to clean up code and improved error messages


fnuecke:
Looking great, thanks as always! :)

PR #1969, opened 2016-07-16
Added JEI handlers

Vexatos:
Basically the NEI handlers, but re-done for JEI.

`CallbackDocHandler` is a little more pretty than it used to be.

`OredictTooltipHandler` has not been implemented (yet) for lack of config access.

Item highlighting only works one-way for lack of methods to access the selected overlay slot.

Focused input check has not been added for lack of methods to do so.

mezz/JustEnoughItems#203 May be interesting if the caching doesn't work too well.


fnuecke:
Thanks a bunch!

PR #1973, opened 2016-07-17
improved terminal interruptions with smart shell support

payonel:
Also:

Fixed a bug with command substitution not concatenating strings
Split /bin/ls for ultra simple formatting for low mem systems
Remove all instances of absolute paths to lib/tools and use package searching


payonel:
@fnuecke i like this PR

couple things that make this really nice
1. /bin/list for low memory systems (if `ls` fails to load, it suggests to the user to use `list`)
2. soft and hard interrupts are handled nicely. i had been using term to handle interrupts during term.read. This was wrong, it should have closed/aborted the term.read, and shell should be handling the interrupts. This way, applications using term.read won't get back valid looking data when the user had actually wanted to interrupt. Also, term.read no longer consumes hard interrupts, which I think is definitely what our users want

PR #1989, opened 2016-07-25
Change scheduleIC2Add to use IEnergySink instead of IEnergyTile

### LINKED ISSUE #1937: [1.10.2] Computer system not accepting power
This bug was observed on a testing server for ME^EX 1.10.2, our curse modpack, which is running the latest version of OpenComputers for MineCraft 1.10.2. In essence, I was not able to feed power into my capacitor, my power converter, or my tier 3 computer case, the only OC objects I tried it with. I tried to feed power in using the following methods, each one resulting with total failure: a hand cranked generator as well as an advanced power cube from Calculator, an enhanced energy conduit and a basic capacitor bank from EnderIO, a clockwork engine from forestry, a generator from TechReborn (I know EU isn't supposed to work, but I tried it anyways), and a power relay from RFTools. I am most certainly positive that every system I tried had power to supply, but none of the apparatus would accept it: conduits wouldn't connect, and clockwork engine didn't auto-align; further, an analyzer and many failed attempts at starting my computer confirm that it indeed accepted none of my power. If you wish to join the server to see for yourself, please give me your MineCraft IGN so I can whitelist you.

~Sam

Spyboticsguy:
Fixes #1937.
Tested and confirmed working with IC2-exp v2.6.15 as well as EnderIO v3.0.1.73 for 1.10.2.

### DIFF
--- a/src/main/scala/li/cil/oc/common/EventHandler.scala
+++ b/src/main/scala/li/cil/oc/common/EventHandler.scala
@@ -113,7 +113,7 @@ object EventHandler {
   def scheduleIC2Add(tileEntity: power.IndustrialCraft2Experimental) {
     if (SideTracker.isServer) pendingServer.synchronized {
       pendingServer += (() => if (!tileEntity.addedToIC2PowerGrid && !tileEntity.isInvalid) {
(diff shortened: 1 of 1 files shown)


fnuecke:
Thanks!

PR #1992, opened 2016-07-29
math.pow() was removed in Lua 5.3

Griptzion:
math.pow() no longer exists in Lua 5.3, causing errors when attempting to use function calls such as note.play(). The caret '^' will work for both 5.2 and 5.3 architectures.


fnuecke:
Ah. I decided to fix it more globally via a fallback method added by the sandbox, but thanks for pointing this out!

PR #1993, opened 2016-07-31
Added whitelisting of debug card owners.

makkarpov:
Extended current debug card binding system with whitelisting capability.

The features of this implementation are:
1. Whitelist can be managed with `/oc_debugWhitelist` command.
2. Losing bound card is not a big issue, since all cards could be revoked with `/oc_debugWhitelist revoke` (user can revoke own cards even without OP permissions) and must be rebound then.

image


makkarpov:
Anyone here? Two weeks passed without any reaction.

PR #1994, opened 2016-07-31
Added SignChangeEvent to sign upgrade

makkarpov:
It would be useful for server (especially Bukkit, but maybe some others too) where plugins treat specific signs in a special manner (e.g. shops or warps). Without this event it's hard to prevent robots from creating these signs, this event introduces an explicit capability to do so.


Vexatos:
The event is not `@Cancelable`. Also, you could simply check for the boolean returned by `post()`.

Issue #2016, opened 2016-08-13
Spam in Logfile if HQM is installed


feldim2425:
This error is repeats like 1000 times. It only comes when I start minecraft there is no error after it has started.

```
java.lang.NullPointerException
    at li.cil.oc.util.Tooltip$.get(Tooltip.scala:21) ~[Tooltip$.class:?]
    at li.cil.oc.common.item.traits.Delegate$class.tooltipLines(Delegate.scala:86) ~[Delegate$class.class:?]
    at li.cil.oc.common.item.ControlUnit.tooltipLines(ControlUnit.scala:3) ~[ControlUnit.class:?]
    at li.cil.oc.common.item.Delegator.func_77624_a(Delegator.scala:181) [Delegator.class:?]
    at net.minecraft.item.ItemStack.func_82840_a(ItemStack.java:610) [adz.class:?]
    at hardcorequesting.client.interfaces.edit.GuiEditMenuItem$Search.initItems(GuiEditMenuItem.java:502) [GuiEditMenuItem$Search.class:?]
    at hardcorequesting.quests.QuestLine.loadAll(QuestLine.java:122) [QuestLine.class:?]
    at hardcorequesting.quests.QuestLine.init(QuestLine.java:130) [QuestLine.class:?]
    at hardcorequesting.HardcoreQuesting.preInit(HardcoreQuesting.java:52) [HardcoreQuesting.class:?]
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_91]
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_91]
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_91]
    at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_91]
    at net.minecraftforge.fml.common.FMLModContainer.handleModStateEvent(FMLModContainer.java:595) [FMLModContainer.class:?]
    at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source) ~[?:?]
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_91]
    at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_91]
[...]

PR #2018, opened 2016-08-14
Fix Exception spam if HQM is installed

feldim2425:
Just some additional null checks because HQM wants the Item Tooltip in PreInit.
Fix for #2016


fnuecke:
Ah, sorry, didn't see the crossreference when I saw the issue. Oh well, fixed either way. Thanks nonetheless!

PR #2060, opened 2016-09-17
Check if the world is remote before killing the drone

Polyana-Fontes:
Drones were getting invisible when a player attempts to drop it (wrench + sneak + right click) and a server mod cancels the interaction.


fnuecke:
Thanks!

PR #2074, opened 2016-10-02
Update EnderStorage driver to 1.10 API

thiakil:
I ported forward the EnderStorage driver to the latest 1.10 API, and changed/added some of the its component api functions.

Full commit history is at https://github.com/thiakil/OpenComputers/tree/enderstorage-1.10


fnuecke:
Thanks!

Issue #2065, opened 2016-09-23
Request - inventory.getAllStacks() for WorldInventoryAnalytics (Inventory controller/transposer/etc)


thiakil:
I'm looking to make a form of inventory management with OC, but enumerating itemstacks in an inventory takes quite a while with getStackInSlot (around 1min for 300 slots). 

getAllStacks from the inventory driver can get it instantly.

I'm not sure whether this is from the Converters adding their info to the table (getAllStacks seems to not do this), or from the context switch. If it is, perhaps also an option in getStackInSlot to not do this?

Vexatos:
I guess this can be closed now.

PR #2075, opened 2016-10-02
World Inventory Analytics Improvements

thiakil:
I have ported getAllStacks & getInventoryName from the vanilla inventory driver so that Inventory Controller Upgrades, Transposers, etc can use those functions.

My Reasoning is in #2065
My function also has a benefit of not fooling Lua with array size, an empty slot with the Vanilla driver seems to stop iterating upon a nil element. My function gives it an empty table (as ItemStacks show as tables in Lua getStackInSlot()).

I did getInventoryName because it also seems useful to know if we're interacting with the correct inventory.

### DIFF
--- a/src/main/java/li/cil/oc/api/prefab/ItemStackArrayValue.java
+++ b/src/main/java/li/cil/oc/api/prefab/ItemStackArrayValue.java
@@ -1,0 +1,136 @@
+package li.cil.oc.api.prefab;
+
+import li.cil.oc.api.machine.Arguments;
+import li.cil.oc.api.machine.Callback;
+import li.cil.oc.api.machine.Context;
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTBase;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTTagList;
+
+import java.util.HashMap;
+import java.util.TreeMap;
+
+public class ItemStackArrayValue extends AbstractValue {
+
+	private ItemStack[] array = null;
+	private int iteratorIndex;
+
+	private static final byte TAGLIST_ID = (new NBTTagList()).getId();
+	private static final byte COMPOUND_ID = (new NBTTagCompound()).getId();
+	private static final String ARRAY_KEY = "Array";
+	private static final String INDEX_KEY = "Index";
+
+	private static final HashMap<Object,Object> emptyMap = new HashMap<Object,Object>();
+
+	public ItemStackArrayValue(ItemStack[] arr){
+		if (arr != null){
+			this.array = new ItemStack[arr.length];
+			for (int i=0; i< arr.length; i++){
+				this.array[i] = arr[i] != null ? arr[i].copy() : null;
+			}
+		}
+		this.iteratorIndex = 0;
+	}
+
+	public ItemStackArrayValue(){
+		this(null);
+	}
+
+	@Override
+	public Object[] call(Context context, Arguments arguments) {
+		if (this.array == null)
+			return null;
+		if (this.iteratorIndex >= this.array.length)
+			return null;
+		int index = this.iteratorIndex++;
+		if (this.array[index] == null)//TODO 1.11 change to ItemStack.EMPTY?
+			return new Object[]{ emptyMap };
+		return new Object[]{ this.array[index] != null ? this.array[index] : emptyMap };
+	}
+
+	@Override
+	public Object apply(Context context, Arguments arguments) {
+		if (arguments.count() == 0 || this.array == null)
+			return null;
+		if (arguments.isInteger(0)){//index access
+			int luaIndex = arguments.checkInteger(0);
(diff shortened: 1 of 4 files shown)


Vexatos:
I don't think that `getInventoryName` should be a thing. It's not very good for immersion as it would expose an internal name.

thiakil:
What about if we make it get the localised block/tile label? I was wanting to let a Transposer/Adapter identify what kind of inventory is at the particular side.

For instance currently one needs to know in advance that there is an Ender Chest on the top side, and a normal chest below.

Vexatos:
@thiakil localized label can't work either as localization is undefined on the server side. It is a client-side only thing.

thiakil:
Is there some kind of way of identifying the block? EnderStorage returns "Ender Chest" for the custom inventory name returned by this, so I had incorrectly assumed that we'd get something usable like "Chest" from a vanilla chest

Vexatos:
@thiakil But what happens once you change the language of the game? Suddenly it makes no sense anymore.

Relying on localization server-side is generally a bad idea. Nothing like names should be exposed from OC, in my opinion.

thiakil:
Names/IDs are already exposed when allowItemStackInspection is enabled, else we need to severely restrict what we can do to specific slots, blindly moving between them; I don't really see much difference with the inventory's name.

Would you rather I remove it?

Vexatos:
Honestly, I'd prefer item name exposure to be removed too, but @fnuecke doesn't want that and, in the end, it's his mod so he has the final say. I'd wait for his opinion on this.

xarses:
I'm not sure how I feel about returning a single object that would get to be 300+ stacks, this way seems memory intensive. What are your thoughts on returning an iterator object instead?

Vexatos:
This does sound like a much better idea.

thiakil:
I don't have a huge issue with that but:
- is there elsewhere in oc that does this? (it it even possible without resorting to ugly functional programming with a 'cookie' value?) 
- The old ME driver does it the same as mine for a potentially larger result set (minus the filter option which doesn't make sense here) 
- With thousands of item types the ME driver would rightly trigger a memory error on a tier one computer but worked well within the default memory limits on a tier 3 when I tried it (this is without a filter) 

I'm running on default settings and can easily make a big table of 5 x 300 stack capacity inventories.

thiakil:
Actually, now that I'm more awake... 
- An iterator would be a little redundant as you can already do this with getStackInSlot and getInventorySize
- Further to my how would it work question above, unless you do it outside the component you'd still get the performance hit from context switching that this function aims to avoid (correct me if that's wrong). Such I think would be outside the scope of this pull request

Vexatos:
> is there elsewhere in oc that does this? (it it even possible without resorting to ugly functional programming with a 'cookie' value?)

Userdata, see `component.list`.

> An iterator would be a little redundant as you can already do this with getStackInSlot and getInventorySize

Would still be a lot faster if cached.

thiakil:
And you can still iterate in plain Lua with the returned table. I fail to see how that changes memory requirements, which I don't even think is much of an issue to begin with

Vexatos:
The difference is that Lua only knows about a pointer to the userdata instead of everything, meaning much less RAM is used by Lua itself.

thiakil:
Finally getting back to this after a while (pointing to componenet.list is rather confusing as it uses raw lua api pushing)

How would I convert an ItemStack to a value I can return to Lua from Java? returning one from Value.call or Value.apply doesn't work, the log say ItemStack is not supported. in Scala it seems to be ResultWrapper that does it somehow?

I'm adding to the Java API because the object seems like a good candidate to be reused both internally (e.g. AE2 integration) and externally.

thiakil:
Seems I found it; Value objects didn't seem to be going through Registry.convert. See dd08424 for my solution.

I have rebased the pull request & updated to use a userdata Value.

    it = inventory.getAllStacks(side)
    =it() -- get the current table (empty if null) and advance the internal index
    =it.n or =it.count() -- get the size of the internal array
    =it[n] -- get element n
    it.reset() -- reset the internal pointer to 0 (reusable iterator)
    =it.getAll() -- get the full array as a lua table (like the function originally did). I don't see much use for this as you can use the object much like a Lua list, but it's there in case it's wanted by someone.

The Value object is in the API so others can use, it simply needs to be constructed as `ItemStackArrayValue(ItemStack[])` and can then just be passed back to Lua. A deep copy is performed on the array, so the ItemStacks can change after it is called without worry. In future we should use this for the AE2 integration as counting all the items in the system was able to crash the computer with an out of memory error.

Please really have a look at https://github.com/MightyPirates/OpenComputers/pull/2075/commits/1273968142e3614457c6c2a0e5649e9da43cabe7#diff-ca6d9df7dc7aea567235b4fbd3061a0fR97 (getInventoryName) as Scala is definitely not a well known language to me.

thiakil:
Ping @Vexatos

Vexatos:
I'm not the right person to review this.

PR #2078, opened 2016-10-03
Added further functionality to the player object in the debug card

InfinitySamurai:
Please criticize this all you like. This is my first attempt at contributing so there may well be issues


fnuecke:
Looks mostly fine, thanks. Just made a few minor adjustments.

PR #2080, opened 2016-10-04
Created and added functionality to debug Scoreboard class

InfinitySamurai:
As with my other pull request, please critique as needed, this may well not be the best way to do this.


InfinitySamurai:
Would I be able to get some feedback as to why this wasn't merged? @fnuecke

fnuecke:
Hey, super sorry for the late reply. Looks good, except for the save/load code but I'll fix that real quick. Thanks!

Issue #1866, opened 2016-06-10
"Driver" disks are not intuitive at all right now


skyem123:
So... I've had to walk someone though installing a "driver", and it was not fun.
So I have a list of things that, if solved, would make it a lot more intuitive.
1. How does one get these disks. I know it's possible to do by cycling loot disks, but that means that they're impossible to get if that setting is disabled. Maybe craft a disk with a component, with the component staying in the crafting grid?
2. There are probably too many disks. :P OC should just have one driver disk, as that'd make driver installation a lot easier.
3. This also means that there needs to be some way for driver disks to detect what hardware there is and refusing to install (unless forced) if it isn't present.
4. Also, there should be an easy way of installing drivers. Suggestion for @payonel: there should be a utility that scans for loot / driver disks and prompts the user to install them.

Sorry if this sounds a bit rude, I'm tired and I thought that I should write this down before I forget it.

payonel:
the work on installer is nearly done, but it takes more than just a couple nights. Sorry for the delay. But the new install tool will HELP. Let's discuss these pain points individually, and I'll speak to what is already solved with the upcoming install
1. Sangar has added an api that I can call from OpenOS to help find what disk is need. When someone tries to run **xyz**, and the file is not found on their system, I can check a lookup table and report what disk they need to craft. I haven't resolved everything you're saying in your item 1, but this is part of the same problem I feel. [OpenOS support of this feature will not be available in the openos-1.6.-rc, but an update following it]
2. I have nothing in the works to resolve this, but we talked in irc about having a driver folder in the openos loot disk that would NOT install by default. We could check what components are already in the system during the main os install, and include needed drivers (this would be easy). We could also rescan the system every time install is run with the openos disk (also easy, but doing this may not be as intuitive to users).
3. This is definitely doable. The next update to openos will introduce support for "driver disks" to provide custom .install scripts. The .install script can be hand tailored to achieve this purpose. Though I'm not sure it is necessary. But it doesn't matter to me what component driver devs do with their own .install (note .install is OPTIONAL, there is definitely default behavior that /bin/install can perform that is absolutely valid)
4. That is exactly the work in progress. I think you'll like it. I have something in the works locally that is already better, but here is the current PR of the next install (again, this is now outdated, a better upgrade even is soon to come [probably this weekend]): https://git.io/voZWU

skyem123:
Yay for the installer!

fnuecke:
[...]

PR #2109, opened 2016-10-27
wget and pastebin are now core to openos install

payonel:
Working on #1866 

Just making sure I'm making the correct change here -- but wget and pastebin are installed with openos now so there isn't a need to have them in the program locations table


payonel:
I added the openos support for getProgramLocations()

PR #2131, opened 2016-11-14
Fix Arbitrary Read/Write exploit

### LINKED ISSUE #2116: Arbitrary Read (and limited write) exploit in Open Computers
The path for disk drives is not sanitized to stop ../ elevating the path, if the address is modified (either with debug card or with custom mod) to let's say "HAX/../.." then the /mnt/HAX will show the world save folder which you can freely read from. Because the size of this folder exceeds the max size of the disk, it is impossible to write to it, but you can write to any area with smaller size and you could probably delete files too. NBT is not safe when in creative mode as any stack can be modified via the C10PacketCreativeInventoryAction. There are similar exploits in many mods, but I have not seen any that can be so malicious.
Also note too many ../ will crash server as it would iterate through every file and folder in that directory without sleeping.

gamax92:
This fixes #2116 by validating the address stored in hard disks and floppy disks if it exists and generate a new address if it has been tampered with.

A thing to note, this only fixes being able to exit the sandbox and access host files, but drives can still be modified to access other drives contents, given a valid but different UUID.

EDIT: This does not check for uppercase hex characters, since by specification: "Software generating the hexadecimal representation of a UUID shall not use upper case letters." of which Java complies with.
Any upper case letters in an address would indicate it's been tampered with and could cause issues on case sensitive filesystems.

### DIFF
--- a/src/main/scala/li/cil/oc/integration/opencomputers/DriverFileSystem.scala
+++ b/src/main/scala/li/cil/oc/integration/opencomputers/DriverFileSystem.scala
@@ -2,6 +2,7 @@ package li.cil.oc.integration.opencomputers
 
 import li.cil.oc
 import li.cil.oc.Constants
+import li.cil.oc.OpenComputers
 import li.cil.oc.Settings
 import li.cil.oc.api
 import li.cil.oc.api.network.EnvironmentHost
@@ -17,6 +18,8 @@ import net.minecraft.item.ItemStack
 import net.minecraft.nbt.NBTTagCompound
 
 object DriverFileSystem extends Item {
+  val UUIDVerifier = """^([0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12})$""".r
+
   override def worksWith(stack: ItemStack) = isOneOf(stack,
     api.Items.get(Constants.ItemName.HDDTier1),
     api.Items.get(Constants.ItemName.HDDTier2),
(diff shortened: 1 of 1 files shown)


src/main/scala/li/cil/oc/integration/opencomputers/DriverFileSystem.scala
```diff
@@ -17,6 +17,8 @@ import net.minecraft.item.ItemStack
 import net.minecraft.nbt.NBTTagCompound
 
 object DriverFileSystem extends Item {
+  val UUIDVerifier = """^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$""".r
+
```
magik6k, line 22: `^([0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12})$`

src/main/scala/li/cil/oc/integration/opencomputers/DriverFileSystem.scala
```diff
@@ -82,7 +84,13 @@ object DriverFileSystem extends Item {
 
   private def addressFromTag(tag: NBTTagCompound) =
     if (tag.hasKey("node") && tag.getCompoundTag("node").hasKey("address")) {
-      tag.getCompoundTag("node").getString("address")
+      UUIDVerifier.findFirstIn(tag.getCompoundTag("node").getString("address")) match {
+        case Some(address) => address
```
magik6k:
``` scala
tag.getCompoundTag("node").getString("address") match {
  case UUIDVerifier(address) => address
  case _ =>  [...]
```

magik6k (commented):
Thanks. This would probably already work, but I prefer matching this way. Could you also add some warning when new UUID is generated?

gamax92:
Changed the regex, changed the matching style, and added a warning message as suggested.

magik6k (approved):

payonel:
tested and works great, thanks @gamax92 and @magik6k

PR #2134, opened 2016-11-16
Re-added EnderIO integration and added Bundled Redstone support for EnderIO's Insulated Redstone Conduits

### LINKED ISSUE #613: Bundled Redstone Cable Support Request: EnderIO
I would like to request support for the [EnderIO Insulated Redstone Conduit](http://wiki.enderio.com/Insulated_Redstone_Conduit).  It supports 16 colored channels, similar to items like Rednet cable already supported by OC.  The catch is that EnderIO does not seem to have an API.  The code is [available on GitHub](https://github.com/CrazyPants/EnderIO/), though, so that should at least help.

I mentioned this idea on IRC and it sounds like the feasibility of this feature will depend on how easy it is to interact with EnderIO.  Here's to hoping it is easy!

For what it's worth, I tried a test with an OC computer with Tier 2 Redstone Card outputting a strong redstone signal on the White channel of a bundled cable and connecting via Rednet cable and then via EnderIO Insulated Redstone Conduit to a Redstone Lamp - the Rednet cable works great and lights up the lamp, but the EnderIO Insulated Redstone Conduit does not light up the lamp.  This was on Minecraft 1.6.4 using a 1.3.x version of OC.

Vexatos:
Closes #613.

Dependencies are on `deobfCompile` right now as the mod is useful for testing. I can change it to `provided` in case you prefer that.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,7 +106,12 @@ repositories {
     maven {
         name = "IGW"
         url = "http://maven.k-4u.nl/"
-    }/*
+    }
+    maven {
+        name 'EnderIO'
+        url "http://maven.tterrag.com"
+    }
+    /*
     maven {
         name = "dmodoomsirius"
         url = "http://api.dmodoomsirius.me/"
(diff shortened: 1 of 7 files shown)


fnuecke (approved):
Looking good, seems to work as intended.

PR #2135, opened 2016-11-16
Updated JEI integration.

Vexatos:
Not sure if this should be merged. Added a version dependency or we would probably have the same problems as the highlighting PR had, with people not updating JEI and reporting crashes.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -1,6 +1,6 @@
 minecraft.version=1.10.2
 minecraft.mappings=snapshot_20160720
-forge.version=12.18.1.2014
+forge.version=12.18.2.2122
 
 oc.version=1.6.0
 oc.subversion=dev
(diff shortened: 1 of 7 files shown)


fnuecke:
What exactly were those crashes? Did an interface change? If it's just added/removed methods, a good way to stay compatible with both is to just have the union of the interfaces of different versions and omit the `override` on the implementations, IIRC.

Vexatos:
`override` doesn't even matter, it doesn't exist once you're past the compiler IIRC.

Vexatos:
`override` doesn't even matter, it doesn't exist once you're past the compiler IIRC. `IIngredients`, however, is a completely new class.
Updating JEI is required to launch Minecraft with #2134 merged as EnderIO requires it. Unless I change that to use `provided`, which can, of course, be done.

fnuecke:
It's a compile-time thing, yes. But if some signature changed and Optional.Method and/or Injectable.Interface won't help, sure, let's require the new version. Makes sense anyway IMHO :P

fnuecke (approved):
And I just noticed that fancy new button there...

fnuecke:
Interesting, so that's what that does!

PR #2136, opened 2016-11-16
Allow multiple environment providers to return environments for the same ItemStack.

Vexatos:
Deprecated the old method. I don't know if any mod actually uses it.

This is useful in case of environments accessed through Drivers, as there might be multiple environments for a single block.


payonel:
Is there any testing I can do for this before I merge it?

PR #2139, opened 2016-11-17
Fixed RedLogic redstone input reading.

### LINKED ISSUE #1599: Weird redstone signals?
I'm not sure if I'm just being dumb here or what...
(Thats on OpenComputers-MC1.7.10-1.5.20.38-universal, in ATLauncher Revolution Pack)
Setup: 
PC is standard creative-spawned (/oc_sc) PC with a tier 1 redstone card added
Clearly notes 10 on left: 

Using my test program which toggles the power on side 5 between 0 and 15 when I click x and outputs the redstone_changed events, measures the actual value on getInput on their sides and sets a timer to measure it again 2s later:

So it claims both sides change to 15, and then when turning off back to 10 for side 4 and 0 for side 5...
Waila and redstone textures say side 4 stays 10 though?

(Test program: https://gist.github.com/Inari-Whitebear/d00498b6767b1f9fa3c1 )

### LINKED ISSUE #2137: redstone.setOutput() propagates signal to the opposite side of a block
redstone.setOutput sets also input signal to the opposite side. Happens only with RedLogic installed.

Steps to reproduce:
1. set up a computer with T1 redstone card
2. call lua interpreter
3 lua> component.redstone.setOutput(sides.back, 15)
4. add redstone dust to the front and to the back - there will be 15 level at the back and 0 at the front, so far so good.
5. lua> =component.redstone.getInput(sides.front) 
 will print 15

Tested with 
OpenComputers-MC1.7.10-1.5.20.38-universal.jar 
redlogic-59.1.11.jar

This bug makes it impossible in some cases to use microcontroller as a redstone signal processor, which as I understand is one of the main reasons for its existence.

Vexatos:
Should fix #2137. Should also close #1599 as that appears to be the same issue.

If I'm not mistaken, `pos` in `computeInput` is the position of the receiver, i.e. the computer case, so the input actually needs to be offset. Right now, the case is basically reading its own output from the opposite side whenever redlogic was installed.

### DIFF
--- a/src/main/scala/li/cil/oc/integration/redlogic/ModRedLogic.scala
+++ b/src/main/scala/li/cil/oc/integration/redlogic/ModRedLogic.scala
@@ -19,7 +19,7 @@ object ModRedLogic extends ModProxy with RedstoneProvider {
   }
 
   override def computeInput(pos: BlockPosition, side: ForgeDirection): Int = {
-    pos.world.get.getTileEntity(pos) match {
+    pos.world.get.getTileEntity(pos.offset(side)) match {
       case emitter: IRedstoneEmitter =>
         var strength = 0
         for (i <- -1 to 5) {


payonel:
tested, repro'd the bug and confirmed fixed, thanks @Vexatos

PR #2175, opened 2016-12-10
Rewrote InventoryUtils to use IItemHandlers.

### LINKED ISSUE #2152: OC is not using the IItemHandler capability
Already preventing it from working with some mods, inventory wise.
Probably only gonna get worse :c

Vexatos:
Checking `isUseableByPlayer` is unfortunately not possible anymore due to this.

Also changed `inventoryFor` accordingly because it makes much more sense.
`inventoryFor` now also checks for this capability, making `ItemInventoryControl` a lot more useful.

OC's own inventories are being wrapped into `IItemHandler`s when necessary.

**WARNING:** This is a big rewrite and there are most likely bugs just about everywhere. Please test this as much as possible, especially transposers, inventory controller upgrades and the robot and drone `suck`, `drop` etc., anything interacting with inventories in some way.

Should close #2152 when merged.

### DIFF
--- a/src/main/scala/li/cil/oc/server/agent/Inventory.scala
+++ b/src/main/scala/li/cil/oc/server/agent/Inventory.scala
@@ -10,11 +10,8 @@ import net.minecraft.item.Item
 import net.minecraft.item.ItemStack
 import net.minecraft.nbt.NBTTagCompound
 import net.minecraft.nbt.NBTTagList
-import net.minecraftforge.items.IItemHandlerModifiable
-import net.minecraftforge.items.wrapper.PlayerInvWrapper
 
 class Inventory(val agent: internal.Agent) extends InventoryPlayer(null) {
-  protected implicit def toItemHandler(inv: Inventory):IItemHandlerModifiable = new PlayerInvWrapper(inv)
 
   def selectedItemStack = agent.mainInventory.getStackInSlot(agent.selectedSlot)
 
@@ -57,7 +54,7 @@ class Inventory(val agent: internal.Agent) extends InventoryPlayer(null) {
 
   override def addItemStackToInventory(stack: ItemStack) = {
     val slots = this.indices.drop(agent.selectedSlot) ++ this.indices.take(agent.selectedSlot)
-    InventoryUtils.insertIntoInventory(stack, this, slots = Option(slots))
+    InventoryUtils.insertIntoInventory(stack, InventoryUtils.asItemHandler(this), slots = Option(slots))
   }
 
   override def canHeldItemHarvest(block: Block): Boolean = block.getMaterial.isToolNotRequired || (getCurrentItem != null && getCurrentItem.canHarvestBlock(block))
--- a/src/main/scala/li/cil/oc/server/agent/Inventory.scala
+++ b/src/main/scala/li/cil/oc/server/agent/Inventory.scala
@@ -1,6 +1,7 @@
 package li.cil.oc.server.agent
 
 import li.cil.oc.api.internal
+import li.cil.oc.util.ExtendedInventory._
 import li.cil.oc.util.InventoryUtils
 import net.minecraft.block.Block
 import net.minecraft.entity.player.EntityPlayer
@@ -9,9 +10,12 @@ import net.minecraft.item.Item
 import net.minecraft.item.ItemStack
 import net.minecraft.nbt.NBTTagCompound
 import net.minecraft.nbt.NBTTagList
-import li.cil.oc.util.ExtendedInventory._
+import net.minecraftforge.items.IItemHandlerModifiable
+import net.minecraftforge.items.wrapper.PlayerInvWrapper
 
 class Inventory(val agent: internal.Agent) extends InventoryPlayer(null) {
+  protected implicit def toItemHandler(inv: Inventory):IItemHandlerModifiable = new PlayerInvWrapper(inv)
+
   def selectedItemStack = agent.mainInventory.getStackInSlot(agent.selectedSlot)
 
   def inventorySlots = (agent.selectedSlot until getSizeInventory) ++ (0 until agent.selectedSlot)
--- a/src/main/scala/li/cil/oc/server/component/traits/ItemInventoryControl.scala
+++ b/src/main/scala/li/cil/oc/server/component/traits/ItemInventoryControl.scala
@@ -16,12 +16,12 @@ trait ItemInventoryControl extends InventoryAware {
     withItemInventory(args.checkSlot(inventory, 0), itemInventory => result(itemInventory.getSlots))
   }
 
-  @Callback(doc = "function(inventorySlot:number, slot:number[, count:number=64]):number -- Drops an item into the specified slot in the item inventory.")
+  @Callback(doc = "function(inventorySlot:number, slot:number[, count:number=64]):number -- Drops an item from the selected slot into the specified slot in the item inventory.")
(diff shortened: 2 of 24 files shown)


src/main/scala/li/cil/oc/server/agent/Inventory.scala
```diff
@@ -9,9 +10,12 @@ import net.minecraft.item.Item
 import net.minecraft.item.ItemStack
 import net.minecraft.nbt.NBTTagCompound
 import net.minecraft.nbt.NBTTagList
-import li.cil.oc.util.ExtendedInventory._
+import net.minecraftforge.items.IItemHandlerModifiable
+import net.minecraftforge.items.wrapper.PlayerInvWrapper
 
 class Inventory(val agent: internal.Agent) extends InventoryPlayer(null) {
+  protected implicit def toItemHandler(inv: Inventory):IItemHandlerModifiable = new PlayerInvWrapper(inv)
```
fnuecke: Change that to `new InvWrapper(inv)` instead to avoid NPEs in the player wrapper. Yes, agents are sorta-kinda fake players, but our agent inventory is just a wrapper around their regular inventory anyway, so this fine.

src/main/scala/li/cil/oc/util/InventoryUtils.scala
```diff
@@ -61,40 +87,31 @@ object InventoryUtils {
    * The number of items inserted can be limited, to avoid unnecessary
    * changes to the inventory the stack may come from, for example.
    */
-  def insertIntoInventorySlot(stack: ItemStack, inventory: IInventory, side: Option[EnumFacing], slot: Int, limit: Int = 64, simulate: Boolean = false) =
+  def insertIntoInventorySlot(stack: ItemStack, inventory: IItemHandler, slot: Int, limit: Int = 64, simulate: Boolean = false): Boolean =
     (stack != null && limit > 0) && {
```
fnuecke: Add an `&& stack.stackSize > 0` here, otherwise we'll insert empty stacks into item handlers (e.g. vanilla chests). In particular when called via `insertIntoInventory` due to the loop in there (filling all the slots with empty stacks after the input has been consumed).

This was caught by the same check below before.

src/main/scala/li/cil/oc/util/InventoryUtils.scala
```diff
@@ -120,32 +137,31 @@ object InventoryUtils {
    * also be achieved by a check in the consumer, but it saves some unnecessary
    * code repetition this way.
    */
-  def extractFromInventorySlot(consumer: (ItemStack) => Unit, inventory: IInventory, side: EnumFacing, slot: Int, limit: Int = 64) = {
+  def extractFromInventorySlot(consumer: (ItemStack) => Unit, inventory: IItemHandler, slot: Int, limit: Int = 64): Boolean = {
     val stack = inventory.getStackInSlot(slot)
     (stack != null && limit > 0) && {
```
fnuecke: Throw in that `&& stack.stackSize > 0` here, too, seeing as that check existed here before, too. At least it'll save us some unnecessary work.

src/main/scala/li/cil/oc/util/InventoryUtils.scala
```diff
@@ -160,14 +176,11 @@ object InventoryUtils {
    * item stack will be adjusted to reflect the number items inserted, by
    * having its size decremented accordingly.
    */
-  def insertIntoInventory(stack: ItemStack, inventory: IInventory, side: Option[EnumFacing] = None, limit: Int = 64, simulate: Boolean = false, slots: Option[Iterable[Int]] = None) =
+  def insertIntoInventory(stack: ItemStack, inventory: IItemHandler, limit: Int = 64, simulate: Boolean = false, slots: Option[Iterable[Int]] = None): Boolean =
     (stack != null && limit > 0) && {
```
fnuecke: For consistency, add an `&& stack.stackSize > 0` here, too, easy early exit.

src/main/scala/li/cil/oc/util/InventoryUtils.scala
```diff
@@ -61,40 +87,31 @@ object InventoryUtils {
...
-            existing.stackSize += amount
-            inventory.markDirty()
-            true
-          }
+      val amount = math.min(stack.stackSize, limit)
+      if (simulate) {
+        val toInsert = stack.copy()
+        toInsert.stackSize = amount
+        inventory.insertItem(slot, toInsert, simulate) match {
+          case remaining: ItemStack => remaining.stackSize < stack.stackSize
```
fnuecke: Shouldn't this be `remaining.stackSize < amount`? If `limit < stack.stackSize` nothing could have happened, and this would still be true, no?

src/main/scala/li/cil/oc/util/InventoryUtils.scala
```diff
@@ -61,40 +87,31 @@ object InventoryUtils {
...
-          if (simulate) amount > 0
-          else {
-            inventory.setInventorySlotContents(slot, inserted)
-            true
-          }
+      } else {
+        val toInsert = stack.splitStack(amount)
+        inventory.insertItem(slot, toInsert, simulate) match {
+          case remaining: ItemStack =>
+            val result = remaining.stackSize < stack.stackSize
```
fnuecke: Ditto `remaining.stackSize < amount` I think.

fnuecke (changes requested):

xarses:
We should be able to test this against the drawer controller https://github.com/jaquadro/StorageDrawers/wiki/Drawer-Controller as they noted that this was the preferred interface

asiekierka:
Also, Charset pipes prefer IItemHandlers, and, in 1.11+, barrels support them exclusively.

fnuecke:
> 1.11

Implying someone will port OC. Nice one.

asiekierka:
>Implying

src/main/scala/li/cil/oc/server/component/traits/ItemInventoryControl.scala
```diff
@@ -7,21 +7,21 @@ import li.cil.oc.api.machine.Context
...
+    withItemInventory(args.checkSlot(inventory, 0), itemInventory => result(itemInventory.getSlots))
   }
 
   @Callback(doc = "function(inventorySlot:number, slot:number[, count:number=64]):number -- Drops an item into the specified slot in the item inventory.")
   def dropIntoItemInventory(context: Context, args: Arguments): Array[AnyRef] = {
     withItemInventory(args.checkSlot(inventory, 0), itemInventory => {
       val slot = args.checkSlot(itemInventory, 1)
       val count = args.optItemCount(2)
-      result(InventoryUtils.extractAnyFromInventory(InventoryUtils.insertIntoInventorySlot(_, itemInventory, None, slot), inventory, null, count))
+      result(InventoryUtils.extractAnyFromInventory(InventoryUtils.insertIntoInventorySlot(_, itemInventory, slot), inventory, null, count))
```
fnuecke: While we're fixing this hot garbage, this should IMHO also be
```scala
result(InventoryUtils.extractFromInventorySlot(InventoryUtils.insertIntoInventorySlot(_, itemInventory, slot), inventory, null, selectedSlot, count))
```
i.e. only try to insert the *selected* stack in an agent's inventory. Not try each and every item in its inventory until one fits into the item inventory.

fnuecke (changes requested):
That thing, and I think this is all done, seems to work find with the test cases I could come up with with vanilla chests and furnaces.

PR #2196, opened 2016-12-21
use string and not unicde in text streams

payonel:
text streams erronously assumed unicode mode. Rather, they can use string utils exclusively, and pass the read/write mode to the buffer lib, leaving string vs unicode selection to the latter


payonel:
fixed in the 1.6.1 build from this commit: https://github.com/MightyPirates/OpenComputers/commit/c86ce4d28ccd54f8d6dbfc51c7fcc0573d00b11f

PR #2251, opened 2017-02-02
1.6.1 update: memory, performance, devfs, relative links

payonel:
- reduced memory cost to boot to shell!
- READONLY mount option added: `mount -r ...`
- os.execute (and shell.execute) now correctly spawn child shells with isolated shell environment (i.e. shell env vars such as PWD do not proprogate to the parent shell) Note that require("sh").execute(env, command, ...) is an option for direct command execution if dirtying the current shell is desired. Note that /bin/source uses sh.execute for this purpose
- /dev/components added with smart adapters and links
- supported devfs components at this time:
- computer, internet, filesystem, screen, gpu, modem, eeprom
- components can be labeled with /bin/label
- /dev/components/by-label respects component labels
- /etc/udev/rules.d/ supports boot labels
-- component labeling and udev rules should be considered experiemental
- links can now be relative
- devfs significantly simplified, api reduced to single create method
- various code cleanup
- /bin/cp and /bin/mv hardened and now share common code
- multiple fs mount point now share appropriate meta data (e.g. links)
- hardened fs library
- fs now detects and appropriately handles link cycles (a->b->a)
- grep fix: show multiple matches with -o
- install .prop added support for "ignore" key, for floppies to NOT be detected as an install source candidate. If you have a floppy and you don't want install to list it for install options, add {ignore=true} in a .prop file
- less and more call cat when there is no tty (e.g. piped via |)
- ln can create relative links
- pwd -P option added to show real path
- a messy term gpu resize issue fixed where the gpu can change resolution before the terminal is able to handle the resize event
- smart hinting (where single hint results can extend to the next set) has been improved and put in shared code: now shell has it not just lua shell
- text stream reader/writer now supports unicode as well as binary mode, e.g. writing binary data to /dev/eeprom now fixed


fnuecke:
Awesome! Thanks as per usual for taking such wonderful care of OpenOS :)

PR #2256, opened 2017-02-03
Allow debug cards to send messages to one another and text to a player's clipboard. Also added a command for sending messages to a debug card.

### LINKED ISSUE #2253: Admin command to broadcast modem messages
I'm in the process of creating a challenge map, and I'm using OC to display elapsed time and more. Currently I start the timer by placing a redstone block in a predetermined location. This means the quests are locked to a single map. To help alleviate this, it would be great if there was an admin command to broadcast modem messages.

The current use case is where I execute the command using Better Questing when a quest is completed. Upon completion it runs the command, which starts a timer.

Vexatos:
Closes #2253. I think.


fnuecke:
Well, it didn't crash, not even in MP, so yay :D

PR #2260, opened 2017-02-04
Re-added ProjectRed integration.

### LINKED ISSUE #2259: Readd ProjectRED support to 1.10.2
ProjectRED has updated to 1.10.2 now, so could the integration be readded?

Just FYI.

Vexatos:
Closes #2259.


src/main/scala/li/cil/oc/integration/projectred/EventHandlerProjectRed.scala
```diff
@@ -6,10 +6,11 @@ import net.minecraft.item.ItemStack
 
 object EventHandlerProjectRed {
   def useWrench(player: EntityPlayer, x: Int, y: Int, z: Int, changeDurability: Boolean): Boolean = {
```
brisingraerowing: Shouldn't the x, y, and z parameters be replaced with a BlockPos parameter? Other integrations do that.

Issue #2294, opened 2017-03-01
Allow the ability to get the maximum parts of a packet that a modem can send.


skyem123:
This should be simple...
Modems already have `maxPacketSize()`, and as the number of parts is configurable, why not have a `maxPacketParts()` too?

MajorGeneralRelativity:
That would be nice to have.

skyem123:
This is probably really simple to add, and I'd do it myself if setting up a dev environment for OC wasn't so painful.

brisingraerowing:
I could probably do it. I usually just edit in Notepad++ or Atom, and then build with Gradle. It's slow, but it works (and I really don't want to install Eclipse or Netbeans, as I don't really do any Java development anyways).

brisingraerowing:
I got it written (didn't take long), I just have to test it now. Will report back.

brisingraerowing:
It works perfectly. @fnuecke, would you accept a pull request to add this method?

payonel:
@BrisingrAerowing you don't need to ask permission to make a PR, you just submit it and see where that goes.

skyem123:
@BrisingrAerowing, odd, last time I tried to get a dev environment working, it didn't even build. Though that was some time ago...

skyem123:
Anyway, please do submit a PR for this! It would be good to have this! :D

brisingraerowing:
I get right on it.

brisingraerowing:
@skyem123 I build everything from the command line, and edit with a basic editor. Avoids having to work around bugs and incompatibilities in various IDEs.

PR #2297, opened 2017-03-06
Added maxPacketParts method to the Network Card

brisingraerowing:
Ref #2294 

This adds a method to get the maximum number of parts a packet can have. Simple as that.


payonel:
please resubmit, target master-MC1.7.10

PR #2298, opened 2017-03-07
Added maxPacketParts method to the Network Card

brisingraerowing:
Ref #2294

### DIFF
--- a/src/main/java/li/cil/oc/api/driver/DeviceInfo.java
+++ b/src/main/java/li/cil/oc/api/driver/DeviceInfo.java
@@ -69,7 +69,8 @@ public interface DeviceInfo {
         public static final String Size = "size"; // actual size of the device, e.g. actual usable space on a disk
         public static final String Clock = "clock"; // bus clock (in Hz) of the device, e.g. call speed(s) of a component
         public static final String Width = "width"; // address width of the device, in the broadest sense
-
+		public static final String MaxParts = "maxparts"; // Maximum number of parts for, e.g. a network packet.
+		
         private DeviceAttribute() {
         }
     }
--- a/src/main/java/li/cil/oc/api/driver/DeviceInfo.java
+++ b/src/main/java/li/cil/oc/api/driver/DeviceInfo.java
@@ -69,7 +69,8 @@ public interface DeviceInfo {
         public static final String Size = "size"; // actual size of the device, e.g. actual usable space on a disk
         public static final String Clock = "clock"; // bus clock (in Hz) of the device, e.g. call speed(s) of a component
         public static final String Width = "width"; // address width of the device, in the broadest sense
-
+        public static final String MaxParts = "maxparts"; // Maximum number of parts for, e.g. a network packet.
+        
         private DeviceAttribute() {
         }
     }
--- a/src/main/scala/li/cil/oc/server/component/NetworkCard.scala
+++ b/src/main/scala/li/cil/oc/server/component/NetworkCard.scala
@@ -48,7 +48,8 @@ class NetworkCard(val host: EnvironmentHost) extends prefab.ManagedEnvironment w
     DeviceAttribute.Description -> "Ethernet controller",
     DeviceAttribute.Vendor -> Constants.DeviceInfo.DefaultVendor,
     DeviceAttribute.Product -> "42i520 (MPN-01)",
-    DeviceAttribute.Capacity -> Settings.get.maxNetworkPacketSize.toString
+    DeviceAttribute.Capacity -> Settings.get.maxNetworkPacketSize.toString,
+    DeviceAttribute.MaxParts -> Settings.get.maxNetworkPacketParts.toString
   )
 
   override def getDeviceInfo: util.Map[String, String] = deviceInfo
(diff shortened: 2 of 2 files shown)


fnuecke:
I'd prefer this to go into the [device info](https://github.com/BrisingrAerowing/OpenComputers/blob/9d0b7438bcac338b799b281141e493cd177506bd/src/main/scala/li/cil/oc/server/component/NetworkCard.scala#L51) (hence the "todo remove" comment on the other, similar method). As for the attribute name, I suppose "width" could do? But feel free to use something more specific.

skyem123:
It's worth noting that I feel like it should be on the network card component because it's like the MTU of real networks?

skyem123:
It's worth noting that I feel like it should be on the network card component because it's like the MTU of real networks?

fnuecke:
Not sure what your point is? The device info is still part of the network card component.

skyem123:
I'm sorry... I don't really know what you're talking about with regards to "device info", as there doesn't seem to be anything on http://ocdoc.cil.li?

Vexatos:
[This](https://github.com/BrisingrAerowing/OpenComputers/blob/9d0b7438bcac338b799b281141e493cd177506bd/src/main/scala/li/cil/oc/server/component/NetworkCard.scala#L46-L52) is the device info.

brisingraerowing:
@fnuecke 

So how would that work? Would I add a new device attribute and set it to the value, similar to what is there?

What should that attribute be called?

Vexatos:
Yes, exactly.

src/main/scala/li/cil/oc/server/component/NetworkCard.scala
```diff
@@ -48,7 +48,8 @@ class NetworkCard(val host: EnvironmentHost) extends prefab.ManagedEnvironment w
     DeviceAttribute.Description -> "Ethernet controller",
     DeviceAttribute.Vendor -> Constants.DeviceInfo.DefaultVendor,
     DeviceAttribute.Product -> "42i520 (MPN-01)",
-    DeviceAttribute.Capacity -> Settings.get.maxNetworkPacketSize.toString
+    DeviceAttribute.Capacity -> Settings.get.maxNetworkPacketSize.toString,
+	DeviceAttribute.MaxParts -> Settings.get.maxNetworkPacketParts.toString
```
Vexatos: Fix the indentation here.

src/main/java/li/cil/oc/api/driver/DeviceInfo.java
```diff
@@ -69,7 +69,8 @@
         public static final String Size = "size"; // actual size of the device, e.g. actual usable space on a disk
         public static final String Clock = "clock"; // bus clock (in Hz) of the device, e.g. call speed(s) of a component
         public static final String Width = "width"; // address width of the device, in the broadest sense
-
+		public static final String MaxParts = "maxparts"; // Maximum number of parts for, e.g. a network packet.
```
Vexatos: Fix the indentation here. I don't like the attribute name, but I can't think of anything other than @fnuecke's "width" either.

src/main/java/li/cil/oc/api/driver/DeviceInfo.java
```diff
@@ -69,7 +69,8 @@
...
+
```
Vexatos: Another formatting error. Remove the spaces.

Vexatos (changes requested):

src/main/java/li/cil/oc/api/driver/DeviceInfo.java
brisingraerowing: Already done. Notepad++ Is rather stupid when it comes to Tabs vs Spaces.

src/main/scala/li/cil/oc/server/component/NetworkCard.scala
```diff
@@ -48,8 +48,9 @@ class NetworkCard(val host: EnvironmentHost) extends prefab.ManagedEnvironment w
     DeviceAttribute.Description -> "Ethernet controller",
     DeviceAttribute.Vendor -> Constants.DeviceInfo.DefaultVendor,
     DeviceAttribute.Product -> "42i520 (MPN-01)",
-    DeviceAttribute.Capacity -> Settings.get.maxNetworkPacketSize.toString
-  )
+    DeviceAttribute.Capacity -> Settings.get.maxNetworkPacketSize.toString,
+    DeviceAttribute.MaxParts -> Settings.get.maxNetworkPacketParts.toString
+    )
```
Vexatos, line 53: Still not formatted properly.

src/main/java/li/cil/oc/api/driver/DeviceInfo.java
```diff
@@ -69,7 +69,7 @@
         public static final String Size = "size"; // actual size of the device, e.g. actual usable space on a disk
         public static final String Clock = "clock"; // bus clock (in Hz) of the device, e.g. call speed(s) of a component
         public static final String Width = "width"; // address width of the device, in the broadest sense
-
+        public static final String MaxParts = "maxparts"; // Maximum number of parts for, e.g. a network packet.
         private DeviceAttribute() {
```
Vexatos, line 73: And an empty line here, please.

Vexatos (changes requested):

PR #2363, opened 2017-04-28
Add Manpages

JakobDev:
Just add a few manpages


payonel:
some of the wording needs some correction, but i appreciate the work you started here. I'll merge and fix

PR #2369, opened 2017-05-06
Added support for forge energy.

### LINKED ISSUE #2314: Not detecting certain power/energy mods (1.10)
After installing some mods with OpenComputers, this message appeared in chat:
image
The problem is, I *have* mods that generate/use power. The mods are Ender I/O, Extra Utilities 2, and Thermal Expansion.

Vexatos:
Still needs testing.

A big problem is that now, an energy-providing mod will _always_ be available. Not sure how to detect this properly...

Will probably fix #2314, or at least most of it.

### DIFF
--- a/src/main/scala/li/cil/oc/integration/Mods.scala
+++ b/src/main/scala/li/cil/oc/integration/Mods.scala
@@ -45,6 +45,7 @@ object Mods {
   val ExtraCells = new SimpleMod(IDs.ExtraCells, version = "@[2.2.73,)")
   val Factorization = new SimpleMod(IDs.Factorization, providesPower = true)
   val Forestry = new SimpleMod(IDs.Forestry, version = "@[5.2,)")
+  val Forge = new SimpleMod(IDs.Forge, providesPower = true)
   val ForgeMultipart = new SimpleMod(IDs.ForgeMultipart)
   val Galacticraft = new SimpleMod(IDs.Galacticraft, providesPower = true)
   val GregTech = new ClassBasedMod(IDs.GregTech, "gregtech.api.GregTech_API")()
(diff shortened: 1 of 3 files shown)


fnuecke:
Ah, great, was planning to look at that coming week or so, now I won't have to \o/

While you're at it, mind also slapping that on the tile entities?

As for detection, I'd say let's throw out the auto-detection. Just have people change the setting for their pack.

Vexatos:
>mind also slapping that on the tile entities?

Not quite sure what you mean... Anything extending `PowerAcceptor` should support forge energy now.

fnuecke:
A copy of [this](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/scala/li/cil/oc/common/tileentity/traits/power/RedstoneFlux.scala), just for the Forge system. Or is the 1.11 branch upstream? I forgot which way around we're currently merging :x Then that would be better to do directly in there and merge back ofc.

Vexatos:
This PR doesn't work like that, it works more like the Tesla integration. Forge energy is not an interface on the TE, it is a capability, so there is no need for it to be polluting the TE directly.

The implementation is the same on both 1.10 and 1.11, and we are still merging upwards.

thisischrys:
Can we get a build for this?

LizzyTrickster:
@thisischrys Check http://ci.cil.li under the "dev" builds

fnuecke:
Latest builds on Curse also have this.

thisischrys:
Odd, still having this problem with the latest curse version from 1.10.2.
Gonna try the dev.

Reggle:
Yes and no. EnderIO is detected as a power mod and computers require power, but the Electronics Assembler does not. I can make an entire robot without using power.
Version OpenComputers-MC1.10.2-1.6.2.102-dev

PR #2378, opened 2017-05-13
fix insertIntoInventorySlot not updating input stack in simulate mode

yorickvP:
The use from `hasRoomForItemStack` expects `insertIntoInventory` to update the input stack, even when `simulate = true`. Whether this is a valid use-case, or if the check should be rewritten, is for you to decide, but this fixes that usecase and fixes trading.


Vexatos:
Closed via #2379.

PR #2421, opened 2017-06-09
openos 1.6.4

### LINKED ISSUE #2366: make event.pull compatable with coroutines
Apparently there are users that have had a very bad experience trying to use coroutines with openos.

My guess is that they are trying to call `event.pull` from a coroutine state, which (currently) will cause that coroutine to lock on a direct call to `computer.pullSignal`.

The solution will need to intercept event.pull calls to dispatch computer.pullSignal.

payonel:
the "thread" update

**Threads**
OpenOS 1.6.4 brings the new thread library api. Documentation in our ocdoc wiki soon to come. An openos thread is an autonomous non-blocking detachable child process
 * Autonomous: Threads asynchronously begin execution immediately after creation without needing to call resume. The thread proc may call coroutine.yield, but will resume on its own
 * Non-Blocking: Threads can call computer.pullSignal (or any higher level wrapper such as event.pull, io.pull, etc) without blocking the main kernel process nor any other thread
 * Detachable: By default, threads are scoped to the process in which they are created, i.e. their parent process. Any thread will block the parent process from closing unless:
  A. The thread detaches from the parent process. In which case it does not block any process and runs independently, e.g. `t:detach()`
  or
  B. The parent process throws an exception or calls os.exit in which case all attached threads are killed, e.g. `os.exit()`
  or
  C. The thread is manually suspended, e.g. `t:suspend()`

**Command Redirection**
The other major improvement in this update is highly improved shell parsing for command substitution and io redirection. Some highlights include
 * Can place before the command now, e.g. `2>/dev/null ./run_my_scripts.lua`
 * Can properly use globbing or env vars as redirect targets, e.g. `./run_my_scripts.lua >$my_log_file`
 * Fixed various bugs related to redirect and argument evaluation

**Memory**
50k free! As I love to do, this update reduces memory allocation needed to reach the shell prompt. The majority of the recent memory improvements are not just delaying allocation, but actual code cleanup and optimizations. Many of the changes are minor but they are numerous. With 1 stick of tier 1 RAM, openos reaches shell with 50k bytes free.

Changelog:
Bump version from 1.6.3 to 1.6.4
Cause grep to yield when taking too long
Remove additional gpu.bind calls during boot, reduces the number of screen resize calls
Move SHELL env var creation to /boot/94_shell.lua - This is an important user workflow fix to allow users to specify a custom SHELL without openos ever needing to load /bin/sh and its libraries
Fix event dispatch to not double call event timers in some scenarios
new /lib/thread.lua
fixed /lib/process from hiding exceptions in some scenarios
significant memory optimizations and code cleanup for /bin/sh
refactoring of command redirection, variable evaluation, glob expansion, and argument lists in /bin/sh
fixed term.setCursorBlink so that it properly waits for inf time for the next event when not blinking, rather than pulling every .5 seconds
fixed /bin/sh shell from losing exit_code when using || and && with multiple commands
fix /bin/lua error message when using os.exit
cherry picking serialization fix from 62471f7d320758bebb280666ed98388ea61cb4c8
fixed shebang support:  respect _ENV and removed hack in loadfile

Closes #2366

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/grep.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/grep.lua
@@ -9,6 +9,7 @@ https://raw.githubusercontent.com/OpenPrograms/Wobbo-Programs/master/grep/grep.l
 local fs = require("filesystem")
 local shell = require("shell")
 local tty = require("tty")
+local computer = require("computer")
 
 -- Process the command line arguments
 
@@ -298,7 +299,14 @@ local function test(m,p)
     m.close = true
   end
 end
+
+local uptime = computer.uptime
+local last_sleep = uptime()
 for meta,status in readLines() do
+  if uptime() - last_sleep > 1 then
+    os.sleep(0)
+    last_sleep = uptime()
+  end
   if not meta then
     if type(status) == 'table' then if flush then
       flush(status) end -- this was the last object, closing out
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/lua.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/lua.lua
@@ -22,6 +22,6 @@ end
 
 buffer, reason = pcall(script, table.unpack(args, 2))
 if not buffer then
-  io.stderr:write(reason, "\n")
+  io.stderr:write(type(reason) == "table" and reason.reason or tostring(reason), "\n")
   os.exit(false)
 end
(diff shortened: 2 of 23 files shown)


magik6k:
I feel like what is in `/opt/core` should be in `/lib/core`. Other than that looks great!

payonel:
Closes #2366

payonel:
> /opt/core should be in /lib/core

I debated /opt/core to /lib/core -- the main reason is that /lib things are meant for require() and I didn't intend /opt things to be required. Any thoughts on this @magik6k ?

magik6k:
In most Linux distros(not sure about other posix-es), programs have subdirectories in `/lib` and `/usr/lib` for all the stuff they need to function(modules, scripts, internal programs, libraries, anything that doesn't belong to /usr/share, but is needed fot a program/system to function. etc). `/opt` is meant for whole program bundles(much like Windows `program files`)

payonel:
> for all the stuff they need to function

Then /lib/core would be perfect. I'll refactor file locations in the next update. Thanks

PR #2428, opened 2017-06-18
Thread and Pipe patches

payonel:
openos 1.6.5 update fixes and optimizations, one api removed

process: process.running has been deprecated for over 2 years and is now obsolete, please use process.info
01_process: cleaner coroutine_handler intercept code
event: removed unnecessary local function and added an optional handlers param to register to simplify thread code
io: update popen to use new /lib/pipe.lua library
pipe: new /lib/pipe replace /lib/pipes
* superior coroutine stack code, can reparent stacks
* handles all pipe work, sh calls /lib/pipe now for pipes

sh: now uses /lib/pipe to build pipe chains, calls shell.run to run a pipe chain.
thread: greatly improved reliability of embedded and detached threads
boot: removed dead code


Vexatos:
Having both `shell.run` and `shell.execute` might be a bit confusing.

payonel:
> shell.run ... might be a bit confusing.

Completely agree, and, this method really has nothing to do with the shell specifically. It fits much better in /lib/process

PR #2432, opened 2017-06-21
Optimized filesystem.name() / Fix filesystem.copy()

SDPhantom:
There is no need to split and run path parsing. Just grab the substring at the end.
Also added argument checking.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/filesystem.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/filesystem.lua
@@ -283,7 +283,7 @@ end
 
 function filesystem.name(path)
   checkArg(1, path, "string")
-  return path:match("[^\\/]+$")
+  return path:match("([^\\/]+)[\\/]*$")
 end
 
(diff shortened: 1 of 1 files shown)


SDPhantom:
Added in a fix for the error return values when filesystem.copy() is unable to open either input or output file.

payonel:
The regex doesn't work when the path is ".", the expected name is nil
other expected results:
./ is nil
../ is nil
.. is nil
a/d/.. is a
a/. is a
./. is nil
./../a./. is a

the file.copy failure for input and output is a good fix.

SDPhantom:
> filesystem.name(path: string): string
Returns the file name component of a path to a file, i.e. everything after the last slash in the canonical form of the specified path.

Aren't those expectations the reverse of the documented effect?

SDPhantom:
I see I need to work a little more on it, though I'll have to do some testing on the way segments() walks through the path. I have ideas on improving it using gsub().

payonel:
for crying out loud
this was the wrong branch

PR #2433, opened 2017-06-22
Optimized wrapped functions returned by coroutine.wrap()

SDPhantom:
Streamlined the process by stack manipulation rather than storing, indexing, and dumping dynamic tables.
assert() returns everything it receives if it doesn't raise an error.


payonel:
thanks, but we target master-1.7.10, not 1.10 -- i'll merge after you change the target branch

payonel:
i was making another change to 01_process anyways, and including this optimization. thank you very much

PR #2436, opened 2017-06-23
Optimized sleep.lua interval parsing

SDPhantom:
Modified regex to use captures, consolidated invalid interval checks.
Note: **tonumber(_nil_)** safely returns **_nil_**

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/sleep.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/sleep.lua
@@ -39,7 +39,7 @@ end
 local total_time = 0
 
 for _,v in ipairs(args) do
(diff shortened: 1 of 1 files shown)


payonel:
this is better, thanks
Seems if we're trusting tonumber to validate anyways, we could allow for sleep intervals such as `.1` -- which currently isn't allowed
How would you feel about ` local interval, time_type = v:match('^([%d%.]+)([smhd]?)$')` ?

SDPhantom:
I had considered that as an idea, but didn't actually make that change. Initially, I was going to have **match()** take care of invalid inputs like ".1." or "0..1", but with the consolidated **if...then** block, this became a moot point. If Lua can see a valid number out of the input, I don't see a problem with making the switch.

PR #2437, opened 2017-06-24
Added 'quiet' mode to serialization.serialize()

SDPhantom:
'quiet' suppresses errors and writes nil plus a commented reason to any unsupported values.
This change also implements a fix allowing limitless serialization with 'pretty' off.
This is the new default action. Setting 'pretty' to true preserves the default limit to 10 lines.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
@@ -70,7 +70,7 @@ local function loadConfig()
       if f then
         local serialization = require("serialization")
         for k, v in pairs(env) do
-          f:write(k.."="..tostring(serialization.serialize(v, math.huge)).."\n")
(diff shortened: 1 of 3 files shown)


payonel:
The serialization library does need some improvements (such as using tables for buffering instead of string concats) and could use a lot of cleanup. And this option to not abort of failures or recursion would be nice. However, I'm not sure this is the direction I want to take the library at this time. I do think there should be some option to the user to allow ignoring failures such as this. But I don't want to make changes to the parameters at this time, nor the long standing default behavior of the library. I do appreciate the work you are putting into this, please keep it up and feel free to chat with me on our irc channel on esper in the #oc channel

PS: There was a typo in the code `eseif` instead of `elseif`

SDPhantom:
I had a commit where I went back and corrected it, it appears GitHub grabbed the wrong one when making this request.

PR #2439, opened 2017-06-25
Better scrollwheel handling in edit.lua

SDPhantom:
Now scrolls the screen up/down instead of jumping the cursor a bunch of lines.
This tries to preserve cursor position if it's still on screen.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
@@ -630,7 +630,7 @@ local function onScroll(direction)
     scrollY = sy
 
     term.setCursorBlink(false)
-    gpu.copy(x, math.max(y, y + dy), w, h - math.abs(dy), 0, -dy)
+    if math.abs(dy) < h then gpu.copy(x, math.max(y, y + dy), w, h - math.abs(dy), 0, -dy) end
     for py = 1, math.min(-dy, h) do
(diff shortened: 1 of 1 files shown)


payonel:
this looks like a good approach. i'd like to test/play with it first. i'm on vacation without a mouse -- will review next week

SDPhantom:
There is an issue that I foresee with it and that has to do with X-scrolling snapping back and forth as you're scrolling lines. It's not easy to counter unless we overhaul how scrolling works and allow direct manipulation of scrolling instead of having to jump the cursor around in order to proc a redraw of the screen.

Ideally, I'd like the cursor to keep its position in the buffer even if it's scrolled offscreen. From there, you can click to change the cursor position as usual or the scrolling will snap back to the cursor when you start typing. Much like a real text editor.

SDPhantom:
The newer commit should handle half of the issue I explained earlier. Due to imitations with the term library, I have to have the scroll code drag the cursor with it when it reaches the edge of the screen.

Previously I moved the cursor one line above or below the screen to force a redraw and at that time, the cursor X position was set to 1. Now I have a similar GPU copy method to **setCursor()** handle scrolling and shift the cursor Y position (relative to the buffer) if would have been moved off screen. This may still cause the X scroll to snap to the left if the line the cursor moves to isn't long enough. At the very least, there won't be the rare occasion where it would snap to the left when scrolling occurs and back to the right when the cursor position is attempted to be restored.

SDPhantom:
Oops, didn't mean to hit close :P

payonel:
in an upcoming update, i'm replacing edit. I'll review this improvement in that version

PR #2443, opened 2017-06-26
openos 1.6.6

payonel:
More memory savings, significant boot time savings

**Memory Savings**
The largest library in terms of cost in bytes in 1.6.5 was /lib/filesystem. Most of this library is not needed for boot, and in 1.6.6 the library is half loaded for boot with a delay load of the rest of the library when needed. There are also other minor memory improvements, see **Changes** for details. Free Memory now at ~55k on a Lua 5.2 CPU, and ~61k on a Lua 5.3

**Boot Time**
This update also introduces a newly written /bin/sh.lua that takes advantage of a common stdio system. This work was primary done to reduce memory needed to load the shell. However, during development of a simplified /bin/sh, I found that the /bin/source, /etc/profile, and /home/.shrc feature added to openos 1.6 had a significant boot time cost. I began optimizing /bin/source heavily and found that I could not make sufficient speed improvements due to the nature of how shell commands execute. Thus I feel that some breaking changes were necessary for the overall user experience. /etc/profile has been changed to a lua script now called /etc/profile.lua and /home/.shrc is now /home/.shrc.lua. /bin/source is still available and can still execute shell script files as before, but it is not used during boot. This saves about ~2s on boot time with floppies, ~1s with hard drives.

**Breaking Changes**
/etc/profile and /home/.shrc are now /etc/profile.lua and /home/.shrc.lua
Custom settings users may have added to either will not be sourced during boot. Those changes will have to be migrated in lua code form to these new files. This decision was made with respect to significant boot time savings.

**Changes**
* /bin/sh.lua rewrite that takes advantage of standard io and uses the new /etc/profile.lua shaving 6s off boot time
* /bin/source.lua improve performance by reducing the number of processes created
* rename /etc/profile to /etc/profile.lua and /home/.shrc to /home/.shrc.lua and recode their actions in lua-form
* move motd to start of /etc/profile.lua to get the shell display as soon as possible
* update greetings that referenced /etc/profile to properly refer to /etc/profile.lua now
* update openos version to 1.6.6 to the impact of these changes


Vexatos:
Not sure I like ~/.shrc being gone. It was very convenient compared to a Lua script.

payonel:
@Vexatos rerunning my tests turns out the 6s time savings was a strange outlier. Looks more like 3s saved on floppies, and only 1s saved on hdds

also, not gone, just "renamed" :)

But I know what you're saying

payonel:
@Vexatos `~/.shrc` is back, should appear no different to end users now

if `~/.shrc` exists, `/etc/profile.lua` will `source` it. All previous shell commands (there were only a few, such as a /bin/ls alias `l`) are moved to `/etc/profile.lua` - thus sourcing ~/.shrc on a clean install is essentially a no-op

The time cost in sourcing the empty file is negligible

PR #2449, opened 2017-07-03
vt100 support

payonel:
vt100 support

* huge terminal performance boost
* echo -e option, interpret backslashes
* favor io.write instead of print on low level code as print is slightly more memory expensive
* use vt100 ansi color coding for PS1
* use tty as std io back end stream
* consolidate buffer delay code
* add ctrl+backspace and ctrl+w options in io.read
* delay load more
* term code cleanup
* add buffer:size(), returns size of pending data on the stream
* memory free 59k
* openos 1.6.7


payonel:
Gifs demonstrating the terminal performance boost
It may appear that frames are missing in the openos 1.6.7 gif below, they are not, it's just that much faster

openos-1 6 6
openos-1 6 7

Issue #2451, opened 2017-07-05
UNIX-like LS_COLORS variable


ghost:
As now OpenOS have ANSI escape codes, LS_COLORS variable should look more UNIX-like.
So it would have following format, just a colon-separated list:
```
di=0;36:fi=0:ln=0;33:*.lua=0;32
```
 - `di` used for *directories*
 - `fi` used for *files*
 - `ln` used for *symbolic links*
 - `*.extension` used as before

Colors are specified as a semicolon-separated list of SGR-codes.

payonel:
closed by e1e2e3c80061d2a5e5abb3ca2cb6adb4e183ed39

PR #2453, opened 2017-07-05
OpenOS: Add tree utility

ghost:
I found tree useful so I wrote it for OpenOS.
[Here is screenshots](http://imgur.com/a/okG4q)

It doesn't use recursion so it would not fail with stack overflow. Also I found it faster (and better) than using `ls -R`

```
Usage: tree [OPTION]... [FILE]...
  -a, --all             do not ignore entries starting with .
      --full-time       with -l, print time in full iso format
  -h, --human-readable  with -l, print human readable sizes
      --si              likewise, but use powers of 1000 not 1024
      --level=LEVEL     descend only LEVEL directories deep
      --color=WHEN      WHEN can be
                        auto - colorize output only if writing to a tty,
                        always - always colorize output,
                        never - never colorize output; (default: auto)
  -l                    use a long listing format
  -f                    print the full path prefix for each file
  -i                    do not print indentation lines
  -p                    append "/" indicator to directories
  -Q, --quote           quote filenames with double quotes
  -r, --reverse         reverse order while sorting
  -S                    sort by file size
  -t                    sort by modification type, newest first
  -X                    sort alphabetically by entry extension
  -C                    do not count files and directories
  -R                    count root directories like other files
      --help            print this help and exit
```

Colors are currently blocked by  #2451

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/tree.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/tree.lua
@@ -115,10 +115,7 @@ local function list(path)
   end)
 end
 
-local function digRoot(root)
-  local rootPath = fs.realPath(shell.resolve(root))
-  coroutine.yield(stat(rootPath), {})
-
+local function digRoot(rootPath)
   if not fs.isDirectory(rootPath) then
     return
   end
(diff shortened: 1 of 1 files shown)


SDPhantom:
I can work on recursion later, I'm taking a break currently. I've developed methods that'll allow recursive operation without increasing the call stack by much, but it increases memory usage as a result.

payonel:
@Vexatos i've tested this and it works fine. But I would like your input on whether this should be an oppm project, or added to openos proper.

PR #2460, opened 2017-07-13
Upgrading Russian translation to the current state

cyber01:
Upgrading to the current state (compared to the original OS), translated the missing parts. Merged with @MoonlightOwl changes


fnuecke:
Thanks!

PR #2462, opened 2017-07-15
Fix renderLeash patch signature.

### LINKED ISSUE #2431: [1.11.2] Error running the class transformer
So here is the message that I get in-world:

My forge version: 13.20.0.2315
OpenComputer version: 1.7.0.24

Mod:
-Fast Leaf Decay
-Guide-API
-Inventory Tweaks
-JEI
-OpenComputers
-Waila
-Wawla

And my latest.log:
https://pastebin.com/ZQmaZFzw

Edit:
So, small update:

To figure if it was due because of a mod, I disabled (I'm using Twitch to manage minecraft) every mod but OpenComputer and I still seem to have that error.
Here is the latest.log of it: https://pastebin.com/Uuaadvch

thiakil:
Looks like the signature wasn't updated when the class name was. Should fix #2431


fnuecke:
Doh. Thanks!

PR #2465, opened 2017-07-18
Added Applied Energistics Support for 1.10 (rv4)

28Smiles:
I have reintegrated the support for Applied Energistics (Now rv4).
Please note that the Code compiles, but due issues between tile-registration-names (tis3d controller and ae controller) you are not able to run the build in the dev enviroment. Also i should point out, that i have added the libary for ae2-rv4, because there is no maven repository for ae right now.
Now, here are some screenshots, for testing i used FTB-Beyond and substituted the Open-Computers mod:

As you can see I tested the fundamental functions of the API, but I haven't checkt every function. Also this is the first time I programmed in Scala, so my code may not be very genius, but it is functional and isn't  Scala a functional language... :rofl: 

Please let me know what you think


build.gradle
```diff
@@ -208,7 +209,11 @@ dependencies {
     provided "appeng:RotaryCraft:${config.rotc.version}:api"
     provided ("appeng:appliedenergistics2:${config.ae2.version}:dev") {
         exclude module: 'buildcraft'
-    }
+    }*/
+
```
thiakil, line 213: You should only need to use the API jar. Even if that means some code changes (I can see some usages of an internal interface).

src/main/scala/li/cil/oc/integration/appeng/NetworkControl.scala
```diff
@@ -86,11 +93,12 @@ trait NetworkControl[AETile >: Null <: TileEntity with IGridProxyable with IActi
...
   }
 
-  @Callback(doc = "function():table -- Get a list of the stored fluids in the network.")
+  // NOT Supported right now
+  /*@Callback(doc = "function():table -- Get a list of the stored fluids in the network.")
   def getFluidsInNetwork(context: Context, args: Arguments): Array[AnyRef] =
     result(tile.getProxy.getStorage.getFluidInventory.getStorageList.filter(stack =>
-      !Mods.ExtraCells.isAvailable || ECUtil.canSeeFluidInNetwork(stack)).
-      map(_.getFluidStack).toArray)
+      !Mods.ExtraCells.isModAvailable || ECUtil.canSeeFluidInNetwork(stack)).
```
thiakil: EC is not needed to see fluids in the network. You can just remove the filter.

build.gradle
28Smiles: Oh ok, should I rewrite the whole thing? I'm confused, in the old Version also the whole Mod was used

src/main/scala/li/cil/oc/integration/appeng/NetworkControl.scala
28Smiles: Ok I see

build.gradle
thiakil: No not the whole thing, just the bits that use things not in the API as they may change. I know the old code uses the whole thing but it's generally not a good idea. And I do recall the reason this code is not already back in OC is because Alpha mods are not supported

28Smiles: I now looked into the api and there are many things which aren't accessible via api.
The api is made for creating new devices not accessing the existing ones

src/main/scala/li/cil/oc/integration/appeng/NetworkControl.scala
28Smiles: ECUtil is excluded and not available for compiling, and also Applied Energistics on it's own doesn't support Fluids

thiakil: The API itself does support fluids. Any mod can add them

build.gradle
thiakil: Just because it uses things that aren't in the api, doesn't mean there aren't other ways of doing it.

fnuecke: @28Smiles Thanks for updating the integration! What's the status here, i.e. do you think you can port the bits that currently require non-api classes to use the API?

src/main/scala/li/cil/oc/integration/appeng/NetworkControl.scala
fnuecke: @DrummerMC care to comment on this, seeing as you did the original fluid integration? Thanks!

DrummerMC: I used the filter to hide the gas fluids. EC is not updated. You can remove it.

build.gradle
28Smiles: @fnuecke 
> What's the status here, i.e. do you think you can port the bits that currently require non-api classes to use the API?

You could modify some Code to use the AE2-API Interfaces, but this wouldn't work for everything, so the lack of upgradeability stays due not completely using the api. I suggest to use this for now and update the whole AE2 integration to the api by using the Open-Computer adapter or adding an network access block or something like this. For now I think this is the simplest and best solution.

Edit: If you agree with me, I will start Programming something like this

thiakil:
`IGridProxyable.getProxy().get<blah>` -> `IGridHost.getGridNode(<side>).getGrid().getCache(<blah>.class)` (does need null checks)

`isInstanceOf[Part<Blah>]` -> `AEApi.definitions().parts().<blah>().isSameAs(IPart.getItemStack(PartItemStack.PICK))`

`AEItemStack.loadItemStackFromNBT` - the one instance I can find right now doesn't seem to need to store the AE stack, as the only function I can see referencing it just calls `getItemStack()` to return a vanilla stack.

If you do need the extra info, this one is an API oversight; you would need to save the MC `ItemStack` and potentially the `IAEItemStack.getStackSize()` if > MAX_INT, then load the itemstack and recreate the AEStack (though again, you'd either need to save the extra info you need and reload it)

Another way around this currently would be to write it to a packet and store the `ByteBuf`'s bytes and then load a reconstructed `ByteBuf` via the API. You still might lose some info as this is intended for client sync.

build.gradle
thiakil: I've added a comment on the main conversation thread with alternatives, if there are other non-api usages you can't work out I'd be happy to assist

src/main/scala/li/cil/oc/integration/appeng/DriverBlockInterface.scala
```diff
@@ -15,13 +15,14 @@ import li.cil.oc.util.ExtendedArguments._
 import li.cil.oc.util.ResultWrapper._
 import net.minecraft.item.ItemStack
 import net.minecraft.util.EnumFacing
+import net.minecraft.util.math.BlockPos
 import net.minecraft.world.World
 
 object DriverBlockInterface extends DriverSidedTileEntity {
```
28Smiles, line 24: @thiakil how about those?
`@Callback(doc = "function([slot:number]):table -- Get the configuration of the interface.")
    def getInterfaceConfiguration(context: Context, args: Arguments): Array[AnyRef]`
and
`@Callback(doc = "function([slot:number][, database:address, entry:number[, size:number]]):boolean -- Configure the interface.")
    def setInterfaceConfiguration(context: Context, args: Arguments)`

thiakil: `getInventoryByName`? That's part of `appeng.api.implementations.tiles.ISegmentedInventory`

thiakil: which is a superclass of `IInterfaceHost`

28Smiles: Ok thanks, i will try my best

28Smiles:
@thiakil I can't find an altenative to accessing IGridProxyable it isn't part of the Api and without this i cannot access the Network

thiakil:
I mentioned in one of my previous comments how you need to get around that. All of the `GridProxyable` methods are just wrappers for methods available on `IGrid`, you should be able to accesss the Grid by using `IGridHost.getGridNode(<side>).getGrid()` - each TE/Part should be an `instanceof IGridHost`

28Smiles:
@thiakil Thanks for your help
I now have rewritten DriverController and DriverInterface but I have trouble at the Parts.
`export.getConfigManager.getClass.getMethod`
How should I get/bypass the ConfigManager
(Line 65 at DriverExportBus)

thiakil:
That particular method should be fine, it's in `IConfigurableObject` - so as long as you know it is an export bus (and can then test for `IConfigurableObject`) calling that method is ok

The hardest part I can see for Parts is testing which part it is. e.g.

`EnumFacing.VALUES.map(container.getPart).exists(_.isInstanceOf[PartExportBus])`

would need to be something like (please forgive my terrible Scala) `EnumFacing.VALUES.map(container.getPart).exists(AEApi.definitions.parts.exportBus.isSameAs(_.getItemStack(PartItemStack.PICK))`

28Smiles:
Now everything is working! Big thanks to @thiakil in supporting me during the conversion of the code to fully use the API.
The AE2 integration is now fully Ported to Api, building and so far I can tell working properly.
@fnuecke I think now we can merge

src/main/scala/li/cil/oc/integration/appeng/DriverBlockInterface.scala
```diff
@@ -14,17 +17,20 @@ import li.cil.oc.integration.ManagedTileEntityEnvironment
...
-  def getTileEntityClass: Class[_] = classOf[TileInterface]
+  def getTileEntityClass: Class[_] = AEUtil.interfaceClass
 
-  def createEnvironment(world: World, x: Int, y: Int, z: Int, side: EnumFacing): ManagedEnvironment =
-    new Environment(world.getTileEntity(x, y, z).asInstanceOf[TileInterface])
+  def createEnvironment(world: World, pos: BlockPos, side: EnumFacing): ManagedEnvironment =
+    new Environment(world.getTileEntity(pos).asInstanceOf[TileEntity with ISegmentedInventory with IGridHost with IActionHost])
```
thiakil: `IActionHost` extends `IGridHost` ;)

src/main/scala/li/cil/oc/integration/appeng/DriverExportBus.scala
```diff
@@ -12,86 +21,90 @@ import li.cil.oc.util.ExtendedArguments._
...
+            var count = export.getInstalledUpgrades(Upgrades.SPEED) match {
+              case 1 => 8
+              case 2 => 32
+              case 3 => 64
+              case 4 => 96
+              case _ => 1
+            }
+            // We need reflection here to avoid compiling against the return and
+            // argument type, which has changed in rv2-beta-20 or so.
+            val fuzzyMode = (try export.getConfigManager.getClass.getMethod("getSetting", classOf[Enum[_]]) catch {
```
thiakil: I think this reflection could be removed? rv2 is long gone

src/main/scala/li/cil/oc/integration/appeng/DriverExportBus.scala
```diff
@@ -12,86 +21,90 @@ import li.cil.oc.util.ExtendedArguments._
...
+            val source = new MachineSource(export)
+            var didSomething = false
+            for (slot <- 0 until config.getSizeInventory if count > 0) {
+              val filter = AEApi.instance.storage.createItemStack(config.getStackInSlot(slot))
+              val stacks =
+                if (export.getInstalledUpgrades(Upgrades.FUZZY) > 0)
+                  itemStorage.getStorageList.findFuzzy(filter, fuzzyMode).toArray.toSeq
+                else
+                  Seq(itemStorage.getStorageList.findPrecise(filter))
+              for (ais <- stacks.filter(_ != null) if count > 0) { //TODO: Savety Copy removed no public clone function
```
thiakil: I can't understand the Scala here; which object doesn't have a public copy method?

src/main/scala/li/cil/oc/integration/appeng/DriverExportBus.scala
28Smiles: Sorry but i don't fully understand the fuzzy-mode code and i will not remove this if i don't exactly know what this means

28Smiles: Good idea i will change this

src/main/scala/li/cil/oc/integration/appeng/DriverExportBus.scala
28Smiles: The AEStack, it's protected

28Smiles: Forget it, my fault

28Smiles:
@fnuecke Sorry Florian, but I don't get the error. Can you please tell my why the CI  build failed?
Locally the Commit builds

thiakil:
Something on Travis is killing the process, possibly it's using "too much" memory, or taking too long

payonel:
These CI builds posted by PRs have been failing for a long time. I wouldn't worry about it. Give @fnuecke some time to return and review the work. We've discussed this work you've been doing and we do plan to merge it.

asiekierka:
Please note that AE2 now has a 1.12 release, too!

28Smiles:
Since this Implementation uses the Api it works(should) for all AE2 versions and MC versions

Donny208:
Looks great, really hoping this is added.

stephan48:
Hi, are there still any blockers left which could prevent this from being included? Currently thinking about porting my existing FTB Beyond RS System to AE cause I like AE way more so far - feels not as OP. But I would really like an OpenComputers Integration :) -- Stephan

PR #2466, opened 2017-07-18
Update forge version

modmuss50:
Dont merge just yet.

This PR updates the forge version to the latest. I plan to fix some of the other issues over the next few days.

## Known issues:

- ~~I disabled the re-mapping event for now, I will add that back in later~~

- Some items arent rendering correctly

The examples on the new FTB map seem to work fine apart from that NO testing has been done.

Note: This is my first time using Scala so please point out where I went wrong.

### DIFF
--- a/src/main/scala/li/cil/oc/common/init/Items.scala
+++ b/src/main/scala/li/cil/oc/common/init/Items.scala
@@ -523,7 +523,9 @@ object Items extends ItemAPI {
 
       override def getSubItems(tab: CreativeTabs, list: NonNullList[ItemStack]): Unit = {
         super.getSubItems(tab, list)
-        configuredItems.foreach(list.add)
+        if(isInCreativeTab(tab)){
+          configuredItems.foreach(list.add)
+        }
       }
     }, "misc")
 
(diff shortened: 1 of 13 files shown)


modmuss50:
Ok, I think this is all I can do. Apart from most of the items not rendering everything seems to work ok. If anyone has any input that could help me fix the items please let me know.

src/main/scala/li/cil/oc/common/init/Items.scala
```diff
@@ -52,26 +52,26 @@ object Items extends ItemAPI {
...
 
-  def registerBlock[T <: Block](instance: T, id: String): T = {
+  def registerBlock(instance: Block, id: String): Block = {
     if (!descriptors.contains(id)) {
       instance match {
         case simple: SimpleBlock =>
           instance.setUnlocalizedName("oc." + id)
           instance.setRegistryName(id)
-          GameRegistry.register(instance)
+          GameData.register_impl(instance)
```
thiakil, line 61: in 1.12 you need to use the event! `register_impl` is an internal use method as far as I know. The models should also usually be registered in the event, after the block/item itself has registered

modmuss50:
Humm,I think I tried that, it's a bit too much work for me to move everything to events, and after speaking to @Vexatos he agreed to stick with the GameData method. (I use it for all my other mods just fine) I could try registering the models in the event though.

fnuecke:
Thanks for getting it this far! I'll see how much of what remains I can get done today.

PR #2490, opened 2017-08-12
update to Minecraft 1.12.1, various fixes

asiekierka:
* Item texture registration finally works.
* The coremod has been adapted to only use SRG and deobfuscated names - far less maintenance required!
* Forestry's deprecated IAllele.getName() has been replaced with IAllele.getAlleleName()
* Ensured that duplicate JEI versions from other mods do not leak through to OpenComputers's workspace.


Vexatos (approved):

Vexatos:
We could technically make this mod compatible with both 1.12 and 1.12.1, but seeing that 1.12.1 is only a bugfix update itself and 1.12 mods are compatible with it, there is no reason not to use 1.12.1.

thiakil:
> The coremod has been adapted to only use SRG and deobfuscated names

I'm not sure that this works in a non-dev environment due to the checks of e.g. `if (ObfNames.Class_EntityLiving.contains(name.replace('.', '/'))) {`

`insertInto` should be logging a line like `log.info(s"Successfully patched ${classNode.name}.${methodNames(0)}.")` but I can find no such line in my log. Even adding extra logging directly inside the `if` statement does not log.

Vexatos:
Thanks for spotting this. It should be properly detecing classes again now.

thiakil:
Can confirm. :)

PR #2519, opened 2017-09-13
Upgrade AE2 API & Fix NPE in Import/Export/PartInterface drivers

stephan48:
Hi,

thanks for merging the AE2 Integration. Sadly I found that upon testing the export interface did not work like it should.

Instead I was greeted by a mile long stacktrace with a NPE which was caused by "container.getPart" returning null when no object could be found for a given side. Also I managed to crash the server twice by removing the fluix cable going to the Exportbus and the Exportbus itself.

I fixed it by adding a != null filter into the chain.

Since I don't know enough about the internals to be certain that "_.getItemStack(PartItemStack.PICK)" can't ever return null I added a second filter there as well.

I found the same issue in the ImportBus and PartInterface drivers to which i applied the same fix(after seeing similar NPEs).

Please excuse the verbose filter callback. I am not really familar with Scala(having used it for the first time during this debugging session) - feel free to make changes or point me into the right direction :)

I also took the chance to upgrade the AE2 Integration to the lastest version - rv4-stable-1. I think, this varrants a second commit but not really a second pullrequest. I will split it if needed/requested throu.

I confirmed the working of both changes on my own server(linux; FTB Beyond - All mods manually upgraded to upstream stable via twitch; self compiled build of OpenComputers from 1.10 master) on a ExportBus.

Kind Regards,
Stephan


thiakil:
The null (empty in 1.11+) check on the pick block shouldn't be required, but better to be safe I guess.

I do have concerns about the map+filter+map+filter+exists chain though - might need a scala wizard to take a look.

In psuedocode:

```
foreach EnumFacing
    if container.getPart(face) != null and container.getPart(face).getItemStack != null/empty
        AEUtil.isPartInterface(stack)
    end
end
````

**Comment by payonel** [comment]
@stephan48 thank you so much for doing this. I'm not the scala dev on the team but I hope to get some time this week to test this. :+1:

**Comment by stephan48** [comment]
Yes, I believe that the conditions can be greatly simplified - I just lack the dedication to scala to look up how :D 

It is a pleasure, especially as I am currently using it for a nice project(sort of automatic sorting system for putting Items from the ME System into Storage Drawers when they exist in both).

**Comment by thiakil** [comment]
What about something like (warning: noobish scala)

```
trait DriverPartBus {
  def hasPart(container: IPartHost, partFunc: (ItemStack) => Boolean) : Boolean = {
    EnumFacing.VALUES.exists(face => container.getPart(face) match {
      case part: IPart => part.getItemStack(PartItemStack.PICK) match {
        case is: ItemStack => partFunc(is)
        case _ => false
      }
      case _ => false
    })
  }
}
```

Then 

```
object DriverImportBus extends driver.SidedBlock with DriverPartBus {
  override def worksWith(world: World, pos: BlockPos, side: EnumFacing) =
    world.getTileEntity(pos) match {
      case container: IPartHost => hasPart(container, AEUtil.isImportBus)
      case _ => false
    }
```

28Smiles:
First, thank you for this, I saw your Issue and wanted to fix this, but then I also saw that you also want to make a pull request.
Good job

stephan48:
Hi again, I just decided to include @thiakil's suggestion into my build. It compiles without problems so far. Sadly my current world is not far enough to test this - and i stubbornly refuse creative :D Will test in a few days if it works!

Thanks!

--stephan

payonel:
I've recently merged re-integration of AE for the 1.10 and 1.12 branches, and I cannot consistently repro this, but i do see the exception. I'll merge this by hand. thanks for the work

payonel:
given the large change in the base i cherry picked this work. thanks for your help

PR #2535, opened 2017-09-20
Added ability to get real time

grandmaster8:


skyem123:
Ooo... I wonder if we should ever be allowed to access real time, the only uses are for (a) looking cool (b) interfacing with the real world.
I wonder how much it'd break to use Minecraft time for files...

payonel:
@Avaja sorry, the opinion of our group is to not expose real time, something something immersion :)

skyem123:
Here's a suggestion, maybe have the Internet card have a real time API? Because immersion is already broken then... You also cannot use NTP because that's UDP... (should I make a new issue for this?)

Fingercomp:
The in-game time is not reliable at all. It can be stopped (/gamerule) or skipped (/time). `computer.uptime()` does a better job, but it has a very low precision. If the server freezes, the reported time may lag behind, too. If the server stops or the chunk is unloaded, the computers would have no way to know that.

As @skyem123 said, there's the internet card, so "immersion" is already broken.

Some examples? Well, I've written quite a few programs, and needed to access the real time many times, so I can name some reasons why exposing it can be useful.

* It's used in some Internet protocols.
* Logging (players nearby, received messages, amount of energy stored, etc.). It's obvious that using the in-game time is a pain when searching the log history.
* Playing audio requires high-precision timing. There are some other uses for this, like profiling.
* Clocks. Well, I wouldn't call this useful, but it's still an example how the feature could be used.
* Some periodic actions. E. g., rebuilding the cache if it hasn't been updated for a week or longer (this could be used for OPPM).

Remember that `filesystem.lastModified` returns the real time. Why not make it less ugly, then?

PR #2544, opened 2017-10-06
Make actions work in the IRC client again..

CaitlynMainer:
disables printing CTCPs for now.. as this makes actions appear twice.
better fix?


payonel:
thanks @CaitlynMainer

skyem123:
Hmm... couldn't you just do an
```Lua
else
  print("[" .. name(prefix) .. "] CTCP " .. ctcp)
end
```
as the end of the `if` statement chain, instead of always (not) printing at the start?

CaitlynMainer:
yeah, didn't even consider it.. :/

PR #2599, opened 2017-10-25
Fix the Russian translation

Fingercomp:
Fixed lots of grammar mistakes, stylistic errors, translation errors, typos, etc. Probably some are still left, but the manual pages are readable now, at least.

I've kept the translation of achievements intact because of #1326.


AntonBoch1244 (approved):
Okay for me.

PR #2601, opened 2017-10-26
Allow getting yaw of player through tablet component

ghost:
Allow getting the yaw using the tablet component


payonel:
i like it

PR #2609, opened 2017-11-02
add redstone to chip diamond recipe

### LINKED ISSUE #1852: Recipe conflict Diamond Shards
This was reported to me by another user on the server I play, but it looks like there is a recipe conflict with one of the diamond recipes. A potential solution could be to OreDictionary nuggets as an optional item in the recipes, instead of just shards.
thing

payonel:
closes #1852


Vexatos:
But why redstone? That doesn't make much sense.

PR #2631, opened 2017-11-14
[1.12] Re-added AE, ExtraCells and Mekanism Integration

DrummerMC:
Testet with EC [1.12.2-2.5.2a18](https://ci.appveyor.com/project/DrummerMC/extracells2/build/1.0.18/artifacts)
and ae2 rv5-stable-2 on minecraft 1.12.2 (1.12.1 not working because ae and ec only supports 1.12.2)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,14 @@ repositories {
     maven { url "http://maven.cil.li/" }
     maven { url "http://maven.ic2.player.to/" }
     maven { url "http://tehnut.info/maven" }
+	ivy {
+        name 'Mekanism'
+        artifactPattern "http://addons-origin.cursecdn.com/files/${config.mekanism.cf}/[module]-[revision].[ext]"
+    }
+	ivy {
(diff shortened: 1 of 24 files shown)


build.gradle
```diff
@@ -66,6 +66,14 @@ repositories {
     maven { url "http://maven.cil.li/" }
     maven { url "http://maven.ic2.player.to/" }
     maven { url "http://tehnut.info/maven" }
+	ivy {
+        name 'Mekanism'
+        artifactPattern "http://addons-origin.cursecdn.com/files/${config.mekanism.cf}/[module]-[revision].[ext]"
```
thiakil: You should not need the full jar, nor to use curse.  API jars are available on Aidan's site or my maven @ http://maven.thiakil.com/Mekanism/

payonel:
thank you for switching to the api, this was a point i was concerned about. I'll continue to review this over the weekend

PR #2636, opened 2017-11-19
Miscellaneous tweaks to OpenOS

SquidDev:
I realise this is a bit of a hodgepodge of a PR, but it seemed cleaner to put these changes into one than creating lots of small ones. There isn't really any common theme to these commits: it's just features I found myself missing when messing with OpenOS. 

The commit messages describe each feature in a little more detail and the rational behind them. Feel free to cherry pick the ones you want (or I can create separate PRs for them if preferred).

 - Define `package.config` constant. 
 - The Lua REPL will work with expressions on there own.
 - Do not insert duplicate results into `term.read`/`tty.read`'s history.
 - Allow specifying an initial string for `term.read`/`tty.read`. The implementation for this is a little ugly, so better implementations are welcome!

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/core/lua_shell.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/core/lua_shell.lua
@@ -101,7 +101,10 @@ while tty.isAvailable() do
   if string.sub(command, 1, 1) == "=" then
     code, reason = load("return " .. string.sub(command, 2), "=stdin", "t", env)
   else
-    code, reason = load(command, "=stdin", "t", env)
+    code, reason = load("return " .. command, "=stdin", "t", env)
+    if not code then
+      code, reason = load(command, "=stdin", "t", env)
+    end
   end
   if code then
     local result = table.pack(xpcall(code, debug.traceback))
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/tty.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/tty.lua
@@ -32,11 +32,11 @@ function tty.key_down_handler(handler, cursor, char, code)
   elseif code == keys.enter or code == keys.numpadenter then
     cursor:move(math.huge)
     cursor:draw("\n")
-    if #data > 0 then
+    if data:find("%S") and data ~= handler[1] then
       table.insert(handler, 1, data)
       handler[(tonumber(os.getenv("HISTSIZE")) or 10)+1]=nil
-      handler[0]=nil
     end
+    handler[0]=nil
     return nil, data .. "\n"
   elseif code == keys.up or code == keys.down then
     local ni = handler.index + (code == keys.up and 1 or -1)


src/main/resources/assets/opencomputers/loot/openos/lib/package.lua
```diff
@@ -1,6 +1,7 @@
 local package = {}
 
 package.path = "/lib/?.lua;/usr/lib/?.lua;/home/lib/?.lua;./?.lua;/lib/?/init.lua;/usr/lib/?/init.lua;/home/lib/?/init.lua;./?/init.lua"
+package.config = "/\n;\n?\n!\n-\n"
```
payonel: [package.config](https://www.lua.org/manual/5.3/manual.html#pdf-package.config) would have limited value
1. The filesystem is always going to use / as a directory separator because OpenOS expects that, and the jvm real fs api then also uses /
2. Custom separators in the path could help with unusual package names and paths (though `;` and `?` are rather excellent defaults). However your PR doesn't include actually using this.
3. This isn't on windows, but OpenOS, and we're not using `!` to determine the library path
4. We're not using `-` 

We could add logic to use these fields, but I feel this is adding complexity with very little user benefit. Unless you have a strong counter to keep this, I'd ask that we remove this part from the PR

src/main/resources/assets/opencomputers/loot/openos/lib/core/lua_shell.lua
```diff
@@ -101,7 +101,10 @@ while tty.isAvailable() do
   if string.sub(command, 1, 1) == "=" then
     code, reason = load("return " .. string.sub(command, 2), "=stdin", "t", env)
   else
-    code, reason = load(command, "=stdin", "t", env)
+    code, reason = load("return " .. command, "=stdin", "t", env)
+    if not code then
+      code, reason = load(command, "=stdin", "t", env)
+    end
   end
```
payonel, line 108: should the return check actually be checked first instead of second?
consider `a()` which could be `return a()`

src/main/resources/assets/opencomputers/loot/openos/lib/tty.lua
```diff
@@ -270,6 +270,10 @@ function tty.stream:read()
   tty.window.handler = nil
   handler.index = 0
 
+  if handler.initial then
+    cursor:update(handler.initial)
+  end
+
   while true do
```
payonel: can you justify why a tty library needs a way to specify an `initial` string? We have a similar need for writing to the stdin (yes, writing) in response to vt100 code request for the cursor position, what my solution there lacks is that the data is not echo'd (via cursor:update() ). But I personally don't think it matters that much that case. Anyways - give me a real use case of this.

src/main/resources/assets/opencomputers/loot/openos/lib/tty.lua
```diff
@@ -32,11 +32,11 @@ function tty.key_down_handler(handler, cursor, char, code)
...
     cursor:move(math.huge)
     cursor:draw("\n")
-    if #data > 0 then
+    if data:find("%S") and data ~= handler[1] then
       table.insert(handler, 1, data)
       handler[(tonumber(os.getenv("HISTSIZE")) or 10)+1]=nil
-      handler[0]=nil
     end
+    handler[0]=nil
     return nil, data .. "\n"
```
payonel, line 40: not a bad suggestion! what do you think of `text.trim(data) ~= handler[1]` ?
edit: i'm actually referring to the code a few lines above this -- sorry

payonel:
I believe github's "Squash and merge" option for PRs would help us squash this PR into a single commit. Not to worry too much about the multiplicity of commits. Though, yes, we definitely want PRs to be few commits

payonel (changes requested):

src/main/resources/assets/opencomputers/loot/openos/lib/core/lua_shell.lua
gamax92: This is how Lua does it. https://github.com/LuaDist/lua/blob/lua-5.3/src/lua.c#L364-L380

src/main/resources/assets/opencomputers/loot/openos/lib/package.lua
SquidDev: > Custom separators in the path could help with unusual package names and paths 

Modifying `package.config` doesn't actually modify how the loader operates: they are compile-time constants.

> Unless you have a strong counter to keep this, I'd ask that we remove this part from the PR

There isn't much use for programs written specifically targeting OpenOS. However, if you're trying to run programs written for a general Lua environment it's nice to have. I don't really mind either way, as it's pretty trivial to write some sort of wrapper script.

src/main/resources/assets/opencomputers/loot/openos/lib/core/lua_shell.lua
SquidDev: The code will already attempt to use `return a()` before trying `a()`. I may be misunderstanding what you're saying though.

src/main/resources/assets/opencomputers/loot/openos/lib/tty.lua
SquidDev: Some of the fancier REPLs (such as IPython) will automatically indent lines following a block (such as within an `if`/`while`, etc...). I wanted to recreate a similar behaviour in OC. I realise this is pretty niche, so no complaints if it's a little messy.

I did experiment with alternatives (such as initialising the cursor with data), but sadly this is the only place where one can actual display the text to the user.

src/main/resources/assets/opencomputers/loot/openos/lib/core/lua_shell.lua
payonel: oh you're right (it tries "return " first), ha ... I must be going senile

src/main/resources/assets/opencomputers/loot/openos/lib/tty.lua
payonel: the way the tty read is being developed, the right way to do this might be to create a subtype of the cursor object, one that doesn't actually "return" its input until the code is complete. This custom cursor object could also prefix at will. I'll write up a prototype of this you can play with

PR #2639, opened 2017-11-22
Add GERTiClient to OpenOS

MajorGeneralRelativity:
Requested by payonel

This will add GERTiClient.lua to all computer installing OpenOS from a
new floppy drive. In order to enable full GERTi functionality, only one
computer must be provided with GERTiMNC.lua

GERT adds easier and more powerful networking utilities to OpenOS, and requires little additional space.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/GERTiClient.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/GERTiClient.lua
@@ -1,0 +1,485 @@
+-- GERT v1.0 - Release
+local GERTi = {}
+local component = require("component")
+local computer = require("computer")
+local event = require("event")
+local serialize = require("serialization")
+local modem = nil
+local tunnel = nil
+
+if (not component.isAvailable("tunnel")) and (not component.isAvailable("modem")) then
+	io.stderr:write("This program requires a network or linked card to run.")
+	os.exit(1)
+end
+
+if (component.isAvailable("modem")) then
+	modem = component.modem
+	modem.open(4378)
+
+	if (component.modem.isWireless()) then
+		modem.setStrength(500)
+	end
+end
+
+if (component.isAvailable("tunnel")) then
+	tunnel = component.tunnel
+end
+
+-- addresses
+local iAddress = nil
+local cachedAddress = {{}, {}, {}, {}, {}}
+local addressDex = 1
+
+-- Tables of neighbors and connections
+-- neighbors[x]{"address", "port", "tier"}
+local neighbors = {}
+local tier = 3
+local neighborDex = 1
+
+-- connections[x]{"destination", "origination", "data", "dataDex", "connectionID", "doEvent"} Connections are established at endpoints
+local connections = {}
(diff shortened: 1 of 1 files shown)


payonel:
**Is this a core behavior to OpenOS?**
This is not core to a simple runtime of OpenOS. Core OpenOS is designed to simplify using the minimal set of components. Here is a good list of the expected set of operations to OpenOS
1. Using a set of filesystems in a natural way (that's why I follow linux-isms with mounts, io)
1. Using a natural command line interface (inspired by bash)
1. Running scripts with natural/expected execution and error handling/reporting (error codes, error stacks)
1. Providing a minimal set of operating system support for processes (threads, events, scheduling)
1. Staying true to Lua
A network stack or network manager is not a part of this list

All additional helper libraries are best put in loot disks or oppm projects.

**Is this a loot disk contender?**
The basic needs of a modem or tunnel is sending a message from one computer to the next. Is this the minimal set of actions someone would want to use a modem for? It is not, it is attempting to provide a connection graph of nodes to establish a network. Thus, this would be an oppm project. Plan9k is one exception to this, partially to demonstrate to users that alternative operating systems are available.

Even if this was an appropriate core library, or a loot disk, consider:
1. this is an incomplete PR, there is no /bin application, no rc startup, no man page  to show how this would be used. Why is it called client? what is missing that the server has?
1. this network only repeats messages to the first neighbor, and could create loops
1. resolve doesn't search the whole network, assumes neighbor
1. as a lib/ component checks should not be cached as a system can change at runtime. Nodes leaving the network lose neighbors before notifying the network of the change. 
1. this code doesn't set gAddress, expects a user to know it, and never calls cacheAddress
1. the protocol is a bit messy depending on cached data and first neighbors. Sockets are opened only to cached nodes, sockets do not detect when the transmissions fail or a connection is closed.
1. Rebroadcasting packets is vulnerable to offline nodes (i.e. it'd fail)
1. This networking layer is changing too much about basic modem_messages as understood and seen by client scripts. Ideally, a network layer would be transparent.

Please note that resolving these concerns does not change the reasons I would not include this type of project into the core of openos. A network loot disk is not out of the question, but this isn't it -- a loot disk (as explained) would provide less overhead, less management and custom protocol. It would add the a minimal and transparent support to extend the basic modem api

MajorGeneralRelativity:
As evidenced by this pull request and the accompanying private messages, it is clear that my ideas are not taken seriously. I apologize for apparently wasting your time.

Techokami:
IMHO this would be better suited as a package for OpenPrograms.

MajorGeneralRelativity:
@Techokami OpenPrograms as in OPPM?

payonel:
> this would be an oppm project

I have the same opinion (as stated in my original response)

skyem123:
"OPPM" is the name of the tool used, "OpenPrograms" is the name of the organisation that the projects / repos are in.

MajorGeneralRelativity:
Ah ok. I just wanted to make sure I wasn't confusing it with a mod or something I hadn't heard of before.

Issue #2562, opened 2017-10-21
[Suggestion] Low-range wireless network card.


S0ZDATEL:
I think it's a good idea adding a low-range wireless network card, because I don't always need a wireless network card with a range of 400 blocks. Low-range wireless network card can have a range ~40 blocks, have reduced complexity and lower power consumption. They can simultaneously support open less ports than wireless network card.

skyem123:
That seems like an interesting concept...

Kiritow:
You can configure the range of a wireless card in game using Lua API. And the document also says lower range will consume lower energy.

S0ZDATEL:
You did not understand me. I need a wireless network card with a maximum range of 40 blocks. Network card configured for low range, will receive messages from a distance of 400 blocks.

skyem123:
Hmm... maybe a way to set receive range would be a good idea?

S0ZDATEL:
Or you can do so.

payonel:
i suppose we could consider a tier 1 wireless card

TheCodex6824:
So I was thinking I might try to add a tier 1 wireless card, but I want to get some feedback on some stuff before I would start:

1. 40 blocks for the max range sounds reasonable.
2. Power draw I was thinking would maybe be half the amount of the tier 2 card per block by default, which with default config settings would be .025 power per block.
3. Open ports I was thinking should get nerfed quite a bit on the tier 1 card, to make the tier 2 actually considerable in low-distance settings / make it so that everyone doesn't use the tier 1 everywhere to save a tier 2 slot. So, maybe 2 open ports max by default?

If anyone has any feedback about the above points or the existance of a tier 1 wireless card at all, I'd be glad to hear it.

payonel:
I think the tier 1 wireless card needs to be limited enough such that it is only a good solution in very controlled circumstances. We want the tier 2 wireless card to really be valuable in comparison, without having to increase(i.e. change our default settings for) the t2 wireless card

I'm open to feedback on this:

1. 16m radius
2. same power draw [ same cost per block for same data, no freebies! ]
3. 1 open port by default. [ you want a router? upgrade your wireless card ]

TheCodex6824:
[...]

PR #2657, opened 2017-12-05
Add tier 1 wireless card

TheCodex6824:
See #2562


payonel:
i'm pretty excited about this. thank you for your work. i'll review over the next couple of days, thank you

TheCodex6824:
No problem, let me know if there are any issues :)

skyem123:
Hmm... Question, what happens if one puts a T1 wireless card in a network relay?

TheCodex6824:
Currently I allow it to work. My line of thinking with that was that they already paid for the normal network card when crafting the relay, and 16 blocks is a lot less useful to relays.

skyem123:
I see...

TheCodex6824:
I can change it if people think that allowing the T1 card in relays is a bad idea

src/main/resources/application.conf
```diff
@@ -695,14 +695,14 @@ opencomputers {
       # exhaustion, attacking an entity generates 0.3 exhaustion.
       robotExhaustion: 10.0
 
-      # The amount of energy it costs to send a wireless message with signal
-      # strength one, which means the signal reaches one block. This is
-      # scaled up linearly, so for example to send a signal 400 blocks a
+      # The amount of energy it costs to send a wireless message using each tier of
+      # each wireless card with strength one, which means the signal reaches one block.
+      # This is scaled up linearly, so for example to send a signal 400 blocks a
```
payonel: while I understand the added value in having more configuration options, you're changing the types (in multiple places, here for cost, later for ports, and also range) from an int to an array, and I'll need to test if that causes issues(a crash?) for existing configs. Also, leaving the comment explanation unchanged might be easier to read, if we can and want to use an array for the config value type, perhaps just add a comment after this existing comment block sayingas something like, `# costs for tiers 1 and 2, respectively`

src/main/resources/assets/opencomputers/lang/en_US.lang
```diff
@@ -153,7 +153,8 @@ item.oc.UpgradeTank.name=Tank Upgrade
 item.oc.UpgradeTankController.name=Tank Controller Upgrade
 item.oc.UpgradeTractorBeam.name=Tractor Beam Upgrade
 item.oc.UpgradeTrading.name=Trading Upgrade
-item.oc.WirelessNetworkCard.name=Wireless Network Card
+item.oc.WirelessNetworkCard0.name=Wireless Network Card (Tier 1)
+item.oc.WirelessNetworkCard1.name=Wireless Network Card (Tier 2)
```
payonel, line 157: tedious, but, it would be nice to update the other lang files renaming the `item.oc.WirelessNetworkCard`

src/main/resources/assets/opencomputers/recipes/default.recipes
```diff
@@ -69,7 +69,7 @@ microcontrollerCase2 {
 }
 terminal {
   input: [[nuggetIron, "oc:solarGeneratorUpgrade", nuggetIron]
-          ["oc:circuitChip3", "oc:screen2", "oc:wlanCard"]
+          ["oc:circuitChip3", "oc:screen2", "oc:wlanCard2"]
           [nuggetIron, "oc:keyboard", nuggetIron]]
```
payonel: I understand that the remote terminal doesn't need a wired message device but unless we are changing the available component in code of the crafted terminal to have a tier1 card - this crafting recipe could give the wrong impression. Also, do we also want to reduce the range of the remote terminal? Perhaps not, perhaps it is best to leave this as a tier 2 wireless card.

src/main/scala/li/cil/oc/Settings.scala
```diff
@@ -193,7 +193,13 @@ class Settings(val config: Config) {
   val robotTurnCost = config.getDouble("power.cost.robotTurn") max 0
   val robotMoveCost = config.getDouble("power.cost.robotMove") max 0
   val robotExhaustionCost = config.getDouble("power.cost.robotExhaustion") max 0
-  val wirelessCostPerRange = config.getDouble("power.cost.wirelessCostPerRange") max 0
+  val wirelessCostPerRange = Array(config.getDoubleList("power.cost.wirelessCostPerRange"): _*) match {
+    case Array(tier1, tier2) =>
+      Array((tier1: Double) max 0.0, (tier2: Double) max 0.0)
+    case _ =>
+      OpenComputers.log.warn("Bad number of wireless card energy costs, ignoring.")
+      Array(0.05, 0.05)
```
payonel, line 201: i suppose this is how you're handling "migration" of old configs

src/main/scala/li/cil/oc/Settings.scala
```diff
@@ -523,6 +542,12 @@ object Settings {
     // Upgrading to version 1.5.20, changed relay delay default.
     VersionRange.createFromVersionSpec("[0.0, 1.5.20)") -> Array(
       "switch.relayDelayUpgrade"
+    ),
+    // Upgrading past version 1.7.0 (?), changed wireless card stuff for t1 card.
+    // TODO Update next version when its known what it will actually be
+    VersionRange.createFromVersionSpec("[0.0, 1.7.1)") -> Array(
+      "power.cost.wirelessCostPerRange",
+      "misc.maxWirelessRange"
```
payonel: and ports?

payonel: tbh, I don't know how this is used

src/main/scala/li/cil/oc/Constants.scala
```diff
@@ -161,7 +161,8 @@ object Constants {
     final val UpgradeContainerTier1 = "upgradeContainer1"
     final val UpgradeContainerTier2 = "upgradeContainer2"
     final val UpgradeContainerTier3 = "upgradeContainer3"
-    final val WirelessNetworkCard = "wlanCard"
+    final val WirelessNetworkCardTier1 = "wlanCard1"
+    final val WirelessNetworkCardTier2 = "wlanCard2"
```
payonel, line 165: Is it allowed to change oredic names (wlanCard -> wlanCard2)? How does this affect pre-existing worlds?

src/main/resources/application.conf
TheCodex6824: I'm on mobile so I can't link it but I added a config patch in Settings.scala for the wireless range and power cost, so it should just update it from older configs (which would mean the value would be reset to default, but this isn't the first time that happened). The old key for ports was changed to only apply to wired, with a new array for wireless.

The main reason why I had to change all that stuff was because nothing was set up to use multiple tiers of wireless cards - everything went off the assumption that there was only 1, hence the arrays replacing single ints/doubles.

src/main/scala/li/cil/oc/Constants.scala
TheCodex6824: Yeah this was already here, I'm assuming it was added when the tier 2 and 3 data cards were added. Pretty much it will remap the old name to the new one without breaking anything. I tested it on 1.7.10 with a lot of cards, both in chests / player inventory and in computers.

src/main/scala/li/cil/oc/common/tileentity/Relay.scala
```diff
@@ -232,6 +237,7 @@ class Relay extends traits.SwitchLike with traits.ComponentInventory with traits
       case driver if driver.slot(stack) == Slot.HDD => maxQueueSize = queueBaseSize
       case driver if driver.slot(stack) == Slot.Card =>
         isWirelessEnabled = false
+        wirelessTier = Tier.One
```
payonel: I don't think it adds any benefit to change this value - it is a bit misleading because if the "is enabled" booleans are respected than this value has no effect

src/main/scala/li/cil/oc/Settings.scala
TheCodex6824: The old port setting does not need to be changed because 16 for wired cards is find, wireless is a new setting.

src/main/scala/li/cil/oc/server/component/WirelessNetworkCard.scala
```diff
@@ -58,36 +52,49 @@ class WirelessNetworkCard(host: EnvironmentHost) extends NetworkCard(host) with
...
+    strength = math.max(0, math.min(args.checkDouble(0), maxWirelessRange))
     result(strength)
   }
 
   override def isWireless(context: Context, args: Arguments): Array[AnyRef] = result(true)
-
+  
+  override def isWired(context: Context, args: Arguments): Array[AnyRef] = result(shouldSendWiredTraffic)
+
+  override def open(context: Context, args: Arguments): Array[AnyRef] = {
```
payonel: could we instead share the same open check (i.e. NOT duplicate this method, and instead override the max ports in this wireless class?

payonel (commented):
summary of comments
1. is it okay to change oredic names?
2. remove open() from wireless and reuse network, override max ports instead
3. i think that's it? i dont know, review the comments

src/main/scala/li/cil/oc/Settings.scala
skyem123: Wouldn't this lose the value already set? That doesn't sound like a good idea...

src/main/scala/li/cil/oc/Constants.scala
skyem123: Does it work with crafting from other mods? I think that's what the OreDic is for?

TheCodex6824:
So stuff I will change when I get home later:

- [x] Update other language files for new card names
- [x] Remove open() override
- [x] Alias old card name to new one for crafting
- [x] ~Look into keeping old config values~ Decided against this because drawing attention to the new value that the tier 1 card has is probably a good idea, so that can be changed as well if needed
- [x] Although my testing showed it was all ok, to be 100% sure I'm going to load some very old server worlds I have and make sure nothing breaks

src/main/scala/li/cil/oc/Settings.scala
TheCodex6824: With the current config patch system, yes it would. There is precedent for that, I think some geolyzer values and power conversion rates were changed before. However, since the value is becoming an array, I might be able to update the config patcher to optionally  include preexisting values .

src/main/scala/li/cil/oc/Constants.scala
TheCodex6824: The main issue here is the item id itself, which would normally cause forge to get confused and say the world is missing items. However, OC handles the event for when forge finds an invalid item, and that's where the remaps come into play. To the user, everything is transparent. As for crafting, I will probably make wlanCard into an alias for wlanCard2

src/main/scala/li/cil/oc/common/tileentity/Relay.scala
TheCodex6824: I have to set this in order to have the relay respect the maximum signal strength for each card. If anything, it should replace isWirelessEnabled and should be set to -1 to mean no wireless.

skyem123:
Testing old worlds sounds like a very good idea. Sadly I haven't got any to test.

payonel:
if there are any concerns or problems that arise from changing the oredic name, I just want to throw it out there that it is an option to leave the old name for the T2 card unchanged, and only create a new name for this new card.

payonel:
also, this isn't going to be merged this week, vex and I are hardening the current dev build for another release first.

TheCodex6824:
I'm pretty sure I can add a remap for wlanCard and everything will work, going to be testing with computronics for that. As for waiting to merge I can't blame you, quite a bit got changed

src/main/scala/li/cil/oc/Settings.scala
skyem123: That sounds like a good idea.

src/main/scala/li/cil/oc/Constants.scala
skyem123: Alright, that sounds like it should work, though I have NO idea if it will because I haven't actually worked with MC mod code for a very long time. :P

payonel:
I did a small amount of testing and it looks great
Again, this'll be merged after we cut our next release
There are a few things I'm thinking over, but I'm happy to make a commit on top of this after I make some decisions
1. maybe i want all 3 configs in 1 array for max ports
1. oredic needs to specify 2 names for the t2 wireless card: wlanCard and wlanCard2

TheCodex6824:
Sorry for the late reply.

> maybe i want all 3 configs in 1 array for max ports

I thought about doing that and initially decided against it because I thought it would be confusing. After thinking it over a bit more, having it different config keys is probably *more* confusing, so I changed it to your suggestion.
>oredic needs to specify 2 names for the t2 wireless card: wlanCard and wlanCard2

It took me a while to push the OreDict stuff because I wanted to make sure it worked first, and Computronics did not have a recipe using the T2 wireless card like I thought. I ended up using OpenSecurity for testing and I got that working.

payonel:
I do want the two oredic names though, wlanCard and wlanCard2
We do want to add the new wlanCard2 name as an alias
Whether or not you want to revert the old wlanCard refs (such as in the recipes) is up to you. It makes sense to update them, makes them more readable perhaps

TheCodex6824:
I guess I'm a little confused why we need 2 entries. "oc:wlanCard" pointing to "oc:wlanCard2"  is only in there so that addon mods that were expecting the old item name in recipes still work. "oc:wlanCard2" is the actual name for the item, so an oredict entry would be redundant unless I'm missing something. All of the recipes in OC itself were already updated to use the new name as well.

payonel:
> "oc:wlanCard" pointing to "oc:wlanCard2" is only in there so that addon mods that were expecting the old item name in recipes still work

That's all I'm asking for

TheCodex6824:
> That's all I'm asking for

In that case, it's already done :)

payonel:
https://github.com/MightyPirates/OpenComputers/commit/22c6b15d392cbb223d0d4f8e2d75d3b3ea970da0

payonel:
moving comments on commit to PR discussion:

> TheCodex6824 
> Sorry I keep bothering you but I noticed that the cpuComponentCount config array for the creative component bus isn't in the config patch list either. My bad for not catching that with the other one...

> Xyxen 
> Yeah, I wasn't sure about that. Seemed like it'd be  better to have a warning in the log on each launch so server owners  would know specifically why their old settings weren't in effect, but  perhaps there's other infrastructure in place. Thinking about it now, it  would cause default configs on existing worlds to warn on each launch  too, wouldn't it? Thanks for catching it. Perhaps I could draft up a couple of procedures/checklists on the Wiki (or in a markdown file) for changes like this?

> TheCodex6824 
> Yeah config patching could probably benefit from a log message every time it has to overwrite something in the config. I pointed it out because OC won't like that the old array only has 3 things in it instead of 4 and will always end up using default values until someone adds in a 4th.
And +1 to the checklist file, sounds like it would help out a lot.

PR #2661, opened 2017-12-06
[1.10] Remove @Optional annotations that specify our lowercase mod ID

### LINKED ISSUE #2649: injectEnvironmentImplementation should remove @Optional annotations
When [injectEnvironmentImplementation](https://github.com/MightyPirates/OpenComputers/blob/06e7149573cc310254b691b3944eeb5d99619ff8/src/main/scala/li/cil/oc/common/asm/ClassTransformer.scala#L268) makes its changes to a class implementing SimpleComponent, it should remove any @Optional annotations for the interface itself as well as for its methods, to ensure that Forge doesn't strip them but leave our injections. See McJty/RFTools#1466 for details as to why.

josephcsible:
Fixes #2649. Explanation:
* Other mods can integrate with us by implementing the SimpleComponent
  interface. They annotate its methods with @Optional to avoid a hard
  dependency on us.
* Our mod ID changed from "OpenComputers" in 1.10 to "opencomputers" in 1.11.
* Some mods (like RFTools) can't refer to our mod by the right-case mod ID,
  because the same .jar works on both 1.10 and 1.11.
* In most cases, using the wrong-case mod ID would just mean that integration
  wouldn't work, but for us, it causes a crash.
* The crash is because we use ASM to inject some extra methods and make the
  class implement a different interface. This means that if the @Optional mod
  ID is wrong, our new interface will still be there but some of the original
  methods won't, which leads to AbstractMethodErrors like McJty/RFTools#1466.
* Forge looks for the @Optional annotations before our coremod can do
  anything, so we can't just remove them from the class. Instead, we need to
  remove them from Forge's internal list.

### DIFF
--- a/src/main/scala/li/cil/oc/common/asm/ClassTransformer.scala
+++ b/src/main/scala/li/cil/oc/common/asm/ClassTransformer.scala
@@ -1,10 +1,13 @@
 package li.cil.oc.common.asm
 
+import com.google.common.collect.ListMultimap
 import li.cil.oc.common.asm.template.SimpleComponentImpl
 import li.cil.oc.integration.Mods
 import net.minecraft.launchwrapper.IClassTransformer
 import net.minecraft.launchwrapper.LaunchClassLoader
+import net.minecraftforge.fml.common.asm.transformers.ModAPITransformer
 import net.minecraftforge.fml.common.asm.transformers.deobf.FMLDeobfuscatingRemapper
+import net.minecraftforge.fml.common.discovery.ASMDataTable.ASMData
 import org.apache.logging.log4j.LogManager
 import org.objectweb.asm.ClassReader
 import org.objectweb.asm.ClassWriter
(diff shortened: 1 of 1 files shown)


josephcsible:
Note that because of the nature of this problem, no Minecraft versions newer than 1.10.2 need this fix or any equivalent.

payonel:
why is this needed for 1.10 then if the name didn't change until 1.11?

josephcsible:
So that mods like RFTools can use the 1.11 mod ID in 1.10 without crashing. It would otherwise be impossible for RFTools to integrate with OC, since it uses the same jar for 1.10 and 1.11, and annotation values (which the mod ID is) have to be compile-time constants.

payonel:
right....that makes sense :)
doesn't this depend on our api being the same between 1.10 and 1.11?

josephcsible:
Yes, but that can be worked around in the other mods' code, unlike this mod ID issue. (And nothing that RFTools uses appears to have changed.)

payonel:
yes very good - I wasn't overly concerned about that aspect for that very reason you explain
I'm reviewing the code right now, I'll have some nit picks (hopefully that's okay with you)
I'll be done in a few minutes

payonel:
Actually I think the code looks good. Thank you! I'll test this and merge soon

payonel:
so this resolves the world crash, good, but -- we can't use the rftools blocks from oc
if we try to use a component proxy, it fails:
image

We could handle this if we replaced the mod id with the upper case name..? any thoughts on that?

Note that this PR is definitely better as is than nothing, because no crash on world load, so that's good.

josephcsible:
That seems to be just a cosmetic error. The getters and setters all seem to work, even though you can't = the object as a whole.

Also, how would replacing the mod ID with the uppercase version fix that?

payonel:
Perhaps you are unfamiliar with our lua repl but returning a table invokes a serializer on the object, and the it is a fast way of listing the api on a proxy. It was simply a short test I ran. Manually checking the api on the object does confirm what you say - so apparently I have found a bug in my own openos code that I'll have to investigate at another time.

As for how this would fix anything -- I see now that makes no difference :) I'm a bit sleep deprived

josephcsible:
Does that mean that no changes are needed to this PR then?

payonel:
no changes are needed, the code is good thank you. I'll merge soon

josephcsible:
This was only intended for the 1.10 branch, but it looks like it ended up in 1.11 and 1.12 too.

payonel:
yeah, it snuck it :) thanks to some automerging i didn't double check. thanks

PR #2668, opened 2017-12-10
Make cable collision box closer to cable shape

SquidDev:
This overrides `addCollisionBoxesToList` to ensure the entity collision boxes match the shape of cables a little more.

My Scala knowledge is pretty minimal, so I'm open to potential improvements there.

This'll require changes to run on 1.8.9 and 1.9.4. I'd be happy to provide patches for those if necessary.

## Before and after shots
Before After

Sorry that they're not the same size. The texture errors aren't a problem with this PR, just something broken in my development setup.

### DIFF
--- a/src/main/scala/li/cil/oc/common/block/Cable.scala
+++ b/src/main/scala/li/cil/oc/common/block/Cable.scala
@@ -1,5 +1,7 @@
 package li.cil.oc.common.block
 
+import java.util
+
 import codechicken.lib.vec.Cuboid6
 import codechicken.multipart.JNormalOcclusion
 import codechicken.multipart.NormalOcclusionTest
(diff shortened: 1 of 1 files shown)


payonel:
We are now only supporting 1.7.10, 1.10, 1.11, and 1.12
btw, i like this change. I'll review and test for merge soon(tm)

SquidDev:
> We are now only supporting 1.7.10, 1.10, 1.11, and 1.12

Oh good, that makes things a little easier. They'll still be changes required for 1.10.2, but I'm pretty sure it's stable after that.

payonel:
could you make a 2nd PR then for what I can merge from 1.10 and up?

payonel:
there appears to be a jitter problem
Walking into the cable from the -x side, or from the -z side, gets you somewhat stuck. If you walk at certain angles very near perpendicular, the jitter doesn't happen and instead you step over the cable
Walking into the cable from the positive side has no issue.


jitter

SquidDev:
The calculations for each segment/part's bounding box was slightly off: resulting in overlapping bounding boxes and other issues. Whilst it would be possible to fix them, it ended up being simpler to specify a list of each part's dimensions.

PR #2670, opened 2017-12-14
Add creative component bus

### LINKED ISSUE #2633: Creative component bus?
for 1000 components

Xyxen:
Supports 1024 components by default, and fits into a tier 3 slot. Resolves #2633.

I'm happy to build patches for 1.10 through 1.12 as well. This is a backport from 1.12 anyway, because building on 1.7.10 was a bit fussy at first.


src/main/scala/li/cil/oc/common/init/Items.scala
```diff
@@ -540,6 +540,8 @@ object Items extends ItemAPI {
     registerItem(new item.DiamondChip(multi), Constants.ItemName.DiamondChip)
     Recipes.addSubItem(new item.UpgradeMF(multi), Constants.ItemName.MFU, "oc:mfu")
 
+    registerItem(new item.ComponentBus(multi, Tier.Four), Constants.ItemName.ComponentBusCreative)
```
payonel, line 543: add comment for version here, // 1.7.2
as was done above previously

payonel: i'll add this comment later

Xyxen: Thanks. I wasn't sure if that should wait; I'm not used to adding version numbers before hitting a milestone.

PR #2671, opened 2017-12-15
[MC1.10] Add creative component bus

### LINKED ISSUE #2633: Creative component bus?

Xyxen:
Continuation of #2670. Should be the last patch that resolves #2633, unless you'd like 1.11 and 1.12. I'm pretty sure it's just case changes in those versions. Let me know?


payonel:
It was very nice of you to do this, but, is there anything different here than in the previous PR to 1.7.10?
I generally build up changes on 1.7.10 and continue testing, and from time to time I merge all commits up to 1.10->1.11->1.12, checking if things break and making adjustments.
If the work done you did for 1.7.10 is the same as these let's close this PR. Thanks!

Xyxen:
Pretty sure the only differences are the model (`componentBusCreative.json`), texture name, and requisite change in `Items.scala` for the new structuring there. It's pretty small changes, so definitely feel free to close this if those are routine and expected.

payonel:
ok i'll leave this open until I do the merges and compare to this, then I'll close this. Thanks

src/main/resources/assets/opencomputers/doc/en_us/item/componentBusCreative.md
```diff
@@ -0,0 +1 @@
+#REDIRECT componentBus1.md
```
TheCodex6824, line 1: The actual content of the file is fine, but the path of the file should be in the en_US folder and not en_us (capitalization) for the 1.7.10 version at least.

Xyxen: Oh, the joy of backporting on Windows. I probably forgot to reset my index before committing, so the 1.12 capitalization stayed. I'm pretty sure this'll cause issues on Linux, so it should be fixed.

payonel:
hi @Xyxen ... i was working hard to merge my sheep power changes from 1.7.10 up that I forgot you made this PR to help merge the creative component bus work ... so now we've both done it and I've merged my commits already. Sorry to have you do this PR for no reason :( However, I suppose in some benefit I have checked all of the changes you made for this PR and confirmed that we made the same merge choices. I've also tested my work here and on 1.11 and 1.12 -- so we're good. Thanks!

PR #2673, opened 2017-12-16
Fix runClient and runServer Gradle tasks

### LINKED ISSUE #2561: Unable to run oc from gradle
When attempting to run oc from gradlew runClient, I receive a forge error telling me that there are 2 instances of ae2, any help?
MC Log: [https://pastebin.com/D60tX652](https://pastebin.com/D60tX652)

### LINKED ISSUE #2614: runClient gradle Task doesn't want to work.
So it seems that for some reason it loads the AE2 jar twice?

https://gist.github.com/Inari-Whitebear/e6bd334deb41b74b7bc37377dfe547f3

Xyxen:
`compile` dependencies are placed on the runtime classpath by Gradle. Because the `compile` configuration is set to extend from the `provided` config, all of those dependencies are put on the runtime classpath as well. That might be okay, except the set of dependencies certainly don't make up a functional modpack, so you end up with Forge spitting up dependency problems before you can reach the menu.

In this patch, `provided` is instead made a separate configuration, and explicitly added to ONLY the compile classpath of the main sourceset. [This is what newer versions of ForgeGradle do,](https://github.com/MinecraftForge/ForgeGradle/blob/f49079dd11c037f74cc740285b2c5504fbf8a7f6/src/main/java/net/minecraftforge/gradle/user/UserBasePlugin.java#L502-L515) and it's also what's being done by setting dependencies to IDEA's "Provided" scope in [the wiki's build instructions.](http://ocdoc.cil.li/tutorial:debug_1.7.10)

Resolves #2561 and resolves #2614.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,7 +178,7 @@ repositories {
 configurations {
     provided
     embedded
-    compile.extendsFrom provided, embedded
+    compile.extendsFrom embedded
 }
 
 dependencies {
@@ -229,6 +229,9 @@ dependencies {
     testCompile "org.scalactic:scalactic_2.11:2.2.6"
     testCompile "org.scalatest:scalatest_2.11:2.2.6"
 }
+
+// Add the "provided" dependencies to the compile (but NOT runtime) classpath.
+sourceSets.main.compileClasspath += [configurations.provided]
 
 idea.module.scopes.PROVIDED.plus += [configurations.provided]
 // TODO Causes errors on Gradle 2 for me (No such property: allDependencies for class: java.io.File).


payonel:
can you make it so the dependencies for OpenComputers_main also import correctly? this seems to only fix OpenComputers_api

also, for 1.7.10 the _api and _main moduels are both pointing to the same output path: `OpenComputers/build/classes/main` and it has to be fixed manually.

Xyxen:
This patch was only intended to fix running from the command line, but I'm happy to go further with it to fix IDEA project import/generation. I'll see what I can do.

payonel:
if  you find what is needed for the other fixes, i'd appreciate the PR for those things. Thanks

PR #2682, opened 2017-12-18
[1.12.2]Chinese Font Lost

Stevexmh:
Some chinese font can't show successfully and become "?".
I don't have any skill in java programing, so I can't help you to fix it.
So just a point:
Chinese Font Unicode range is 4E00 - 9FA5.
Hope can fix it!


payonel:
I need to close this because it doesn't provide any updates to our font.hex and it includes commits that we don't want to merge

we are using our custom font format now, see https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/resources/assets/opencomputers/font.hex

if you want to add characters to our font table, you can add them to that font.hex file

make the change on the master-MC1.7.10 branch, then make a PR to that branch

PR #2683, opened 2017-12-18
Fix the crafting upgrade to use SlotCrafting

### LINKED ISSUE #2532: [MC1.12.1-1.7.0.11] Robot crafting destroys item recipe after crafting
If you craft an item with a robot, that recipe's output item will be replaced with air, breaking the recipe until the server/client is restarted. I tested this with Forge 1.12.1-2485 and with the latest dev build of OpenComputers MC1.12.1-1.7.0.11 and no other mods.

habnabit:
This handles basically all of the funky corner cases around crafting,
including things like decrementing stacks and repairing items and all
of the necessary event posting. Tested to work correctly with IC2's
Forge Hammer too.

Should fix #2532.

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/UpgradeCrafting.scala
+++ b/src/main/scala/li/cil/oc/server/component/UpgradeCrafting.scala
@@ -13,20 +13,14 @@ import li.cil.oc.api.machine.Arguments
 import li.cil.oc.api.machine.Callback
 import li.cil.oc.api.machine.Context
 import li.cil.oc.api.network._
-import li.cil.oc.api.prefab
 import li.cil.oc.api.prefab.AbstractManagedEnvironment
 import li.cil.oc.util.InventoryUtils
 import net.minecraft.entity.player.EntityPlayer
 import net.minecraft.inventory
-import net.minecraft.item.ItemStack
+import net.minecraft.inventory.{InventoryCraftResult, SlotCrafting}
 import net.minecraft.item.crafting.CraftingManager
-import net.minecraft.item.crafting.CraftingManager
-import net.minecraftforge.common.MinecraftForge
-import net.minecraftforge.event.entity.player.PlayerDestroyItemEvent
-import net.minecraftforge.fml.common.FMLCommonHandler
 
 import scala.collection.convert.WrapAsJava._
-import scala.collection.mutable
 import scala.util.control.Breaks._
 
 class UpgradeCrafting(val host: EnvironmentHost with internal.Robot) extends AbstractManagedEnvironment with DeviceInfo {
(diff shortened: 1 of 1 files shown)


payonel:
does this only work for 1.12? Is this not the solution for 1.11? If this is 1.11 compatible, please retarget the PR to 1.11

payonel:
did much testing, ONLY affects 1.12

SirDavidLudwig:
@payonel 
There seems to be a pretty big issue with this fix. It seems to keep crafting as long as *any* a recipe is available, sort of in a loop, even if that recipe changes. For example, let's say that I want to craft as many iron blocks as possible, so I don't specify a quantity of output items when I call `crafting.craft()`. My output slot is somewhere outside of the crafting grid. I have a 3x3 grid of iron ingots, 1 ingot in each slot, but I have 2 ingots in one of the 9 slots. It will craft the first iron block, but then it still finds a recipe because of the one remaining iron ingot, so it also crafts iron nuggets.

habnabit:
@SirDavidLudwig The exact same thing happens if you craft on a crafting table—you place down the ingots, and you first get the block, and then get the nuggets. Calling `.craft()` does, like, 64 crafts or something. So it's as if you shift-clicked the crafting output 64 times. `.craft(1)` would do only one craft and you'd only get the block.

Basically I don't see this as necessarily a _bug_ but I'm not sure how it should be resolved. What happened on older versions of OC?

skyem123:
Could there be a way to give extra information to the program? Sort of a way to differentiate between "the recipe has changed and isn't craftable ", "the recipe has changed and is still craftable", "the recipe has not changed, and is still craftable"? Extra function or return value, perhaps?

SirDavidLudwig:
@skyem123 Yeah, that's the straight forward solution, which you can do by only adding 1 extra line of code. @habnabit This really shouldn't be though. In a crafting table, you would keep crafting until you can't craft the item you want. You wouldn't keep crafting the wrong item like the bot. It forces the user to now check manually how many times it can be crafted, instead of crafting as many as possible.

SirDavidLudwig:
So here is a simple solution that seems to work great. You are already calculating how many you can possibly craft, so let's use it in the while loop's condition, essentially turning it into
```scala
def craft(wantedCount: Int): Seq[_] = {
    copyItemsFromHost()
    var countCrafted = 0
    var maxCraft = amountPossible min wantedCount
    val canCraft = CraftingManager.findMatchingRecipe(CraftingInventory, host.world) != null
    breakable {
        while (countCrafted < maxCraft) {
            val recipe = CraftingManager.findMatchingRecipe(CraftingInventory, host.world)
            if (recipe == null) break()
            val output = recipe.getCraftingResult(this)
            craftResult.setRecipeUsed(recipe)
            craftingSlot.onTake(host.player(), output)
            countCrafted += 1
            copyItemsToHost()
            InventoryUtils.addToPlayerInventory(output, host.player)
            copyItemsFromHost()
        }
    }
    Seq(canCraft, countCrafted)
}
```

habnabit:
@SirDavidLudwig That's the thing, though: I don't consider this _wrong_. When I was putting together this patch I actually specifically considered this use case and thought it was reasonable. 

I'm still curious what older versions of OC did here.

@skyem123 by "the program" do you mean the lua code? Actually, now that I think about it, there's already a "can this be crafted" function: call `.craft(0)` and see if it returns `true` or `false`.

payonel:
even if "stop when recipe changes" is a new concept, i think it is worth discussing

mkualquiera:
I have to agree with @SirDavidLudwig 
Just in terms of the context, if you went to a crafting table and pressed shitf-click on the item to craft, you would not get two different crafting results in your inventory. Only one recipe would be used, and you would have to shift-click again.
This "feature" of crafting as long as there is any recipe available would remove the possibility of instantly crafting as much as possible without having unwanted products. 
However, I think it could be made optional somehow, as there are some very specific cases where this could be useful...

SirDavidLudwig:
@habnabit So, here is my issue, and a lot of other's issues that I've discussed this with. I can see where you're coming from, sure, but this is going to make it extremely inconvenient for the users. The problem is now, the `crafting.craft()` method alone like this without args is now essentially useless. Let's be realistic here. If I'm making an auto-crafter that is supposed to turn iron ingots into iron blocks, I now have to scan through manually, count the items, and manually determine the max output of the recipe, when, like most people I've discussed this with, the "Crafting Upgrade" should do this for you. If anyone really wanted this "Keep crafting as long as any recipe is available" thing, they could stick it in a loop in OC. This is significantly easier than trying to count the amount of output items from my given ingredients.

skyem123:
That's actually a good point... People would expect behaviour similar to the crafting table here...

neumond:
This becomes funny if you place 9 nuggets as recipe for ingot and `robot.select(1)` then `craft()`. You get same 9 nuggets, but in 1st slot.

habnabit:
So, did anyone try this on 1.7.10? I'll do it tonight otherwise.

payonel:
@habnabit yes do please test it and submit a PR if you have a suggested change

habnabit:
@payonel mostly I was trying to use that to judge if there should be a new
bug opened (if this has been the behavior and has just now been noticed) or
this PR should be reverted and redone maybe with a new API.

payonel:
I support this change as an accepted feature request as long as it is tested and appears to work. there does not need to be a new api. I'm okay if the crafting simply stops, the user would need to call again to continue crafting the new recipe. At this time I would prefer to leave the commits of this PR as is (will not revert). You can add to this work with an additional PR if you would like

agentenc:
im really bad at the mod i just want the robot to craft when i input the code,guys how do i make it craft? like what line do i need to input to craft

payonel:
@agentenc see our forums ( https://oc.cil.li/ ) or our irc channel ( #oc on esper net ), which is also reachable via our discord ( https://discordapp.com/invite/0hVukoQ2KYm2aoXh ) and some documentation on our wiki ( https://ocdoc.cil.li/ )

PR #2701, opened 2017-12-27
Change retard condition, so native libraries are available at other OSes

ghost:
So I can run native Lua libraries at OpenBSD...


skyem123:
1. That term is open to misinterpretation in the English language
2. Maybe call it `native_manual` without any file extensions?
3. It would be good to create individual requests for any missing natives, so your work can help others.

ghost:
I thought it reads files from /path/to/minecraft/assets/opencomputers/lib -_-

skyem123:
Seriously though, it would be a very good idea if you help get the binaries building for unmodified later versions of Eris with a PR there and also https://github.com/MightyPirates/OC-Natives so people don't have to bother with work arounds! ^^;;

Issue #2697, opened 2017-12-25
[Question] Which 1.8 branch should I submit a pull request to?


brisingraerowing:
In 1.8+ the Debug Card's world.getMetadata method is broken and returns the block state and not the numeric metadata. I think I know how to fix it (and I can add a getBlockState method as well, returning the string representation of a block state), but I need to know which 1.8.x branch I pull into so I can avoid having to redo the work if I use the wrong branch.

This doesn't affect 1.7.10 as that version doesn't use block states.

payonel:
master-MC1.8.9
But, 1.8 is discontinued. I would accept a PR to that branch, but:
1. I won't be pushing any more 1.8 curseforge builds, and
1. I won't merge your commits up the branches (like we do for the other changes, from 1.7.10 to 1.12)

1.9 is also discontinued

brisingraerowing:
Ah, ok. I'll do a pull into 1.10 then.

It may be a bit. My dev env is goofed up and I don't know what happened.

brisingraerowing:
I'm actually going to implement a converter for a block state (which I already have done) and use that to return a Lua table representing the block state, allowing for easier use of the data in programs.

I ended up doing some work in a non-forked repo before I realized that, so I copied out what I had done and I am now forking the repo to be able to create a pull. I hope to have this implemented before too long. The converter was rather easy, even though I don't really know much about modding.

PR #2705, opened 2017-12-28
Fixed #2313

### LINKED ISSUE #2313: Incorrect documentation / return value of world.getMetadata() in Debug Card
This one is fairly straightforward; getMetadata() says it returns a number matching the metadata of a block, but it actually returns a string with the block's textual ID and metadata value (which itself can be a string, see the second image).

The results in the second image *are* useful, but not when you need the **numeric** metadata

image

image

brisingraerowing:
Added getBlockState method to debug card world value (returns a string as my converter attempt failed)

The getMetadata method now returns the numeric metadata instead of a block state string.


payonel:
closes #2697 I presume?

brisingraerowing:
Yup. Forgot that one.

Issue #2713, opened 2018-01-01
OpenOS - lib/uuid - Invalid Version 4 UUID As Per RFC4122


jobe1986:
As per RFC4122 a version 4 UUID (entirely random) should have a "4" in the first position of the third group of hex digits, and either a 8, 9, A or B in the first position of the fourth group of hext digits.

The following version of openos/lib/uuid.lua resolves this issue:
```
local uuid = {}

function uuid.next()
  -- e.g. 3c44c8a9-0613-46a2-ad33-97b6ba2e9d9a
  -- 8-4-4-4-12 (halved sizes because bytes make hex pairs)
  local sets = {4, 2, 2, 2, 6}
  local result = ""
  local pos = 0

  for _,set in ipairs(sets) do
    if result:len() > 0 then
      result = result .. "-"
    end
    for i = 1,set do
      local byte = math.random(0, 255)
      if pos == 6 then
        byte = bit32.bor(bit32.band(byte, 0x0F), 0x40)
      elseif pos == 8 then
        byte = bit32.bor(bit32.band(byte, 0x3F), 0x80)
      end
      result = result .. string.format("%02x", byte)
      pos = pos + 1
    end
  end

  return result
end

return uuid
```

payonel:
it was never meant to match the official uuid. it was just to provide a random string

payonel:
make a pr for this if you feel that strongly about it. i'll consider it

PR #2715, opened 2018-01-02
Update openos/lib/uuid.lua to generate valid RFC4122 version 4 UUIDs

jobe1986:
Pull request relating to issue #2713

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/uuid.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/uuid.lua
@@ -5,13 +5,21 @@ function uuid.next()
   -- 8-4-4-4-12 (halved sizes because bytes make hex pairs)
   local sets = {4, 2, 2, 2, 6}
   local result = ""
+  local pos = 0
 
   for _,set in ipairs(sets) do
     if result:len() > 0 then
       result = result .. "-"
     end
     for i = 1,set do
-      result = result .. string.format("%02x", math.random(0, 255))
+      local byte = math.random(0, 255)
+      if pos == 6 then
+        byte = bit32.bor(bit32.band(byte, 0x0F), 0x40)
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/opencomputers/loot/openos/lib/uuid.lua
```diff
@@ -5,13 +5,21 @@ function uuid.next()
...
     for i = 1,set do
-      result = result .. string.format("%02x", math.random(0, 255))
+      local byte = math.random(0, 255)
+      if pos == 6 then
+        byte = bit32.bor(bit32.band(byte, 0x0F), 0x40)
+      elseif pos == 8 then
+        byte = bit32.bor(bit32.band(byte, 0x3F), 0x80)
+      end
+      result = result .. string.format("%02x", byte)
+      pos = pos + 1
```
SquidDev, line 22: Can one just use `i` instead of maintaining a separate counter?

jobe1986: The value of i is from 1 to the value of 4, 2 or 6 depending on the part of the uuid being generated. Thus you would still need to maintain an index of which section of the uuid you're generating.

For example the 2 options are the method I used, or a method where the option for byte 6 is byte 1 or part 3 and byte 8 is byte 1 of part 4. So you could change "for _,set in ipairs(sets) do" to "for part,set in ipairs(sets) do" and then use "if part == 3 and i == 1 then" and "if part == 4 and i == 1" instead of "if pos == 6 then" and "if pos == 8 then" respectively. The result would then be the same.

SquidDev: Ahhh, woops. I evidently failed to read the code properly - sorry.

Issue #2693, opened 2017-12-22
(Un?)managed memory


SoniEx2:
I'm not sure what's what in terms of managed vs unmanaged so w/e.

Anyway, this memory would give you a raw block of memory and hopefully the OS (OpenOS) would provide an MMU.

Functions would include setInt getInt setFloat getFloat all the stuff you're used to and also compareAndSwap atomicAddInt etc.

We can then add real threads (well as real as running on a thread pool is, at least) which can access shared memory. This is where compareAndSwap and stuff comes in.

payonel:
you already have shared memory, can do this with a custom table. `_G` is shared between processes and threads (event handlers and what not) already in openos. Thus you can

```
_G.shared_memory = {}

event.listen("key_down", function()
  if _G.shared_memory.halt then
    computer.shutdown()
  end
end)

-- then elsewhere, in another program
_G.shared_memory.halt = true

```

As for compareAndSwap and atomicAddInt, and CRITICAL_SECTION for that matter, these things have no purpose, no value, in a non-preemptive environment (where threads are purely cooperative). Which is what we have in OC with our lua archs

Otherwise, thanks for the suggerstion

xarses:
slounds like a doc'er for #2686

SoniEx2:
OC should be able to load a lua file/string into a new preemptive (or rather concurrent) thread.

skyem123:
Actually... a block of memory with the ability to add and set bytes without taking up overhead would actually be useful, as a normal table to store strings or numbers has a lot of overhead, and a single string cannot be modified in place. While not standard to Lua, it would be helpful for stuff like emulators, and I'm sure there are other uses, too...

purplemonday:
** note
so long you didnt specify it as local and in front of the thread.create it can already be used by other threads, there's no need to use _G namespace for this.
this also counts for coroutines etc
Only need for _G is when you need to expose it to literately everything to an even lower level bit of code and has such its risks in weirdness using it as the preferd method.

DCNick3:
I agree with @skyem123. Of course you can write library, that will provide library for effective buffer manipulation by storing memory in strings, dividing it where it is accessed (break it down to small parts) and merging parts, that are not accessed, but it will still have enormous overhead, so this kind of feature will be very useful, In fact I'm working on some sort of bytecode interpreter and having this feature will reduce it's code size and complexity and increase speed.

skyem123:
In some ways this is a Lua feature request 😛

[...]

PR #2744, opened 2018-02-10
Add mutable strings - constant size byte arrays

DCNick3:
As discussed in #2693

### DIFF
--- a/src/main/scala/li/cil/oc/server/machine/StringBufferValue.scala
+++ b/src/main/scala/li/cil/oc/server/machine/StringBufferValue.scala
@@ -1,0 +1,100 @@
+package li.cil.oc.server.machine
+
+import li.cil.oc.{OpenComputers, Settings}
+import li.cil.oc.api.machine.{Arguments, Callback, Context}
+import li.cil.oc.api.prefab.AbstractValue
+import li.cil.oc.server.machine.luac.NativeLuaArchitecture
+import net.minecraft.nbt.NBTTagCompound
+
+
+class StringBufferValue() extends AbstractValue {
+  /* TODO: fix koefficients */
+  final val limitBase = 10000
+  final val readCost = 0.1
+  final val writeCost = 0.2
+
+  var buffer: Array[Byte] = _
+  def this(size: Integer) = {
+    this()
+    this.buffer = new Array[Byte](size)
+  }
+  def result(args: Any*): Array[AnyRef] = li.cil.oc.util.ResultWrapper.result(args: _*)
+
+  @Callback(direct = true, limit = limitBase)
+  def write(ctx: Context, arg: Arguments): Array[AnyRef] = {
+    ctx.consumeCallBudget(writeCost * buffer.length / limitBase)
+    val offset = arg.checkDouble(0).toInt
+    val value = arg.checkByteArray(1)
+    if (offset < 0) {
+      throw new IllegalArgumentException("offset must be positive")
+    }
+    for (i <- offset until math.min(offset + value.length, buffer.length)) {
+      buffer(i) = value(i - offset)
+    }
+    null
+  }
+
+  @Callback(direct = true, limit = limitBase)
+  def read(ctx: Context,arg: Arguments): Array[AnyRef] = {
+    ctx.consumeCallBudget(readCost * buffer.length / limitBase)
+    val offset = arg.checkInteger(0)
+    var sz = 1
+    if (arg.count() >= 2) {
+      sz = arg.checkInteger(1)
+      if (sz < 0) {
+        throw new IllegalArgumentException("size must be positive")
+      }
+    }
+    if (offset < 0) {
+      throw new IllegalArgumentException("offset must be positive")
+    }
+    val res = new Array[Byte](math.max(math.min(sz, buffer.length - offset), 0))
+    for (i <- offset until math.min(offset + sz, buffer.length)) {
+      res(i - offset) = buffer(i)
+    }
+    result(res)
+  }
+
+  @Callback(direct = true, limit = limitBase * 10)
+  def length(ctx: Context, arg: Arguments): Array[AnyRef] = {
+    result(buffer.length)
+  }
+
+  @Callback(direct = true, limit = limitBase)
+  def tostring(ctx: Context, arg: Arguments): Array[AnyRef] = {
+    ctx.consumeCallBudget(readCost * buffer.length / limitBase)
+    result(buffer)
+  }
+
+  override def load(nbt: NBTTagCompound): Unit = {
+    super.load(nbt)
+    buffer = nbt.getByteArray("buffer")
+  }
+
+  override def save(nbt: NBTTagCompound): Unit = {
+    super.save(nbt)
+    nbt.setByteArray("buffer", buffer)
+  }
+
+  override def dispose(context: Context): Unit = {
+    if (Settings.get.limitMemory) {
+      context match {
+        case m: Machine =>
+          m.architecture match {
+            case arch: NativeLuaArchitecture =>
+              val lua = arch.lua
+
+              // Give memory back to lua code
+              arch.unmanagedMemory -= buffer.length
+              lua.setTotalMemory(lua.getTotalMemory + (buffer.length * arch.ramScale).toInt)
+            case _ =>
+            // Do nothing here, because LuaJ does not even care about memory
+          }
+
+        case _ =>
+          // Can this ever happen?
+          OpenComputers.log.warn("StringBuffer is disposed in unknown context; this will cause memory leak")
+      }
+    }
+  }
+}
--- a/src/main/resources/assets/opencomputers/lua/machine.lua
+++ b/src/main/resources/assets/opencomputers/lua/machine.lua
@@ -1400,6 +1400,15 @@ local libunicode = {
 sandbox.unicode = libunicode
 
 -------------------------------------------------------------------------------
+local libstrbuf = {
+  new = function(size)
+    checkArg(1, size, "number")
+    return wrapSingleUserdata(spcall(strbuf.new, size))
+  end
+}
+sandbox.strbuf = libstrbuf
+
+-------------------------------------------------------------------------------
 
 local function bootstrap()
   local eeprom = libcomponent.list("eeprom")()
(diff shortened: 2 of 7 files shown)


src/main/scala/li/cil/oc/server/machine/StringBufferValue.scala
```diff
@@ -0,0 +1,100 @@
...
+import li.cil.oc.api.prefab.AbstractValue
+import li.cil.oc.server.machine.luac.NativeLuaArchitecture
+import net.minecraft.nbt.NBTTagCompound
+
+
+class StringBufferValue() extends AbstractValue {
+  /* TODO: fix koefficients */
+  final val limitBase = 10000
+  final val readCost = 0.1
+  final val writeCost = 0.2
```
DCNick3, line 14: Please notice, I'm not sure what is best values for this

SoniEx2:
I'm surprised you can't create a full userdata with the array using the API.

DCNick3:
What did you mean?

SoniEx2:
Allocate native memory buffer as a Lua "full userdata".

Have you ever used Lua C API?

DCNick3:
Well, a bit. Never created userdata before.

DCNick3:
Anyway, looks like `LuaState` (Java API) don't have any methods to work with full userdata

src/main/scala/li/cil/oc/server/machine/StringBufferValue.scala
skyem123: "koefficients", what does that mean? It's not a valid english word. :p
From what I understand, there shouldn't be a performance penalty for this, Lua already has an operation limit, and memory operations should be fast.
Memory size limit is harder, theoretically it should be limited by the memory installed in the OC computer?

skyem123:
Hmm... could you allow a user to expand and shrink the buffer dynamically? Would make it more flexiable... also will there be any eqivalent to `string.pack` for this? I know that it already exists for Lua 5.3, but OC is still stuck with Lua 5.2 by default, and also wouldn't it be better if there was less copying of strings (or does that not matter?)

(hopefully my rambling and nitpicks aren't annoying! ^^;)

DCNick3:
`string.pack` can be implemented efficiently in userland and I think that it should be done like this. About resize - I'll (probably) implement it later.

skyem123:
Hmm... @payonel does userland stuff in OpenOS...

SoniEx2:
If it can be done efficiently in userland then how come Lua 5.3 adds it in C?

src/main/scala/li/cil/oc/server/machine/StringBufferValue.scala
DCNick3: Oh, sorry, that should be 'factor' or something like this.
Memory size limit is already bound to computer RAM, and when buffer is allocated max lua memory is being substracted by allocated memory size, so there's no need for additional setup for that.
And yes, this setups costs (limits) for buffer I/O. Keeping in mind that it should be fast it now consumes `(opCost * size + 1) / limitBase` of call budget per call to read/write/tostring and `1 / (limitBase * 10)` for length operation, where call budget is `1.0` for Tier 2 computer; `opCost` is `readCost` or `writeCost`; `size` is requested (not actual) operation size. But I think that this should be tuned...

SoniEx2: I guess I should use tables unless I need very big buffers?

DCNick3:
I think there's 2 points:
- Extending standard library - that's cool, because you get generic interface for this kind of stuff without need to implement it
- It's more efficient in C lua without `machine.lua` stuff, but If I write it as userdata it will end up being slow - because wrapping and unwrapping arguments consumes time by creating tables and checking types this stuff (even without actual wrapping)

src/main/scala/li/cil/oc/server/machine/StringBufferValue.scala
DCNick3: Yes, you won't win that much If you'll use it on small buffers

SoniEx2: You mean I'll always get a negative performance "boost" at a higher data density? Why can't we have both?

DCNick3: Well, if API would be implemented as 'raw' functions, not userdata callbacks then you can win some cpu time. Lua arrays are O(logN) for read/write and when N is small it's close enough to O(1), so you will lose more on calling userdata callback, unwrapping parametes, then wrapping results, etc. But in fact I wrote this API not for speed, but memory consumption optimization - here you should get improvement even on small arrays.

payonel:
if you're needing to change the memory size, and track the these mutable strings separately as "unmanaged memory" with nbt tags, it really feels like you're going about this wrong

I'm feeling more in favor of this than against it, but, it needs to use lua_newuserdata to allocate the buffer, which will use the lua state to make the allocation request. there should be no calls to recompute or change the memory usage directly

Also, there should be no management of these buffers, using lua_newuserdata will allow the object to exist as part of the lua state, persist (save, load) just like the rest of the lua state.

DCNick3:
So, this will require C code to be written? Because in `li.cil.repack.com.naef.jnlua.LuaState` there's no `lua_newuserdata` import.
Where's C code? A saw some repos like JNLua, is it a place for this?

skyem123:
I just realised, this might also have to work with LuaJ as LuaJ is used as a fallback when normal lua doesn't work for some reason.

DCNick3:
Hmm, after some time spent on fixing ancient build scripts I was able to build `OC-Natives`, but now I found out, that It contains only lua 5.2
Is it actual repo for natives library?

DCNick3:
@skyem123 Code that I wrote works on LuaJ and I think It'll be ok to use it when natives are not available

DCNick3:
Just found out, that `OC-Natives` have `master-Lua5.3` branch, will stick with that for now.

DCNick3:
Well, after some time spent thinking I found two problems with C implementation.
- 64-bit systems will have invalid string-buffer memory usage, because memory available to 64-bit lua is multiplied by 1.8, while array sizes stay the same. It means that while on 32-bit host you can allocate only 1 MiB buffer, on 64-bit host on same OC computer you can allocate 1.8 MiB
- I can provide a build for linux and windows, but really dunno how to deal with macos

DCNick3:
Now I have (partially) working C version of this module. But still, in 2 MiB computer you can allocate 3 MiB block. Is it ok?

SoniEx2:
Get a 32-bit OS/java.

SoniEx2:
Yeah we should really split the RAM size tables...

Altho every time I brought it up it got rejected/ignored.

(`("a"):rep(3*1024*1024)`)

DCNick3:
Okay, if community is OK with it - it's OK. After researching JNLua C code I found way to fix it by splitting memory scale per by lua type memory is being allocated for (this is available in lua api), but it will somewhat reduce performance (it will be needed to store allocation list/bin-search tree). I think as for now I will deal with it.

But another problem. What do I do with MacOS? I haven't any Apple devices, and gradle isn't designed to do cross-compilation. Maybe use another build system (like GNU make or cmake) for native library?

SoniEx2:
Just don't worry about it for now. We can change it later.

payonel:
I like this idea because having the custom userdata for a raw char array is one of the basic suggested concepts in the lua documentation, basically it is the default example of what to do with lua_newuserdata

I also like the idea of a cleaner array than using tables

but:

1. I would only accept this feature via lua_newuserdata
2. we will not accept updates to oc natives for this feature's sake
3. there is a work around, use unmanaged drives

So, I do apologize for all the work you've put into this, but this is the conclusion I've come to after giving this much thought. Thank you

skyem123:
The thing is... this is a convinience feature... while neat, unmanaged drives are not convenient. 
Unmanaged drives also have limitations, such as being very slow? with the call limits, which makes them not useful for emulators.

I don't see anything wrong with the work-around, as long as it's possible to reimplement it as userdata without breaking programs that use it. In fact, I think it would be neat for things to be able to steal memory from a system (maybe that could be a new limitation of APUs?).

I understand you not wanting to have a messy hack, but it's a shame it will kill a nice feature like this. :/

DCNick3:
That's sad. But.. Why not update natives? Is it really so problematic?

payonel:
using unmanaged drives could be prohibitively slow if you are trying to run emulations, yes. Reading a 4MB (T3) drive takes ~6.65s, and writing the entire drive takes ~11.95s.

but, the general guideline for OC is that we don't add features for speed unless we have an expected user program that cannot work correctly due to performance.

Additionally, changing the oc-natives would require a necessary bugfix or feature.

skyem123:
There is an emulator written in Lua for OC that was stopped because it was impossible to emulate memory efficiencly. I could try to find it and port it to this if you want a concerte example of what this would be useful for?

The point isn't to make things faster, it's to make cool but impractical things more practical. It's also worth pointing out that some types of computers cannot have both an unmanaged drive and an OS drive, and to me at least, it doesn't make sense in "lore"/"realism" (I admit that they are vague and silly terms especially for a Minecraft mod). 

You also haven't explained why the concept of being able to "steal" memory from the computer, using Scala code is a bad thing, which has left me confused.

Finally, would you be okay with a PR to the natives to add this feature, staying around and only merged when you need to update it anyway?


EDIT: removed random letters because touchscreen keyboards are bad

SoniEx2:
You should have like ByteBuffer API using lua_newuserdata in natives.

DCNick3:
@SoniEx2 If you meant that this stuff should be written in C as part of natives library - I already have done that. See [here](https://gist.github.com/DCNick3/d57e509d4196db55cc940bfba9f64e34).
Or did you mean, that one should be able to allocate any ByteBuffer trough JNLua api?

SoniEx2:
The latter.

DCNick3:
Thought about this. What else (besides strbuf) would require such a feature? (I wasn't able to imagine such thing)

SoniEx2:
You should also allocate the data in the userdata rather than using the alloc function directly, I think. There's some GC stuff that happens there IIRC...

DCNick3:
What's the problem with this? I examined lua allocation chain.. There's nothing special - like in `lua_realloc` function but with telling GC about new object (I do not, because freeing is done with `__gc` callback). Also If no `lua_Alloc` returned NULL, lua tries to collect garbage (like I do) and try again. If it still fails - `not enough memory` error is being thrown.

SoniEx2:
What's wrong with avoiding an unnecessary allocation and __gc (which isn't guaranteed to be called)?

DCNick3:
Ah, forgot to mention that. It's to be able to do `resize` (lua does not give API for userdata resize). Also, lua documentation states, that if object is marked for finalization it WILL be finalized (worst case - in lua_close call). Also, JNLua relies on that method - if it won't be called, that Java Object pushed to lua stack (and eventually became garbage) won't be freed.

PR #2755, opened 2018-02-20
[1.10] Make cable block highlight hug the block

SquidDev:
This allows you to reach behind a cable, as the highlight box taking up the whole block. I'm happy to add a 1.7.10 version, but I'd like to check that people are OK with how everything looks first. I'm also open to ways the code could be improved - I've tried to make the direction handling as generic as possible, but it's not perfect.

One thing I'm not entirely sure about is black notch on the end of cables. I've currently excluded it from the bounding box (as that is what was done before), but it may be better to include it.

### Screenshots
A basic cable intersection A flat cable part The cable 'core'

### DIFF
--- a/src/main/scala/li/cil/oc/client/renderer/HighlightRenderer.scala
+++ b/src/main/scala/li/cil/oc/client/renderer/HighlightRenderer.scala
@@ -1,23 +1,14 @@
 package li.cil.oc.client.renderer
 
-import li.cil.oc.Constants
-import li.cil.oc.Settings
-import li.cil.oc.api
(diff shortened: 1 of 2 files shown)


skyem123:
Personally I feel that it should include the end of the cables, even if it didn't before, because the concept of the visual stuff and the selection box not matching is a little confusing to me. but that's just my opinion. _shrugs_

payonel:
i merged this outside of github 3374cf45e8bf49adc0f1e3f78e5ca0f4ab5f297e

SquidDev:
Thanks!

PR #2762, opened 2018-02-27
Changed filesystem and tile data (Lua state etc.) saving to occur in separate threads.

Vexatos:
Until now, OpenComputers has been saving all computer states, hologram data, and filesystems on the server thread. This change makes all of this be scheduled on a separate thead. On servers with a lot of computers, this has a noticable (up to orders of magnitude) speed improvement. I had to get rid of the cache in `SaveHandler` for best concurrency, so there are more reads and writes happening than there used to be, but on modern operating systems that cache files in RAM, this shouldn't have much of an impact.

Another minor change: directories in filesystems now only change their modification date if any file inside of them actually changed.


payonel:
I merged and tested your changes locally, and pushed myself

PR #2802, opened 2018-04-02
Support for setting private frequencies in EnderStorage (closes #2775).

### LINKED ISSUE #2775: [Suggestion] Allow frequency change of EnderStorage private ender chest / tank
In public servers, i think we're nearly forced to used private ender chests, but computers doesn't allow to change their frequencies. So it would be good to add an options in the config to allow computers to change private ender chests frequencies.

amesgen:
This small PR addresses issue #2775. When trying to change the frequency of an Ender tank/chest, the following rules apply:

1. if the device is not owned: *allow*
2. if the config option `allowModifyingPrivateFrequencies` is set to `2`: *allow*
3. if the config option is set to `1` (default option) and the computer has a single user equal to the owner of the device: *allow*
4. *deny*

I am looking forward to feedback.


payonel:
Thanks for looking into this. I'll be glad to test this when I have some more mod dev time. Personally I have no issue allowing private frequency control without any configuration. These aren't protected chests in that meaning of "private", they are just separate from the universal frequencies. A player concerned about privacy would already need to safeguard the physical chest.

payonel:
we're brought back in this support, thank you for your work here.

PR #2803, opened 2018-04-02
Support for setting private frequencies in EnderStorage (closes #2775).

### LINKED ISSUE #2775: [Suggestion] Allow frequency change of EnderStorage private ender chest / tank

amesgen:
see #2802 

This PR should also work in 1.11 and 1.12 (EnderStorage support has to be readded for these versions, which should be copy-paste mostly).


payonel:
@amesgen thank you for this work. I brought back the enderstorage work and then compared it to your work here to make sure we got it right.

PR #2836, opened 2018-05-18
Proper .prop file recognition

wrothmonk:
Install currently only checks .prop files at the root of a device instead of the actual path where the installation is coming from. This can cause issues specifically when using the --from option if it is targeted at anything other than the root of a filesystem. This minor change fixes that.


payonel:
oooo i like this! thank you! I'll just need to test it. I'll mark it for 1.7.3

PR #2839, opened 2018-05-19
Redo zh_CN Manual(Partly)

tongkl1:
have redone Blocks & General folders. I don't have much time so it's time for someone else to take on the Item folder. Or we'll still use that c**p broken translation.


tongkl1:
Well I'll try to fix it & finish the rest part of OC lolol

payonel:
"All devices can be programmed using Lua 5.3"
we support 5.2 and 5.3, not just 5.3

tongkl1:
emmm I just thought oc updated to support lua 5.3 and we should update the manual as well. ocdoc.cil.li has also changed 5.2 into 5.3 lol

asiekierka:
OpenComputers supports both Lua 5.2 and Lua 5.3. The latter is now the default, but it is not exclusive.

PR #2889, opened 2018-07-10
Fixes lightning issues with ShaderMod

### LINKED ISSUE #2486: Compatibility with Shader Mod
You read the title. This isn't a very important thing as it doesn't break anything.
When you got any shadow manipulating shader pack (every shader pack ever) the screen goes so dim it just barely can be seen or if any, it's like this even in a completely dark room. I was using [KUDA Shaders](http://shadersmods.com/kuda-shaders-mod/#comment-6317) for this particular screenshot:

oc shaders

paulhobbel:
The added line fixes lightning issues with shaders. The monitors are barely readable without this line. There is one issue with this fix and that is that at night the monitors will have a reddish tint.
Below I will show some pictures from before and after the change:

Before day:

Before night:

After day:

After night:

### DIFF
--- a/src/main/scala/li/cil/oc/client/renderer/tileentity/ScreenRenderer.scala
+++ b/src/main/scala/li/cil/oc/client/renderer/tileentity/ScreenRenderer.scala
@@ -9,6 +9,7 @@ import li.cil.oc.integration.util.Wrench
 import li.cil.oc.util.RenderState
 import net.minecraft.client.Minecraft
 import net.minecraft.client.renderer.GlStateManager
+import net.minecraft.client.renderer.OpenGlHelper
 import net.minecraft.client.renderer.Tessellator
 import net.minecraft.client.renderer.tileentity.TileEntitySpecialRenderer
 import net.minecraft.client.renderer.vertex.DefaultVertexFormats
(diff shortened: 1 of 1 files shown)


asiekierka:
Please don't put this in too haphazardly before it's checked for any ways in which it might be affecting non-ShadersMod environments.

paulhobbel:
Yea that might be a good idea, I have no clue if it will effect other environments in any way, I do know however that this fix has worked in CC and did not cause any known issues "yet"

SaphireLattice:
What about putting torches close to the computer? Or glowstone? Something that causes the light value to change at places (or is the monitor max light?). As this is a lightmap, it would be good to check how it interacts with light sources of vanilla and so.

paulhobbel:
Without the line, screen stays the same with any light source (torches, glowstone, you name it), with the added line screen stays the same as seen in the pictures above. The light sources just add extra light to the surroundings of the monitor.
In response to @asiekierka I've been testing without the shadermod today, and haven't noticed any side effects as of yet, screens still look the same with the extra line.

halvors:
Merge this please :) Awesome pull :D

Garagoth:
Do you need any extra testing of this one? How could I help with this one?
Currently with shaders mod I can see the screen fine only if I am _holding_ a light source (torch, glowstone etc) in a hand and only when I am close. Else - all the letters are just almost invisible.

PR #2893, opened 2018-07-16
Turkish Localization

utkumaden:
I wrote the Turkish localization for the mod. I also added a Turkish robot to the robots list. Please feel free to roll the commit back if you only allow internationally famous robots.

I would have loved to translate the manual as well, but I lack the time to do so. This took me hours to complete and I hope Turkish users who do not know English can have an easier time playing with this mod.


asiekierka:
Many of the references are already somewhat obscure. I'm all for the Turkish robot.

utkumaden:
The name is 216, but I wrote it as "İkiyüzonaltı" because someone on discord suggested that it might confuse newcomers into thinking it is an ID.

Edit: Please comment on my decision.

PR #2916, opened 2018-08-07
Add lua 5.2 native library for Solaris i386

Low-power:
This native library is built from https://github.com/MightyPirates/OC-Natives on a Solaris 11 operating system.
BTW the native libraries are named so confusing, a refactoring is recommended.


asiekierka:
Solaris is not and will not be officially supported. We have enough trouble building Windows, Linux and Mac natives. A way to provide custom natives will be added in the future.

PR #2927, opened 2018-08-21
Adding Cut/Uncut operations to OpenOS editor

AntiBlueQuirk:
This patch adds a simple cut buffer to the editor, allowing you to cut whole lines to the buffer, and paste them out elsewhere. This feature was inspired by a similar feature in GNU nano (see [The Cutbuffer](https://www.nano-editor.org/dist/v2.9/nano.html#Editor-Basics)). It's not a full fledged clipboard, but provides simple, line-based cut and paste. A savvy user can get quite a lot of use out of this tool, as it makes removing, duplicating, and rearranging blocks of code much easier.

The basic operation is as follows:

- Pressing Control-K removes the current line and puts it in the cut buffer. Multiple consecutive lines can be copied this way. Moving the cursor to a new line ends the cut, meaning the next cut will start a new buffer.
- Pressing Control-U inserts the lines in the cut buffer before the current line. This doesn't clear the cut buffer, so the user can cut a line (or lines), and then uncut it many times.
- The cut buffer does not interact with the user's clipboard. It is a separate buffer. The cut buffer does not persist if edit is closed.
- The number of lines in the cut buffer is displayed next to the cursor position.

One problem I noticed is that if the user's `/etc/edit.cfg` file does not contain a binding for the new actions, they do not appear and are unusable. Simply deleting the config file so it can be regenerated fixes this, but it may not be obvious to users after an update. It may be worthwhile to use the default bindings when they are not present in the config file, but I considered that outside the scope of this patch.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/edit.lua
@@ -59,7 +59,9 @@ local function loadConfig()
     save = {{"control", "s"}},
     close = {{"control", "w"}},
     find = {{"control", "f"}},
-    findnext = {{"control", "g"}, {"control", "n"}, {"f3"}}
+    findnext = {{"control", "g"}, {"control", "n"}, {"f3"}},
+    cut = {{"control", "k"}},
+    uncut = {{"control", "u"}}
   }
   -- Generate config file if it didn't exist.
   if not config then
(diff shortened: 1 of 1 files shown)


AntiBlueQuirk:
I should clarify that the problem with `/etc/edit.cfg` doesn't occur on a fresh install of OpenOS. It's just when the user has the new version of edit, but the config doesn't have the new bindings.

Vexatos:
This PR should target the master-MC1.7.10 branch. Could you rebase it?

AntiBlueQuirk:
Alright, I think I got it rebased correctly. I'm not sure why it's showing 7b77dba as part of the request now. I didn't touch that commit. EDIT: fixed.

Vexatos:
I'd just like to mention how rare it is to see someone who is actually capable of properly rebasing a PR on GitHub. This stuff is unfortunately rare. Thank you for saving me some time.

AntiBlueQuirk:
I'm not gonna lie, I had to look some stuff up. :D

payonel:
tested this, looks good, works well.
Thanks

PR #2933, opened 2018-08-25
Improved chunkloader upgrade

### LINKED ISSUE #2499: Chunkloader
Right now it only works in drones and robots.
Could you add something like a chunkloader card to use in the normal computers?

### LINKED ISSUE #2768: Chunkloader upgrade dimension blacklist
I'd like to have a setting to prevent the upgrade from loading chunks in the dimensions I specify.

svitoos:
# Improved chunkloader upgrade
This PR adds a lot of features, most of which give server admins more control over the behavior of the chunkloader upgrade.
Namely:

- An option to disable a chunkloader if its owner goes offline.
- A whitelist and a blacklist of dimensions in which the update is allowed to load chunks.
  Closes #2768.
- An option to use the `forgeChunkLoading.cfg`'s settings to limit number of tickets per player.
- A command that lists all chunkloaders in the world (`/oc_cl`).
- More verbose logging (the logging level is configurable).
- If a chunkloader is duplicated in a way so that multiple items share the same address,
  only one of them is allowed to work at a time.
- In case a drone with the chunkloader upgrade is teleported to another dimension, it requests a new ticket.
- Allows adapters and microcontrollers to have the chunkloader upgrade.
  Closes #2499.

A new configuration group, `chunkloader`, is added with the following settings:

- `requireOnline`. If `true`, the chunkloader only loads chunks if its owner is online.
  If the owner goes offline, the chunks the upgrade kept loaded are unloaded;
  the upgrade keeps consuming energy, though, and `isActive()` returns `true`.
  If the option is enabled, tickets are registered for the player, not for the mod.
- `dimBlacklist`. The chunkloader does not load chunks in dimensions whose IDs are listed there.
- `dimWhitelist`. This setting is the opposite of `dimBlacklist`, and lists dimensions in which the chunkloader is allowed to work.
  Ignored if left empty (so the upgrade works in all dimensions not listed in the blacklist).
  If the same entry appears in both lists, the blacklist takes precedence.
  `setActive()` returns `false` if the chunkloader is blocked in that dimension.
- `playerTicket`. If `true`, tickets are assigned to the player who owns the chunkloader rather than the mod,
  so the `forgeChunkLoading.cfg`'s limits get applied for that player.
  Makes it impossible to activate a chunkloader not owned by a player.
- `logLevel`. Controls the level of logging: (`0` for quiet; `1` for more verbose; `2` for debug).

A new command, `/oc_chunkloaders` (`/oc_cl`), is added, which lists chunkloaders in the loaded chunks,
and inactive registered tickets, displaying the following information:

- the component address
- the chunkloader status
  - `inactive` (the chunkloader is inactive)
  - `active` (the chunkloader is enabled, and loads chunks)
  - `active/suspend` (the chunkloader is enabled, but doesn't load chunks)
- the block coordinates
- the chunk coordinates
- the dimension ID
- the owner

### DIFF
--- a/src/main/scala/li/cil/oc/common/tileentity/Microcontroller.scala
+++ b/src/main/scala/li/cil/oc/common/tileentity/Microcontroller.scala
@@ -168,12 +168,12 @@ class Microcontroller extends traits.PowerAcceptor with traits.Hub with traits.C
     if (node == plug.node) {
       api.Network.joinNewNetwork(machine.node)
       machine.node.connect(snooperNode)
+      connectComponents()
     }
     if (plug.isPrimary)
       plug.node.connect(componentNodes(plug.side.ordinal()))
     else
       componentNodes(plug.side.ordinal).remove()
-    connectComponents()
   }
 
   override protected def onPlugDisconnect(plug: Plug, node: Node) {
@@ -182,6 +182,8 @@ class Microcontroller extends traits.PowerAcceptor with traits.Hub with traits.C
       plug.node.connect(componentNodes(plug.side.ordinal()))
     else
       componentNodes(plug.side.ordinal).remove()
+    if (node == plug.node)
+      disconnectComponents()
   }
 
   override protected def onPlugMessage(plug: Plug, message: Message): Unit = {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1563,4 +1563,31 @@ opencomputers {
     # the native C locale).
     disableLocaleChanging: false
   }
+
+  # Chunkloader Upgrade settings.
+  chunkloader {
+    # If `true`, the chunkloader only loads chunks if its owner is online.
+    # If the owner goes offline, the chunks the upgrade kept loaded are unloaded;
+    # the upgrade keeps consuming energy, though, and `isActive()` returns `true`.
+    requireOnline: false
+
+    # If `true`, tickets are assigned to the player who owns the chunkloader rather
+    # than the mod, so the `forgeChunkLoading.cfg`'s limits get applied for that
+    # player. Makes it impossible to activate a chunkloader not owned by a player.
+    playerTickets: false
+
+    # This is a list of blacklisted dimensions. Chunkloaders may only be activated
+    # in dimensions that aren't present in this list.
+    dimBlacklist: []
+
+    # This is a list of whitelisted dimensions. Chunkloaders may only be activated
+    # in dimensions that are present in this list. If this list is empty,
+    # chunkloaders may be activated to all dimensions not blacklisted. Note that the
+    # blacklist is always applied, so if an entry is present in both the
+    # whitelist and the blacklist, the blacklist will win.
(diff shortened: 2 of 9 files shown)


Vexatos:
Not sure I like allowing the chunkloader in an Adapter. With an MCU, it at least has to be assembled into the thing.

Vexatos:
Doesn't this also imply that computers without an owning player can no longer use chunkloaders if `requireOnline` is `true`?

src/main/scala/li/cil/oc/common/tileentity/Microcontroller.scala
```diff
@@ -168,12 +168,12 @@ class Microcontroller extends traits.PowerAcceptor with traits.Hub with traits.C
     if (node == plug.node) {
       api.Network.joinNewNetwork(machine.node)
       machine.node.connect(snooperNode)
+      connectComponents()
     }
     if (plug.isPrimary)
       plug.node.connect(componentNodes(plug.side.ordinal()))
     else
       componentNodes(plug.side.ordinal).remove()
-    connectComponents()
```
Vexatos, line 176: What is this change for?

asiekierka:
@Vexatos I agree. MCUs should be available, whereas Adapters should be opt-in server-side?

>Doesn't this also imply that computers without an owning player can no longer use chunkloaders if requireOnline is true?

Would it be a good idea to make chunkloaders themselves owner-able?

src/main/scala/li/cil/oc/common/tileentity/Microcontroller.scala
svitoos: Maybe I'm wrong but the method is called several times uselessly.

svitoos:
The ability to install in the adapter and the microcontroller is not the main feature and can be removed.

If requireOnline is true then chunkloaders work only in robots and drones.

Sorry for my English.

svitoos:
> Would it be a good idea to make chunkloaders themselves owner-able?

probably yes

payonel:
I've decided to cherry pick a tiny piece of this work, and close the rest of this PR

I removed a lot of the code for two reasons:
1. I don't want player ownership of the chunkloaders. I would prefer, at most, only have a record of what player activated the chunkloader.
2. I didn't want to work with too much new code for chunkloading that I haven't had the time to test for this next patch release. I was reading a lot of new code that loads chunks and it looked to make a bigger change for our next patch release than I wanted.

Some of the features removed provided some nice journaling of how the server is being used. Those things could be brought back in a new PR, intended for OC 1.8 instead of this next patch release.

If you are planning to resubmit this work, consider the 2 parts of the PR -- that I would like them as separate PRs

1. correctness: improvements that made the chunk loading work better and more reliably, but without any difference in management or control
2. feature and code refactoring: logging, chat commands, refactoring ticket management

payonel:
oh one last point, the adapter is for inter-mod interops only, not for loading our own upgrades in a block. that's what microcontrollers should do

PR #2942, opened 2018-09-11
thaumcraft compatibility improvement

seebs:
Item stacks of Thaumcraft items, such as esssentia jars, are a
pain to work with if you can't see their contents. The contents
are not hidden by Thaumcraft, so showing them isn't cheaty.

Making these available makes it possible to use OC to sort
essentia jars by contents, which is SUPER USEFUL since everything
else I have which sorts inventory ignores them.

So we display the two nbt tags corresponding to immediately-visible
essentia things: aspects (kind and quantity) and the "filter", which is
what jar labels do.

### DIFF
--- a/src/main/scala/li/cil/oc/integration/minecraft/ConverterItemStack.scala
+++ b/src/main/scala/li/cil/oc/integration/minecraft/ConverterItemStack.scala
@@ -11,6 +11,8 @@ import li.cil.oc.util.ItemUtils
 import net.minecraft.item
 import net.minecraft.item.Item
 import net.minecraft.nbt.NBTTagString
+import net.minecraft.nbt.NBTTagCompound
+import net.minecraft.nbt.NBTTagList
 import net.minecraftforge.common.util.Constants.NBT
 import net.minecraftforge.oredict.OreDictionary
 
@@ -30,8 +32,35 @@ object ConverterItemStack extends api.driver.Converter {
         output += "size" -> Int.box(stack.getCount)
         output += "maxSize" -> Int.box(stack.getMaxStackSize)
         output += "hasTag" -> Boolean.box(stack.hasTagCompound)
-        output += "name" -> Item.REGISTRY.getNameForObject(stack.getItem)
+	val name = Item.REGISTRY.getNameForObject(stack.getItem)
+        output += "name" -> name
         output += "label" -> stack.getDisplayName
+	val nameString = name.toString()
+
+	// Handle essentia/vis contents for Thaumcraft jars, phials, and crystals
+	if ((nameString == "thaumcraft:jar_normal") ||
+	    (nameString == "thaumcraft:jar_void") ||
+	    (nameString == "thaumcraft:phial") ||
+	    (nameString == "thaumcraft:crystal_essence")) {
+	    if (stack.hasTagCompound &&
+	      stack.getTagCompound.hasKey("Aspects", NBT.TAG_LIST)) {
+	      val aspects = mutable.ArrayBuffer.empty[mutable.Map[String, Any]]
+              val nbtAspects = stack.getTagCompound.getTagList("Aspects", NBT.TAG_COMPOUND).map((tag: NBTTagCompound) => tag)
+	      for (nbtAspect <- nbtAspects) {
+	        val key = nbtAspect.getString("key")
+	        val amount = nbtAspect.getInteger("amount")
+                val aspect = mutable.Map[String, Any](
+		        "aspect" -> key,
+		        "amount" -> amount
+	        )
+	        aspects += aspect
+	      }
+	      output += "aspects" -> aspects
+	    }
+	    if (stack.hasTagCompound && stack.getTagCompound.hasKey("AspectFilter", NBT.TAG_STRING)) {
+	      output += "aspectFilter" -> stack.getTagCompound.getString("AspectFilter")
+	    }
+	}
         if (stack.hasTagCompound &&
           stack.getTagCompound.hasKey("display", NBT.TAG_COMPOUND) &&
           stack.getTagCompound.getCompoundTag("display").hasKey("Lore", NBT.TAG_LIST)) {


payonel:
i'm glad to see item stack conversions like this, to be honest.
but as with any PR you should know that our default branch is 1.7.10
we support 1.7.10, 1.10, 1.11, and 1.12
Some things we merge higher up, particularly when it doesn't have context for 1.7.10
Porting this to 1.7.10 might be simple, but it also might not make sense if this metadata was different/didn't exist previously
However, that's something I would need to know before we can merge this. Ideally, you would make a PR for 1.7.10, and at any point from 1.7.10 up to 1.12 where the merge doesn't "just work", you'd need to specify that in the PR (e.g. "this works from 1.7.10 to 1.11, but 1.12 needs a code change), and each code change step could be a separate PR (unless you want to include instructions in the PR on how to resolve the merges)

Secondly, this PR uses a mix of tabs and spaces. Please use space and follow the style of the file you are editing.

Thanks

seebs:
Unfortunately, it's completely different in 1.7.x, if memory serves, and I don't have access to that anymore to test things. But I believe the new system is new to 1.12, and isn't even how it worked in the few betas released for 1.10.

On the other hand, so far as I know, this won't *hurt* anything by existing in previous editions, it just won't produce interesting results. (And I think in 1.7, it may have been possible to get the desired data some other way; the shift to moving everything into nbt tags was new with thaumcraft 6, as part of prepping for 1.12/1.13 in some way I don't understand.)

It's beyond my ability do to much better with this -- I don't actually know scala, for one thing! I just figured it'd be possibly-useful to someone.

payonel:
i don't need the non-nbt solution for previous versions. If you are reasonably confident this nbt flavor started in 1.12, we are good to work in 1.12. but we need to look at every change/fix/update from the 1.7.10 perspective. If not 1.7.10, then 1.10. If not 1.10, then 1.11. And only is very special cases, just for 1.12. Is this a special case? I dont know, I haven't reviewed thaumcraft meta data. I leave that to you to confirm.

I believe thaumcraft skipped 1.11, i understand that is probably why you only mentioned 1.10

the code style/alignment still needs to be corrected before we would merge this

thanks

seebs:
Thaumcraft 6 up through Beta 8 was released for 1.10. I have no idea what TC5 was like, or which version it ran on, but I think it may have been 1.8?

I don't understand the interactions of mod versions and branches of OC. If the branches are generally similar, I would assume this would apply fine to 1.7, I don't know whether or not it would do anything with Thaumcraft 4.x, though. I don't think it would break anything.

I can try to fix up the indentation, sorry about that. I'm less confident that I can perceive other "style" things, because I just don't know Scala well enough to know which things are significant in it.

payonel:
There is some misunderstanding in this discussion. Try to read my next response here with a new set of eyes.

1. There is only 1 "OC version", that is "latest". We represent this one OC version towards 4 MC versions. We are never talking about fixing/updating/patching old versions of OC. We are only working on the latest OC version. There is only OC for MC 1.7.10, OC for MC 1.10, OC for MC 1.11, and OC for MC 1.12. But these are all the same "OC version". However, OC for MC 1.7.10 is our target branch, and we merge up from there to the other 3 MC version flavors.

2. If "OC for MC 1.12" needs to be updated for "MOD XYZ", we must first consider if that change is meaningful, FIRST, in "OC for MC 1.7.10". Maybe mod xyz didn't exist in 1.7.10. Maybe mod xyz worked differently in its version for MC 1.7.10. But we must first find the "lowest" mc version to update OC for. There is only 1 OC version, which is driven by our OC for MC 1.7.10. 1.10 drives 1.11, and 1.11 drives 1.12. What if mod xyz had that same oc-interaction-problem in 1.11 but we only fixed it in 1.12?

3. If mod xyz updated in a new special way in 1.12 that has no meaning, or significantly different meaning in an early version of MC, then we do not need to backport that concept in OC to earlier versions. That is on par with implementing another mods feature via OC just to have said feature available in older versions of MC. Thus, if this feature of TC is new in 1.12, it doesn't make sense to back port it to 1.10 (and we skip 1.11 because I assume there is no TC for 1.11, and we don't care about 1.8 because we don't support OC for MC 1.8)

4. I dont think your scala style in general was off, just don't use tabs, and fix the spacing-indentation.

seebs:
Oh, hmm.

Well, hang on. I *do* have a 1.7.10 server. I am less sure of my ability to modify it (or my client-side code for it), but I can at least log in and `/mt hand` a couple of things to get some guess as to whether this would make sense for that. If it would, it would indeed probably be saner to apply it to that-and-up. I'll get back to you on that.

(And sorry about the indentation, my editor is forever scarred by working on bitbake recipes, which are single files containing a mix of tabless python and tabs-not-spaces shell script code, and while I know this is not how computers work, I still believe that something in this gave vim PTSD and it's never been okay since.)

payonel:
sure, but if thaumcraft is using a new nbt system for its metadata in 1.12 (new to 1.12), then we can make this change only for OC for 1.12

seebs:
okay, for reference (in case i don't get to pursue it):
phial of essentia is `Thaumcraft:ItemEssence:1`, which I think means it's ItemEssence, metadata 1, but has the same NBT tags. Jar is `Thaumcraft:BlockJarFilledItem`, also same NBT tags. Wisp essences are `Thaumcraft:ItemWispEssence`, and mana beans are `Thaumcraft:ItemManaBean`. And all of them seem to use the same Aspects tag.

So there's some things (the essence shards one finds in things like earth-infused stone) which are using metadata, but apparently the jars and things are using the same system! So this should be backportable.

That said, I'd recommend that someone with more stable access to older servers double-check and/or test this, since I will have a really hard time doing that.

seebs:
Oh. So it turns out, 1.7.10 already has all the code for this, which is probably why I never looked at it.

So the chances are decent that the 1.7.10 integrations/thaumcraft code would actually work fine in the newer system, and I believe the Thaumcraft API is released already. In which case, it might well Just Work, except I don't know enough about the build environment or system to figure out how to acquire/provide whatever's needed for the thaumcraft API to be available. But that code looks a lot cleaner/more stable than mine, and also more comprehensive.

repo-alt:
I verified that TC4 uses the same NBT structure, and the item names are as specified in the comment above.
Note that TC integration is disabled by default, because it shows aspects of all the blocks, which is incompatible with TC4 aspect discovery process. Your patch though does not give any spoilers, because to have aspects in phials or jars they have to be already known.

payonel:
Thank you for this work, I've used your code but cherry picked it to resolve some changes

PR #2943, opened 2018-09-14
Add new symbols for the Ukranian language

viktor-sava:
Please, add new symbols for the Ukrainian language. Unicode character code U+0404 and U+0454. It "Cyrillic Big Letter Ukrainian Ie" and "Cyrillic Small Letter Ukrainian Ie"


payonel:
You made a PR without any files changed. Was this your intention?
Are you asking that we draw new font glyphs for those two unicode values? Our glyphs are defined in our font.hex, and we need to use a font style that has an open license we are free to use. Є and є would be good additions.

Feel free to open a issue asking for these addtions. Reference a free to use license font of those glyphs if you find one. 
or
Feel free to open a PR with your contribution using a free licence glyphs of those characters. Make sure you reference the font you used as reference

PR #2946, opened 2018-09-16
Fix transferItem checking the wrong side (updated)

### LINKED ISSUE #2806: Transposer transferItem argument checkSlot should check for the opposite side
Here:
https://github.com/MightyPirates/OpenComputers/blob/ac67d8ba37c2b2c65f6eafb9c2bd371dd6f13fea/src/main/scala/li/cil/oc/server/component/traits/InventoryTransfer.scala#L28-L29

The sourceSide and sinkSide in these 2 lines should be followed by `.getOpposite`, just like the line 31 below.
This causes issues when accessing inventories that have different number of slots on opposite sides.

### LINKED ISSUE #2863: Transposer in microcontroller not extracting items from ic2:te above
Given a transposer in a microcontroller, an ic2 nuclear reactor on top and a chest with items north of the microcontroller and the following BIOS:
```lua
-- luacheck: globals component computer checkArg
local down,up,north,south,west,east = 0,1,2,3,4,5  -- luacheck: ignore
local ts = component.proxy(component.list("transposer")())

while true do
  local slots = ts.getAllStacks(up)
  for i = 1, slots.count() do
    local stack = slots()
    ts.transferItem(up, north, stack.size, i)
  end
end
```
Expected behavior: The transposer should move items from the nuclear reactor into the chest.
Actual behavior: Nothing happens

If I change up for down and I put stuff on the reactor bellow, items are moved into the chest as expected. If I swap the microcontroller and chest in both the code and the world items are moved into the chest.

Test Setup:
setup
OC: 1.7.2.170

cyb0124:
Fix #2806

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/traits/InventoryTransfer.scala
+++ b/src/main/scala/li/cil/oc/server/component/traits/InventoryTransfer.scala
@@ -25,8 +25,8 @@ trait InventoryTransfer extends traits.WorldAware with traits.SideRestricted {
         result(Unit, reason)
       case _ =>
         if (args.count > 3) {
-          val sourceSlot = args.checkSlot(InventoryUtils.inventoryAt(sourcePos, sourceSide).getOrElse(throw new IllegalArgumentException("no inventory")), 3)
-          val sinkSlot = args.optSlot(InventoryUtils.inventoryAt(sinkPos, sinkSide).getOrElse(throw new IllegalArgumentException("no inventory")), 4, -1)
+          val sourceSlot = args.checkSlot(InventoryUtils.inventoryAt(sourcePos, sourceSide.getOpposite).getOrElse(throw new IllegalArgumentException("no inventory")), 3)
+          val sinkSlot = args.optSlot(InventoryUtils.inventoryAt(sinkPos, sinkSide.getOpposite).getOrElse(throw new IllegalArgumentException("no inventory")), 4, -1)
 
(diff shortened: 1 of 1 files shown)


Vexatos:
Is 1.12 the lowest version that is affected by this bug? If not, can you rebase against a lower branch?

cyb0124:
Rebased to master-MC1.10
(Please ignore the source branch name since I can't change that in a PR)

Vexatos:
1.7.10 is not affected by this?

cyb0124:
In 1.7.10 InventoryUtils.inventoryAt doesn't take a side argument, so it's not affected by this.

Vexatos (approved):

payonel:
@cyb0124 this looks good, and I will test it. But my concern is
#2806 has no specific bug repro, just a personal analysis of the code. While I understand the logical concern, there is no in-game proof of the issue. Yet in this PR you say this change **may** fix #2863 - which does seem to be this exact issue, and has a repro. Your comments on this PR are just confusing to me. What workflow have you tested this with that is affected?

cyb0124:
I just posted a repro to #2806 .
Actually I realized this is unrelated #2863 as this bug only affects the case when the slot numbers are specified when calling transferItem. However in #2863 the slot numbers are not specified. I removed that "may fix". Sorry for the confusion.

payonel:
tested and confirmed this is the right fix. thank you!

PR #2978, opened 2018-11-05
Add "leashHolder" as a name

josephcsible:
`field_110168_bw` was remapped from `leashedToEntity` to `leashHolder` as of the 20170919 mappings. Add this name so that the dev environments of other mods that use these mappings will work without "There were errors running the class transformer" messages.


payonel:
confirmed, thank you

PR #2992, opened 2018-11-28
Fixed T1 wireless cards receiving wired messages

### LINKED ISSUE #2989: T1 wireless modems receive wired messages
MC 1.10.2 (Forge 12.18.3.2185)
OC 1.7.3.224

T1 wireless modems receive wired messages, even though they can't send them and isWired() returns false.

Related questions:
1. With T2 wireless, both wired and wireless copies of the same message are received. Is it intended? Can it be avoided?
2. Is it possible to send/broadcast just a wireless message with a T2, without the wired copy?
3. Is it intended that wired messages reach robots that are adjacent to the computer? It does not quite make sense to me. Why would a robot automatically connect to the wired network whenever it gets close to, say, a monitor?

TheCodex6824:
Fixes #2989

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/WirelessNetworkCard.scala
+++ b/src/main/scala/li/cil/oc/server/component/WirelessNetworkCard.scala
@@ -156,6 +156,9 @@ object WirelessNetworkCard {
     )
 
     override def getDeviceInfo: util.Map[String, String] = deviceInfo
+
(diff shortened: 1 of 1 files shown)


payonel:
hmm, that's reasonable -- thanks for your investigation

TheCodex6824:
No problem. And yeah, your way of checking distance seems to be the only way I can see of checking if a packet is wired or wireless - some sort of marker in the Packet API would be optimal but this works.

PR #2993, opened 2018-12-02
Fix not honoring event handler timeouts and cursor issues in OpenOS

evg-zhabotinsky:
* Fixed `computer.pullSignal()` from `event` library ignoring pending handler timeouts
    - Before the fix it always waited for a signal or until _caller-specified_ timeout was reached.
    - This caused tardy timers when signals weren't arriving often enough.
    - Consequently, this caused waiting (e.g. `os.sleep(n)`-ing) threads to lag or lock up completely in absence of signals.
* Fixed `term.pull()` and terminal input cursor blinking on every signal.
    - That was causing lots of unnecessary calls to the gpu component.
    - Now the cursor blinks exactly at specified intervals.
* Allowed to `term.setCursorBlink(number)` for arbitrary blink interval.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/core/cursor.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/core/cursor.lua
@@ -223,12 +223,18 @@ function core_cursor.read(cursor)
     clipboard = tty.keyboard,
     touch = tty.screen,
     drag = tty.screen,
-    drop = tty.screen
+    drop = tty.screen,
+    interrupted = true
   }
 
+  local blink = tty.window.blink or math.huge
+  if blink == true then blink = .5 end
+  local next_blink = computer.uptime() + blink
+  cursor:echo()
   while true do
     local next_line = cursor.data:find("\10")
     if next_line then
+      cursor:echo(false)
       local result = cursor.data:sub(1, next_line)
       local overflow = cursor.data:sub(next_line + 1)
       local history = text.trim(result)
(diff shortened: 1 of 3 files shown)


payonel:
HA I was wondering if someone was going to notice this issue :)
I already have this fixed in our private dev branch. But, I will compare your choice of fix against ours and see if there is anything I like more about your approach.

evg-zhabotinsky:
Which part? The timeouts or the cursor?
My fix was designed to touch as little as possible, to avoid breaking things I don't understand, so it probably can be improved.

Things that came to my mind just now:
* Fixed `term.pull()` ignoring signals without arguments (i.e. just the name)
* Setting blink interval to a string will actually succeed and then probably crash any input until reboot.
* Fixed `pullSignal` not checking uptime after the real pull and using the stale value
* I initially wanted to implement event timeout handling using proper priority queue, but then realized that you are looping over all handlers anyway.
* I made it so that `globalTimeout` only decreases until it's reached, to avoid handlers that pull messing it up

Question: How are handlers that pull events themselves are supposed to work? If a handler calls `event.pull()`, will the remaining handlers start before that nested pull completes? What if that pulling handler is in a thread? Is the "main" thread also a "thread", in a sense that `event.pull()` just registers a trigger-once handler with a timeout?

payonel:
I see that you're still poking at the cursor blinking issues, and I don't have much time to discuss this now (I will, eventually) but I wanted you to know that we're probably not going in that direction. At a minimum there are two separate concerns here. Proper event pulling timeout, and cursor behavior

evg-zhabotinsky:
> two separate concerns

Yes, probably. It's just that I tripped over both at once, and they seemed to be related.

My RDP implementation was severely lagging due to tardy timers. Fixing it by pushing events on every VirtualGPU update locked it up due to each event generating another update by blinking cursor. Disabling cursor blinking didn't help, because it still was "blinking" off and immediately back on on every event. Deleting cursor blinking code, apart from making cursor invisible and thus editing input hard, locked up the whole machine, threads were no longer waking up to do their jobs. So I fixed the timers to avoid relying on cursor or busy-waiting main process and to reduce delays. I then fixed the cursor to prevent remote display from lagging up client's GPU and network and wasting 200 energy for wirelessly transmitting a blink on every event.

Anyway, I'm actually using those fixes, so when they broke something again, I fixed the fixes.

payonel:
I've pushed my intended fixes for the timeout issue with events, but I've not made changes to the cursor code you've run into. If you would like to make a separate PR for just the cursor, we can discuss

My hesitance is that some of this cursor code you are expanding, or adding more calls to (the echo(false), etc) are driving the usage of a detail of the api I didn't love yet. The cursor lib is somewhat alpha and I didn't want to continue to use it until I really loved how it worked.

I would also be interested in running your rdp code, and then reviewing the cursor behavior myself. I might find an alternative fix for the cursor issues as well. It might be a similar fix, but it might be done in a slightly different approach

PR #3000, opened 2018-12-10
Add .man extension to all Manpages

JakobDev:
I just added the .man extension to all Manpages in CraftOS. Other files in CraftOS have already a extension like .lua or .txt. The .man extension is supported by the man program.


payonel:
the files are plain text files and not man page format, there is no value in changing their extension

PR #3009, opened 2018-12-23
Make it possible to use an analyzer as a tablet upgrade for reading component IDs.

AmandaCameron:
This is a first, MVP, pass at a "Barcode Reader" tablet upgrade. I've confirmed it works with both a single-node device, as well as a multi-node device ( such as an adapter ).

Things still to do:

  - [x] Blacklist it from non-tablet devices ( Or make it work with non-tablet devices)
  - [x] Multiplayer testing
  - [x] A sprite would be nice.
  - [x] (For non-components) Make it possible to get internal component's addresses (such as modems)
    - I'd like some feedback for this part.

The logic behind it for "immersion" for me is often devices will come with a barcode on the back of the unit or on the packaging for the MAC address of the device, for network set up purposes, so I'm working with the idea that that translates to the device's component address.

Picture of it working


payonel:
this is great, i like it! Thanks!

PR #3013, opened 2018-12-28
Fix server-side deadlock in filesystem activity handling

ds84182:
We've managed to hit a reproducible deadlock concerning filesystems. The exact setup is unknown, but using finer lock scopes should fix the issue.

I'm not able to test this change at the moment (I don't have MCP setup), and I can't verify if the change is valid Scala (but it seems like it should be fine).

### DIFF
--- a/src/main/scala/li/cil/oc/server/PacketSender.scala
+++ b/src/main/scala/li/cil/oc/server/PacketSender.scala
@@ -127,8 +127,8 @@ object PacketSender {
   // Avoid spamming the network with disk activity notices.
   val fileSystemAccessTimeouts = mutable.WeakHashMap.empty[Node, mutable.Map[String, Long]]
 
-  def sendFileSystemActivity(node: Node, host: EnvironmentHost, name: String) = fileSystemAccessTimeouts.synchronized {
-    fileSystemAccessTimeouts.get(node) match {
+  def sendFileSystemActivity(node: Node, host: EnvironmentHost, name: String) {
+    fileSystemAccessTimeouts.synchronized(fileSystemAccessTimeouts.get(node)) match {
       case Some(hostTimeouts) if hostTimeouts.getOrElse(name, 0L) > System.currentTimeMillis() => // Cooldown.
       case _ =>
         val event = host match {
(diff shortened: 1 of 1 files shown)


Vexatos:
Where/how exactly would the previous setup deadlock? I can't really see how using two locks instead of one would fix the issue. This at least ensures that there is only one event scheduled at a time.

notcake:
-snip-
_\<same information as ds84182 posted below>_

ds84182:
@Vexatos Essentially, we have a deadlock between 3 threads: The server thread, and two OC computer threads.

The server thread was in the middle of starting a computer and sending a computer.start network message. [Code](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/scala/li/cil/oc/server/machine/Machine.scala#L233) This attempts to lock a Filesystem component for network message deliver, but the Filesystem component is currently locked by the first OC thread.

The first OC thread was in the middle of attempting to send a filesystem activity packet, in `sendFileSystemActivity`. However, the `fileSystemAccessTimeouts` map is locked by the second OC thread, which was also in the middle of `sendFileSystemActivity`, for a different Filesystem.

The second OC thread, which was also in the middle of attempting to send a filesystem activity packet, entered `FileSystemAccessHandler` via the Forge eventbus. This caused a chain reaction where calling `markChanged` on the associated server rack caused a cascade that eventually got to `Machine.isRunning`, which tries to lock the state of the machine being started on the server thread.

Stack trace dumps available here:

[Server thread](https://gist.github.com/SwadicalRag/ef17cc00e72a0d4bb971491497eef96d#file-thread-dumb-L1066)

[First OpenComputers thread](https://gist.github.com/SwadicalRag/ef17cc00e72a0d4bb971491497eef96d#file-thread-dumb-L419)

[Second OpenComputers thread](https://gist.github.com/SwadicalRag/ef17cc00e72a0d4bb971491497eef96d#file-thread-dumb-L332)

The main issue comes from the lock on `fileSystemAccessTimeouts`.

src/main/scala/li/cil/oc/server/PacketSender.scala
```diff
@@ -127,8 +127,8 @@ object PacketSender {
   // Avoid spamming the network with disk activity notices.
   val fileSystemAccessTimeouts = mutable.WeakHashMap.empty[Node, mutable.Map[String, Long]]
 
-  def sendFileSystemActivity(node: Node, host: EnvironmentHost, name: String) = fileSystemAccessTimeouts.synchronized {
-    fileSystemAccessTimeouts.get(node) match {
+  def sendFileSystemActivity(node: Node, host: EnvironmentHost, name: String) {
+    fileSystemAccessTimeouts.synchronized(fileSystemAccessTimeouts.get(node)) match {
       case Some(hostTimeouts) if hostTimeouts.getOrElse(name, 0L) > System.currentTimeMillis() => // Cooldown.
```
Vexatos: These `hostTimeouts` are not synchronized and could _potentially_ cause CMEs. Might have to synchronize those as well?

Vexatos (changes requested):

asiekierka:
Uh?

asiekierka:
This has been open for long enough.

PR #3015, opened 2018-12-29
Update robot.names

anar4732:


payonel:
1. rebase to our default branch, master-MC1.7.10
2. technically it is Pac-Man: https://en.wikipedia.org/wiki/Pac-Man
3. funny name is funny, but no. you can remove it
4. i can find no reference that GUNTER was spelled with periods

PR #3043, opened 2019-02-09
Update pt_BR.lang

henrique-klayton:
Updated the file and fixed some writing errors.


CaelTheColher:
Lua is not an acronym. It should be written as "Lua" and not "LUA"

Issue #2928, opened 2018-08-21
AE2 Cells don't have extra information they should


AntiBlueQuirk:
Based on src/main/scala/li/cil/oc/integration/appeng/ConverterCellInventory.java, it looks like AE2 storage cells are supposed to have extra information attached about the state of the cell `{storedItemTypes=42, freeBytes=1234, etc}`, however, when I inspect the cells in the ME Drive using an Inventory Controller Upgrade in an Adapter, I only get normal item stack information. `{hasTag=true, name="...", etc.}`

I poked around a bit for the source of the issue, but couldn't find it. I can confirm that other AE2 integration is working, because I can access the ME Controller with an Adapter. (I know, because I hoped to check cell state, or get a table of all the cells in the network from the ME Controller.) I can also confirm that the converter is getting properly registered, because it appears in my log as `Registering converter li.cil.oc.integration.appeng.ConverterCellInventory`. It just doesn't seem to work.

- MC: 1.12.2
- Forge: 14.23.4.2745
- OC: 1.7.2.67
- AE2: rv5-stable-11

repo-alt:
Looking at the code I think that the drive contains instances of ItemBasicStorageCell, and logical contents  are actually stored in CellInventory (which can be accessed from the storage cell)
Perhaps converter for storage cell may be added to handle actual inventory, because as it is ConverterCellInventory is useless.

repo-alt:
like this:
`        } else if ((value instanceof ItemStack) && (((ItemStack)value).getItem() instanceof IStorageCell)) {
            IMEInventoryHandler<?> inventory = AEApi.instance().registries().cell().getCellInventory(
                (ItemStack) value, null, StorageChannel.ITEMS);
            if (inventory instanceof ICellInventoryHandler)
                convert(((ICellInventoryHandler) inventory).getCellInv(), output);
        }
`

repo-alt:
Code formatting in the comment is meh, will make a PR  at least for readability.
And, also, perhaps this is a more generic problem, and ConverterItemStack should call converters on ItemStack.getItem()?

PR #3082, opened 2019-04-06
AE2 Cells don't have extra information they should

repo-alt:
https://github.com/MightyPirates/OpenComputers/issues/2928


payonel:
nice, thank you

PR #3098, opened 2019-05-24
fixes issue "itemDamageRate Do not Work" #3041

### LINKED ISSUE #3041: itemDamageRate Do not Work
Hello,

I recently installed opencomputers and wanted to make it so that item took no damage.  After reading through the configs I found "itemDamageRate", which the configs states that when set to 0, item take no damage.  After setting this value to zero, my items still take damage while be utilized by robots.

Thank you for your Time

svitoos:
fixes #3041


payonel:
thank you. tested, works.

PR #3099, opened 2019-05-24
fixes issue "problem with enchanted instruments" #3050

### LINKED ISSUE #3050: [MC1.12.2] [OC 1.7.4.153] problem with enchanted instruments
When some blocks are destroyed with an enchanted tool, the robot returns false.
https://youtu.be/lUXPgaQlE5c

svitoos:
fixes #3050


payonel:
there were cases with blocks from other mods that refuse breaking that we needed to check for with the isBreaking condition - but I have a reasonable fix for both cases checked in now

PR #3100, opened 2019-05-25
More HTTP request methods #1938

### LINKED ISSUE #1938: More HTTP request methods
Some websites (e.g., github API) require to use HTTP request methods different from GET and POST, e.g., `PUT`, `PATCH`, or `DELETE`. It'd nice if OC programs could send such HTTP requests.

kchanakira:
Adds the ability to specify the HTTP method used when sending a request with the internet card.

Fixes https://github.com/MightyPirates/OpenComputers/issues/1938

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/internet.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/internet.lua
@@ -6,10 +6,11 @@ local internet = {}
 
 -------------------------------------------------------------------------------
 
-function internet.request(url, data, headers)
+function internet.request(url, data, headers, method)
   checkArg(1, url, "string")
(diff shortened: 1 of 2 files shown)


payonel:
is post the only true case for `setDoOutput`? see https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods

kchanakira:
> is post the only true case for `setDoOutput`? see https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods

POST, PUT, and PATCH requests should be able to have a request body, so `setDoOutput(true)`. HTTPUrlConnection doesn't support PATCH though.

payonel:
i tested this, works nicely. thanks

PR #3132, opened 2019-07-27
automatically adjust resolution to fit entire screen

Nex4rius:
This changes the resolution to minimize the black border around the screen. It works with any screen sizes and Tier 1, 2 and 3.

Example gif after installing it with:
~~~lua
pastebin run B9Enk6qp
~~~


payonel:
this would be a fine oppm project, but i actually would prefer to not adjust screen size on existing systems in the wild

Nex4rius:
Seems fair but technically this would only adjust the screen size on new installations of OpenOS and not affect the existing systems.

Issue #3090, opened 2019-04-30
AE2 full-block interface doesn't present as component if channels are disabled.


cyb0124:
In 1.12, if AE2 channels are disabled, then OC refuses to present full-block interfaces as components. However, cable-part interfaces work fine.

This should be the cause:
https://github.com/MightyPirates/OpenComputers/blob/7ceb846b5e0e7636d3d9af1dae6e253a1f040106/src/main/scala/li/cil/oc/integration/appeng/AEUtil.scala#L42
I see no reason why we should test for `areChannelsEnabled` here.

Combined with #3055, it means that in the lastest (1.7.4) OC release for 1.12, if AE2 channels are disabled, then there is really no way for OC to query items in a ME network. (#3055 prevents part interface from accessing NetworkControl while this issue prevents full-block interface from accessing NetworkControl.)

I found this issue from a recently released modpack (SkyFactory 4) that has AE2 channels disabled.

PR #3134, opened 2019-08-01
Fixed AE2 integration when AE2's channels are disabled

LeiserGeist:
I believe I properly fixed the issue referenced in #3090 

[The previous fix](https://github.com/MightyPirates/OpenComputers/commit/e660aeceddb72d477293dfa49ec8e860d58da998) did not seem to correctly fix the issue, and resulted in a `NoSuchElementException` being thrown when attempting to connect an adapter to an AE2 ME Interface, when AE2's channels were disabled. Controllers don't exist when channels are disabled, thus the exception.

Prior to that commit, it checked if ME Controllers exist before defining the integration for both ME Controllers *and* ME Interfaces. For controllers this check was necessary, but not for interfaces. The fix removed the checks from both.

By re-implementing the check *only* when defining ME Controllers, this fixed the exception and allows ME Interfaces to be used when channels are disabled.

Sorry for the potentially poor description!

### DIFF
--- a/src/main/scala/li/cil/oc/integration/appeng/AEUtil.scala
+++ b/src/main/scala/li/cil/oc/integration/appeng/AEUtil.scala
@@ -22,11 +22,13 @@ object AEUtil {
 
   def useNewItemDefinitionAPI = versionsWithNewItemDefinitionAPI.containsVersion(
     Loader.instance.getIndexedModList.get(Mods.AppliedEnergistics2.id).getProcessedVersion)
+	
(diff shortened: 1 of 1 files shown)


LeiserGeist:
Heh, something's gone horribly wrong with the Travis instance that's checking these PRs

Alexejhero:
Any clues on when this will be merged?

LeiserGeist:
Looks like this was implemented at some point, closing.

PR #3197, opened 2019-12-25
Switched to Gradle 5 on Minecraft 1.7.10

Vexatos:
Using a fork of ForgeGradle and CurseGradle which made the plugins compatible with Gradle 5. Huge thanks to @CDAGaming for helping with getting this working. Now OpenComputers should build fine in current versions of IntelliJ again, and hopefully still build fine on jenkins.

Lines 81 through 89 are a hack for now because the dependencies in line 122 and onward do not follow a proper maven directory structure and thus cannot be found by gradle anymore. We have to fix this in the future.

Please test this to see if it works for you.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,20 +9,21 @@ buildscript {
             name = "sonatype"
             url = "https://oss.sonatype.org/content/repositories/snapshots/"
         }
+        maven {
+            url = "https://jitpack.io"
+        }
     }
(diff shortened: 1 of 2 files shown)


payonel:
tested this and it works. though, testing this with a new git clone - intellij default jvm was 11 in like 4 different project settings - and this wasn't obvious.

PR #3198, opened 2019-12-25
Downgraded Ubuntu version on travis to hopefully fix #3144.

### LINKED ISSUE #3144: Travis CI Builds fail
Ever since 2 months ago, with build #3358, the Travis CI build has been having a weird error.

`Expected feature release number in range of 9 to 14, but got: 8`

This consistent error makes it difficult to see if a pull request actually breaks something, and ought to be fixed as soon as possible.

EDIT: Been looking around, I think the best way to change this is to change our distribution to `trusty` rather than `xenial`.

Vexatos:
Should fix #3144 and get travis working on 1.7.10 again, according to the research I did. Let's hope so.

Also switched the gradle wrapper to HTTPS because HTTP no longer lets you download anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 language: scala
 scala: 2.11.1
 jdk: oraclejdk8
+dist: trusty
 
 notifications:
   email: false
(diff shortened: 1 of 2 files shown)


Vexatos:
Looks like it's working.

PR #3199, opened 2019-12-26
Added binary data support to 'text.internal.reader'

Cruor:
Mode flag is not passed through to the buffer when creating the reader.
Causes binary files in `devfs` to not function as expected, since they are read as non binary files.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_text.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_text.lua
@@ -181,7 +181,7 @@ function text.internal.reader(txt, mode)
   }, {__index=text.internal.stream_base((mode or ""):match("b"))})
   process.closeOnExit(reader)
 
-  return require("buffer").new(more or "r", reader)
+  return require("buffer").new(mode or "r", reader)
 end
 
 function text.internal.writer(ostream, mode, append_txt)
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_text.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_text.lua
@@ -181,7 +181,7 @@ function text.internal.reader(txt, mode)
   }, {__index=text.internal.stream_base((mode or ""):match("b"))})
   process.closeOnExit(reader)
 
-  return require("buffer").new("r", reader)
+  return require("buffer").new(more or "r", reader)
 end
 
 function text.internal.writer(ostream, mode, append_txt)


payonel:
don't you mean `mode` and not `more` (looks like a typo) ? and can you give an example of a specific read that is misbehaves that this change fixes?

Cruor:
Yea, that is exactly what I meant... Bless typos in the middle of the night...
I set up a devfs to read directories from my webserver file listing. When I fetch midi files it fails because they are read as plaintext instead of binary data.

After applying that change it started playing them just fine.

Cruor:
```lua
if files and #files == 0 then
    target[filename] = {
        read = function(self, n)
            local cache = getInternetResource(url)

            if cache and cache.body then
                return string.sub(cache.body, 1, n)
            end

            return ""
        end
    }

else
    ...
end
```
Is part of the devfs proxying I'm using for files. Looking through the devfs code makes this seems like the right thing to do. ~~Turns out I did not understand the read part of devfs, but in this case it seems to be irrelevant~~

It respects it in the same reader function, but that doesn't seem to change anything for reading in this case.
https://github.com/MightyPirates/OpenComputers/blob/cca7f202d5040156bf93a0ef687fd6447fbfd07d/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_text.lua#L180

Incase this fix is deemed correct, it should probably also be applied to
https://github.com/MightyPirates/OpenComputers/blob/cca7f202d5040156bf93a0ef687fd6447fbfd07d/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_text.lua#L223

Cruor:
Reverted the change to test. Seems like it works when the proxy has open with a read function, but not directly with a read function.

Properly implemented .open works with binary reading, only having .read doesn't respect binary reading without my edit.

Not sure if this is intended or not? The updated code below works as expected.

```lua
if files and #files == 0 then
    target[filename] = {
        open = function(mode)
            if mode and not mode:match("r") then
                return nil, "only reading supported"
            end

            return {
                read = function(self, n)
                    local cache = getInternetResource(url)

                    if cache and cache.body then
                        if self.seek <= #cache.body then
                            local slice = string.sub(cache.body, self.seek, self.seek + n - 1)

                            self.seek = self.seek + n

                            return slice
                        end
                    end

                    return nil
                end,
                seek = 1
            }
        end
    }

else
    ...
end
```

payonel:
thank you for your patience. this change looks good, and yes, it should be applied to the writer as well. i enjoy that you went to so much trouble to figure this out with the devfs feature.

I don't understand what you mean by "Properly implemented .open works with binary reading, only having .read doesn't respect binary reading without my edit." It seems to me that this change is necessary for reading binary files in devfs that are using the text streams created by full_text.lua, but are you saying it isn't needed? if you would like to explain the other questions or concerns, perhaps you could open a discussion in https://oc.cil.li/

PR #3207, opened 2020-01-05
Add Ukrainian language

n1clud3:
It would be a good idea if you find ukrainian translators and translate OpenComputers on ukrainian.


payonel:
this PR target is from 1.12 to 1.7.10 -- that's not right
i'm okay if you only want to add transations to 1.12 because the translation file paths and item names changed and it's a lot of work to support both

Issue #3208, opened 2020-01-06
Make waypoints more safe


hohserg1:
Now waypoints are too unsafe: 
- navigation#findWaypoints result can't be verificated
- anyone can create duplicates of waypoints with same labels
- waypoint label can be changed by other players, who can't break concrete block even

 For solve this problems I suggest follow:
- let navigation#findWaypoints result contains a addresses of waypoint
- check player permissions before setting of label

What do you think about this?

payonel:
1. i'd rather not lock the waypoint
2. findWaypoints returns a table that already contains the position

payonel:
i see from your pr that you want the waypoint data to include the hardware address, i'm okay with adding that

hohserg1:
> 1. i'd rather not lock the waypoint

What?? I don't mean that
I try to say more understandably
```anyone can create duplicates of waypoints with same labels``` - it is a problem, but I suggest to solve it by verification through hardware address, is not by prohibition of the same label

hohserg1:
> i see from your pr that you want the waypoint data to include the hardware address, i'm okay with adding that

Do I need to improve something in this pr?

payonel:
i don't want the player lock on the waypoint

hohserg1:
You mean commit `Added checking player permissions for waypoint editing`?
It commit fix bug about player, who can't interact block, can edit label. Block interaction can be prohibited by something like GriefPrevention or WorldGuard glugins.
It commit does not prohibit block interaction in itself, it simply checks if player really can interact block

payonel:
i'm okay with only adding the component address to the packet response. if you want me to merge this pr, please remove the player locking

hohserg1:
> please remove the player locking

Where do you see it? 
Waypoint#playersWhoCanEdit will be get player id when player click on block and will be lose player id when player set label. 
And it do not save in nbt

payonel:
yep, id rather not negotiate the playing id in the process of setting the label - we're going to keep it simple and open. if you want to add the component address to the waypoint packet, that is fine

hohserg1:
I do not understand what do not you like. Similar method are used in vanilla sign and it work briliance

hohserg1:
> we're going to keep it simple and open

You mean that it is normal, when any player can change the label of waypoint, when he had no premissions in the multiplayer?
If remove the lock, then the labels will cease to be reliable.
Maybe,I do not understand something. You can somewhat explain your desire to unsafe labels?

[...]

PR #3211, opened 2020-01-10
Make waypoints more safe

hohserg1:
Discussion: https://github.com/MightyPirates/OpenComputers/issues/3208

Changes:

- navigation#findWaypoints gives addresses of waypoints also - it can be used for verification
- label of waypoint can't be changed by player, who do not have access to concrete waypoint block(like generalised support for WorldGuard and similar protective plugins and mods)

Label protection are use same method like vanilla sign


payonel:
merged the waypoint address commit from this

PR #3241, opened 2020-03-12
update funscii to 12/03/2020

asiekierka:
The highlight of this update is integrating Viznut's Unscii 2.0 changes. Also removes the Unifont license, as we haven't redistributed it in years, surely?


SNiTEBoBy (approved):
submit feedback and confirm merge #3241

asiekierka:
Well, Unifont has been relicensed! Let's use that instead.

PR #3246, opened 2020-03-15
Support for non-BMP codepoints

asiekierka:
The basic functionality seems to work, but there are a few omissions.

TODO (feel free to take this over):

- [ ] More locations of using .length() over .codePointCount(...) must be fixed
- [x] .codePointAt(x) must be replaced with .codePointAfter(currentIdx) to work correctly in iterators (or use the .codePoints stream, or use .offsetByCodePoints)
- [x] Fix unicode.reverse (uses Scala code which operates on an assumption of an Array[Char])
- [x] Fix unicode.sub (needs to operate based on offsetByCodePoints)

Possibly more?


SNiTEBoBy (approved):
submit feedback and confirm merge, #3246

asiekierka:
Working on a new version of the PR.

Issue #3253, opened 2020-03-26
EEPROMs not being changed when accessed by Transposer


bibo38:
**TLDR** Changing an EEPROM (e.g. label or data) and pulling it out of a computer using a transposer doesn't keep the EEPROM changes.

## Reproduction

Set up a Transposer which is connected to a computer and a chest:
![Bug Setup](https://user-images.githubusercontent.com/1812304/77680325-18caac80-6f94-11ea-9f37-2281014ba2eb.png)

Then install OpenOS on the computer, use the default Lua BIOS and create a `test.lua` file with the following contents:
```lua
component = require("component")

component.eeprom.setLabel("Testname")
component.eeprom.set("Will not work")
print(component.eeprom.getLabel())
component.transposer.transferItem(
  5,  -- Side connected to the computer (may need to be changed)
  1,  -- Chest
  1,  -- 1 EEPROM
  10, -- EEPROM Slot in the computer
  1   -- Chest slot
)
```

Run `test` in the console

### Expected behavior

The chest contains an EEPROM with the name `Testname`, which cannot be used to boot a PC.

### Actual behavior

The chest contains the default Lua BIOS from the Computer. It has the default label and works without a problem.

## Notes

- I didn't tested this with harddrives or floppy disks
- Also didn't tested it with other mods pulling items out of the computer
- If you remove the transposer code, then the tooltip of the EEPROM in the computer shows the wrong Label, but when you remove it (manually), it is changed to the correct label.

bibo38:
Are you already working on this @payonel ? I would also be interested to fix this issue and send a PR. If you have hints where the bug might be located, i would be interested in those.

HateMeLikeAPro:
This seems to not only be a problem with EEPROM, I have a "deploy station" that basically deploys an image to a HDD inserted into a dispenser. This is routed using a transposer. And when I `shell.execute("label set " .. addr .. " SomeLabel")` it only works if I pull out the HDD myself (opening the computer case and pulling the hdd from the inventory into my player inventory). If the transposer does it, the hdd seems to have been wrote to (the fs got formatted and a default install of openOS  is on there) but the label isn't there anymore

bibo38:
> the hdd seems to have been wrote to (the fs got formatted and a default install of openOS is on there) but the label isn't there anymore

[...]

PR #3255, opened 2020-03-27
Simulating the insertion instead of inserting directly

bibo38:
Previously the extraction was simulated and based on the simulated
extraction the extracted items were inserted. This was problematic, as
it relyed on the fact, that the simulation returns the same result as
the actual extraction. This caused problems, as extracting a component
only saves the changes to the component, when it is not simulated.
Furthermore it could cause item deduplication, if the actual extraction
returned something different.

To solve these issues, the insertion is also simulated on the simulated
extraction and the actual insertion is done on the actual extracted
items.

Solves #3253


asiekierka:
This took a bit of time to wrap my head around. Thank you!

PR #3259, opened 2020-03-29
Add boot time information to OpenOS

Ocawesome101:
Simple addition. Run `boottime` or `computer.bootTime()` to get the time it took OpenOS to boot.


payonel:
this is more appropriate in user samples, not a core api

Ocawesome101:
> this is more appropriate in user samples, not a core api

User samples?

PR #3265, opened 2020-04-02
Fixed Relay setStrength unlimited upper bound

jamesordner:
Range checking while setting the wireless signal strength for Relays was incorrect and unbounded. Behavior now matches that of wireless network cards.


asiekierka:
Thank you!

PR #3266, opened 2020-04-03
Make Lua BIOS take use tail call optimisation

skyem123:
I am not sure if this will do much, but for the sake of completeness I think it would be a good idea to do this.

### DIFF
--- a/src/main/resources/assets/opencomputers/lua/bios.lua
+++ b/src/main/resources/assets/opencomputers/lua/bios.lua
@@ -61,4 +61,4 @@ do
   end
   computer.beep(1000, 0.2)
 end
-init()
+return init()


aiverson:
This seems like a good idea to me too. Because all of the variables except `init` in the bios are already wrapped in a `do` block, they will be out of scope when `init` is invoked, but init itself won't be. The lack of a tail call means that the bios invocation will stay on the stack, which will anchor the function prototype of the chunk of the bios, which will also prevent the bytecode loaded from the bios from being garbage collected, which is a kilobyte or so of RAM being released by this change. (machine.lua doesn't hold a reference to the bios after creating the coroutine to run it in)
The local variables of the frame would also be held in the stack preventing their collection, which includes all of the code loaded from init. If the code loaded from init is just responsible for loading some other more complicated operating system and then tail-call into it, this would allow more of the memory involved in the loading to be reclaimed. Openloader for instance. (Openloader should also get this tail call optimization patch applied.)

payonel:
did some local testing, this seems good

PR #3295, opened 2020-05-10
Tiny fix for ls manual

avevad:
Removed unnecessary characters in ls manual (maybe copy+paste mistake)


payonel:
ah thanks. can  you retarget this for 1.7.10 and not 1.12? thanks

PR #3316, opened 2020-06-06
Added GT Seismic Prospector data reading support

repo-alt:


payonel:
thanks :) a good addition

PR #3319, opened 2020-06-06
Added source tank parameter to the transposer.transferFluid

repo-alt:


payonel:
can you suggest a mod to test this with?

repo-alt:
That was added mostly to support https://github.com/kekzdealer/KekzTech multitank. The parameter is optional, so I thought it would do no harm upstream, and I won't have to maintain a fork.
I also remember IHL (IC2 addon) chemical reactor having several outputs.
Perhaps there other mods providing several fluids in one block.

misterzirillo:
Any chance of this being merged? It would be helpful for AE2's Fluid Interface as well.

repo-alt:
There was a bug discovered in this function in our fork, fix pushed.

repo-alt:
Replaced with an updated version.

PR #3322, opened 2020-06-09
Added to robot.names

Bad-At-Usernames:
Added Multivac, a fictional self-aware computer the size of Washington created by Isaac Asimov.


payonel:
retarget 1.7.10 please, thanks

PR #3324, opened 2020-06-15
Disable debug cards by default

kmecpp:
It's a serious security risk to have debug cards enabled by default. Server admins might not know what they can do and I've also experienced the OC config getting reset if I make a minor error or OC isn't happy with my chunkloading world whitelist/blacklist for whatever reason.

Such events could be catastrophic for creative servers. Also many mods also have packet vulnerabilities that make it possible to spawn in any item.

For these reasons I believe it's critical to have them disabled by default.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1559,7 +1559,7 @@ opencomputers {
     # want to enable whitelisting of debug card users (managed by command
     # /oc_debugWhitelist). This will *not* remove the card, it will just
     # make all functions it provides error out.
-    debugCardAccess: allow
+    debugCardAccess: deny
 
     # Whether to always register the LuaJ architecture - even if the native
     # library is available. In that case it is possible to switch between


payonel:
you can't break out of the sandbox to the host even with debug cards. it opens up more griefing and interference options but not security over personal data

if you feel you have more info about a specific threat model with our debug card, please give me the opportunity to review in a more private matter, contact me in our discord/ / irc #oc on esper.net

kmecpp:
@payonel by security risk I mean someone could get OP and destroy the entire server and potentially other servers linked on a Bungee network. Isn't that still a pretty severe problem?

asiekierka:
`It's a serious security risk to have debug cards enabled by default. Server admins might not know what they can do`

If you can cheat in items, you can cause arbitrary server and client crashes in any 1.7.10 modpack of non-trivial size. If you can't cheat in items, you cannot acquire a Debug Card without already having OP rights. Is there a different threat model you had in mind?

Issue #3364, opened 2020-10-05
RFC 5789 (Section 2) Support: Support for "PATCH" HTTP Method


tycoonlover1359:
As the title states, it would be very nice if support could be added for the `PATCH` HTTP Method as defined in [section 2 of RFC 5789](https://tools.ietf.org/html/rfc5789#section-2).

As it stands now, I have run into one website (quite popular) website that only supports the HTTP verbs as a way of handling requests; in other words, only the verb itself is used to tell the website how to process the data sent to it and inserting the `X-HTTP-Method-Override` header results in the website returning a `422 Unprocessable Entity` error, thereby resulting in an error and/or simply malformed (or just `nil`) content when trying to act upon the response from the website

The website in question is [Airtable](https://airtable.com/), a fairly popular online NoSQL (at least in terms of the programming required to interface with it) database similar to that of [RestDB](https://restdb.io).

Airtable requires the use of HTTP verbs in order to process requests sent to its APIs. Due to the mod believing that a `PATCH` request is an invalid HTTP method, **non-destructively** updating records within Airtable is impossible; a non-destructive update being an update that only updates the fields specified within the request body. Fully **destructive** updates are fine, as those make use of `PUT` requests, which are supported. As alluded to earlier, the `X-HTTP-Method-Override` header is not supported by Airtable and will result in `422` error.

Support for the `PATCH` HTTP method in RFC 5789 would be greatly appreciated, as I'm sure Airtable isn't the only website that does not allow for the use of the `X-HTTP-Method-Override` header.

Thanks.

SquidDev:
OpenComputers uses `HttpUrlConnection` to make HTTP requests, which [sadly doesn't support PATCH][methods]. It's fairly odd, but I guess PATCH is a relatively recent addition.

CC: Tweaked got around that by rewriting HTTP to use [Netty's HTTP support][cct] (which has the additional advantage of not requiring a thread for each request). I suspect that's the easiest solution for OC, but I realise that's a fairly hefty piece of work.

[methods]: https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#setRequestMethod-java.lang.String-
[cct]: https://github.com/SquidDev-CC/CC-Tweaked/blob/334ca65482b428821de83070f7b8e0f8a950a7fe/src/main/java/dan200/computercraft/core/apis/http/request/HttpRequest.java#L157

tycoonlover1359:
Hmmm, well alrighty then.

[...]

PR #3374, opened 2020-11-12
Http custom methods 1.7.10

hohserg1:
Fix trouble #3364 by using apache http.
As example, it allow to use PATCH http method: https://docs.github.com/en/free-pro-team@latest/rest/reference/gists#update-a-gist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 /*.iws
 /.idea
 /out
+/classes/
 
 # Debugging
 /eclipse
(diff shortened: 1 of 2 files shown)


asiekierka:
This PR fails to reproduce the HTTP proxy support present in OpenComputers. Is there a reason for that?

hohserg1:
How I can test such case?

asiekierka:
Done.

PR #3378, opened 2020-11-23
AE Network items iterator

repo-alt:
Currently enumerating items in the massive AE network requires a lot of memory. This PR adds an iterator to fetch items one by one.


repo-alt:
(there were some fixes to the iterator since then, need to look it up)

PR #3403, opened 2021-01-16
support video ram buffer.

mckuhei:


Ocawesome101:
What does this do?

aiverson:
From reading the code, this appears to extend the `edit` program in OpenOS to add functionality to restore the terminal to how it was before the program was invoked, so that rather than the terminal being reset to put a prompt at the top of the screen with all the output of previous commands being lost, it instead keeps the previous command output and makes a new prompt after the line where edit was invoked.
Unfortunately, it has a bug in it. The `buff` variable storing the ID of the buffer is not used in the bitblit calls, which instead have a hardcoded index of 1, which means that if some other program has previously allocated a vram buffer, it will use the wrong buffer and trample on that other program's storage, then fail to restore the terminal when it finishes depending on the resolution of whatever buffer is in slot one.
It also modifies the default keybindings to remove shift-backspace for no reason that I can see.

It might be better to implement the feature this adds in the shell though, that way any program that takes over the whole screen can be restored from rather than just this one. However, doing that would prevent any other program from using the vram API on a T1 GPU without patching the shell or providing an escape hatch, and it would require additional complexity in the code to figure out whether a terminal restore is needed. This implementation is simpler and would work fine if the bug I mentioned earlier is fixed.

mckuhei:
> From reading the code, this appears to extend the `edit` program in OpenOS to add functionality to restore the terminal to how it was before the program was invoked, so that rather than the terminal being reset to put a prompt at the top of the screen with all the output of previous commands being lost, it instead keeps the previous command output and makes a new prompt after the line where edit was invoked. Unfortunately, it has a bug in it. The `buff` variable storing the ID of the buffer is not used in the bitblit calls, which instead have a hardcoded index of 1, which means that if some other program has previously allocated a vram buffer, it will use the wrong buffer and trample on that other program's storage, then fail to restore the terminal when it finishes depending on the resolution of whatever buffer is in slot one. It also modifies the default keybindings to remove shift-backspace for no reason that I can see.
> 
> It might be better to implement the feature this adds in the shell though, that way any program that takes over the whole screen can be restored from rather than just this one. However, doing that would prevent any other program from using the vram API on a T1 GPU without patching the shell or providing an escape hatch, and it would require additional complexity in the code to figure out whether a terminal restore is needed. This implementation is simpler and would work fine if the bug I mentioned earlier is fixed.

what?

Ocawesome101:
Ah, so it's a single-purpose version of the Full Screen Buffer feature terminals like Konsole have.

This would probably be better facilitated as an addition to the VT100 emulator or the `term` library.

PR #3428, opened 2021-04-21
Added config options for ranges of sync packet sending

hohserg1:
Useful for some specific modpacks and gamemodes. 
As example, at anarchistic gamemode may have sense to reduce radius of packet sending from server to client for preventing cheating by packet sniffing

Also, fix `gradlew build`, now resolving dependencies from repos is broken


asiekierka:
Resolved by https://github.com/MightyPirates/OpenComputers/commit/87d57cad8f2876d67deb29c551a6099b86f5ce13 - though with a bit less granularity.

PR #3494, opened 2022-01-19
Fixed OC Block GUIs with Y < 0 for CubicChunks compatibility

big-lip-bob:
Basically, OC encodes the slot you clicked for the Server Rack block in the Y coordinate, that sure works with a bitmask in a restricted positive Y range, but said bitmask fails when negative values are needed, thus i replaced `& 0x00FFFFFF` with a more negative friendly `<< 8 >> 8`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,11 +4,11 @@ buildscript {
         jcenter()
         maven {
             name = "forge"
-            url = "https://files.minecraftforge.net/maven"
+            url = "https://maven.minecraftforge.net/"
         }
     }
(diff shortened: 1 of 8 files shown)


asiekierka:
The block GUI fix has been added to OC 1.7.6.

The issue of robot flight height is, in my opinion, too niche to be handled within OC itself - though I will open an issue for increasing the maximum height limit.

asiekierka:
https://github.com/MightyPirates/OpenComputers/issues/3515

PR #3517, opened 2022-06-11
Added source tank parameter to the transposer.transferFluid

repo-alt:
Make fluid transfer speed configurable

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1391,6 +1391,11 @@ opencomputers {
 
     # Radius the MFU is able to operate in
     mfuRange: 3
+
+    # Transposer fluid transfer rate in millibuckets per second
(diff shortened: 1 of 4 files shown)


src/main/resources/application.conf
```diff
@@ -1391,6 +1391,11 @@ opencomputers {
 
     # Radius the MFU is able to operate in
     mfuRange: 3
+
+    # Transposer fluid transfer rate in millibuckets per second
+    # It may transfer unlimited amount per operation, but will then 
+    # wait ((Amount in buckets)/16) seconds
+    transposerFluidTransferRate=16000
```
asiekierka, line 1398: Why is the default changed from OC's 4000 to 16000?

asiekierka: Changed it back to 4000. We can always tweak it later if needed.

PR #3518, opened 2022-06-11
Support for non-BMP codepoints, take 2

asiekierka:


src/main/scala/li/cil/oc/client/renderer/font/FontParserHex.java
```diff
@@ -67,9 +69,11 @@ public void initialize() {
 
     @Override
     public ByteBuffer getGlyph(int charCode) {
-        if (charCode < 0 || charCode >= glyphs.length || glyphs[charCode] == null || glyphs[charCode].length == 0)
+        if (!glyphs.containsKey(charCode))
+            return null;
+        final byte[] glyph = glyphs.get(charCode);
+        if (glyph == null || glyph.length <= 0)
```
payonel, line 75: was this wrong before? why would there be an empty glyph

src/main/scala/li/cil/oc/common/PacketBuilder.scala
```diff
@@ -58,6 +58,12 @@ abstract class PacketBuilder(stream: OutputStream) extends DataOutputStream(stre
     }
   }
 
+  def writeMedium(v: Int) = {
+    writeByte(v & 0xFF)
+    writeByte((v >> 8) & 0xFF)
+    writeByte((v >> 16) & 0xFF)
+  }
```
payonel, line 65: I think "medium" is a naming convention from database languages but we don't use it in code. Could we call this writeInt24? (with readInt24)

src/main/java/li/cil/oc/util/ExtendedUnicodeHelper.java
```diff
@@ -0,0 +1,41 @@
...
+public final class ExtendedUnicodeHelper {
+	private ExtendedUnicodeHelper() {
+
+	}
+
+	public static int length(String s) {
+		return s.codePointCount(0, s.length());
+	}
+
+	public static String reverse(String s) {
```
payonel, line 15: is this reverse code only used in string.reverse? have we tested s.reverse.reverse on invalid bytes? should s == s.reverse.reverse always?

src/main/scala/li/cil/oc/common/PacketBuilder.scala
asiekierka: It's actually [from Netty](https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#readUnsignedMedium--).

src/main/scala/li/cil/oc/client/renderer/font/FontParserHex.java
asiekierka: Overzealous more than wrong

src/main/java/li/cil/oc/util/ExtendedUnicodeHelper.java
asiekierka: 1. Yes, but it occurs twice (luac, luaj), and I didn't want to make the UnicodeAPI code even less understandable by throwing that implementation in there.
2. No, but I do wonder if they can even appear. Lua uses UTF-8, which is then converted to UTF-16 for Java Strings.
3. I'm not sure there's a standard, really? With this code, it should.

payonel:
@asiekierka looks good to me. merge when you're ready

PR #3521, opened 2022-06-12
fix #3182 by passing inventory source to transposer.getInventoryName

### LINKED ISSUE #3182: Transposers getInvenotryName returning airs instead of entity inventory names
When you use the transposer's getInventoryName method and a player stands in the way, the transposer will return minecraft:air instead of the name of the inventory
image

asiekierka:


payonel:
i like the grouping of inventory data. I'll do some testing with this change

asiekierka:
I've already done preliminary testing, seems to work fine

PR #3525, opened 2022-06-16
added new operationg system openOSmod

igorkll:


asiekierka:
First of all, if it's a patched version of OpenOS, why would we distribute the unpatched and patched version?

Second of all, again, if you want @payonel to review your OpenOS changes, you need to provide them in a way where we can properly compare the differences, and ideally be able to merge them in in smaller batches.

We won't accept a "269 new files" PR, especially as neither of us can read Russian.

PR #3531, opened 2022-06-21
Add dimension to machine position debug statements

repo-alt:


src/main/scala/li/cil/oc/server/machine/luac/NativeLuaArchitecture.scala
```diff
@@ -414,19 +414,22 @@ abstract class NativeLuaArchitecture(val machine: api.machine.Machine) extends A
...
+        OpenComputers.log.warn(s"Could not persist computer @ ${machinePosition}.\n${e.toString}")
         nbt.removeTag("state")
     }
 
     // Limit memory again.
     recomputeMemory(machine.host.internalComponents)
   }
+
+  /** Helper method for printing the machine position in error messages and debug statements. */
+  private def machinePosition = s"(${machine.host.xPosition}, ${machine.host.yPosition}, ${machine.host.zPosition}, ${machine.host.world().provider.dimensionId})"
```
asiekierka: Given that both access machine.host, maybe this should become a helper method *on* machine.host, or a static helper accepting machine.host, if possible?

(Also, might wanna check for a null `world()` or `provider`, just in case. Error handling should be watertight)

asiekierka:
Thank you!

PR #3533, opened 2022-06-25
Add fluid container item converter

repo-alt:
(Volumetric flask, certus tank, universal cell, etc.)
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8807

"fluid_hasTag" key may look dubious, but it corresponds to the ConverterItemStack


asiekierka:
Would it be a better idea to consider appending a `fluid` map instead, to both make the relationship clearer (item contains a fluid) and to resolve the naming conundrum? Like how enchantments are a list of maps...

PR #3558, opened 2022-09-15
Fix ls.lua timestamp

Smok1e:
filesystem.lastModified returns timestamp as milliseconds but os.date using it as seconds, so files modification date was shown wrong


asiekierka:
Confirmed and reproduced - thank you for the fix!

PR #3564, opened 2022-09-25
Make native library config options more flexible

### LINKED ISSUE #3561: Improve forceNativeLibWithName config option
Currently, using the `forceNativeLibWithName` option isn't a perfect solution on platforms missing native library support.

There are 3 native libraries to load, one for lua 5.2, 5.3, and 5.4. However, that config option only allows specifying a single string, meaning only one custom library can be loaded. This means that platforms without native support in the code cannot have every version of lua working at the same time as normal. Also, due to how `LuaStateFactory` handles native library loading, if the last library to be loaded is not found (or in other words, the lua 5.4 native isn't used as the single version specified in the config), it thinks that there is no native library support and disables that system entirely.

Possible solutions for this are:

1. Changing the `forceNativeLibWithName` to take a string pattern instead of just a normal string. For example, there could be a "$version" pattern that gets expanded to the lua version being loaded, so users can set the option to something like "libjnlua$version-linux-ppc64.so" to handle the different versions.
2. Changing the `forceNativeLibWithName` option to be a map of lua version to library name instead of a single string, with an entry for each lua version.

Edit: It occurred to me that the issue below is probably happening because of the `forceNativeLibWithName` issue described above

<del>Lastly, using this config option still requires users to pack natives into the OC mod jar itself to be able to load them. Ideally, the native loader should check if an appropriately named native is in the place it would have extracted a native to, and if so, use that native instead of loading from the jar. The final native name as used by the mod does include the OC version though, so users would have to rename their natives after every update - a bit annoying but maybe fine? It might also be a good opportunity to combine this with the `forceNativeLibWithName` changes above, and make that option try loading libraries from the filesystem first instead of looking in the jar. </del>

TheCodex6824:
Closes #3561 

This splits `forceNativeLibWithName` into `forceNativeLibPlatform` and `forceNativeLibPathFirst`. Tested to work with custom native libraries on ppc64 on both 1.7.10 and 1.12.2.

This is technically a breaking config change, but since the native libraries were changed in the latest version, anyone using custom natives before would have had their stuff break anyway.

If someone on a "supported" system could test this to make sure I didn't break the normal configuration, that would be great. I don't think I messed anything up but given I'm not on those architectures anymore I can't really test it myself. I also didn't restrict the path for the native library directory to the Minecraft directory - I figured some people may want to have a shared library path to use across instances / servers or something. I don't think that's a security concern given /tmp can be used for libraries even before this change, but if I'm wrong please correct me.


src/main/scala/li/cil/oc/server/machine/luac/LuaStateFactory.scala
```diff
@@ -170,13 +173,26 @@ abstract class LuaStateFactory {
 
   def isAvailable = haveNativeLibrary
 
-  val is64Bit = Architecture.IS_OS_X64
+  val is64Bit = {
```
asiekierka, line 176: Thank you for the effort, but...

This function in general is pretty silly. We use it to figure out whether or not to compensate for larger pointer sizes when calculating the maximum amount of available RAM - in practice, I should instead add a JNI method to simply *ask* what the Lua instance's pointer size is, and adjust accordingly.

asiekierka:
This is excellent, thank you!

PR #3565, opened 2022-09-27
Fix tablet assigning itself an ID on the client when it should not have one

### LINKED ISSUE #3552: Creative Tablet does not boot
> By the way, the creative (from inventory) tablet does not work at all (hand made tablet works)

_Originally posted by @Bumer-32 in https://github.com/MightyPirates/OpenComputers/issues/3548#issuecomment-1235619842_

TheCodex6824:
Fixes #3552 

The tablet in 1.12 has an [extra cache getter](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.12/src/main/scala/li/cil/oc/common/item/Tablet.scala#L618) that doesn't check if the tablet has an ID, which causes it to generate one if it is missing. Any items in the creative menu only exist on the client, so the tablet then thinks there is a tablet on the server with the same ID when there is not.

### DIFF
--- a/src/main/scala/li/cil/oc/common/item/Tablet.scala
+++ b/src/main/scala/li/cil/oc/common/item/Tablet.scala
@@ -616,12 +616,14 @@ object Tablet {
     override protected def timeout = 5
 
     def getWeak(stack: ItemStack): Option[TabletWrapper] = {
-      val key = getId(stack)
-      val map = cache.asMap
-      if (map.containsKey(key))
-        Some(map.entrySet.find(entry => entry.getKey == key).get.getValue)
-      else
-        None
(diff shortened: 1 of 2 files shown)


asiekierka:
A cleaner solution would be to extend `getId(stack: ItemStack)` with a boolean `createIfNotFound`, which adjusts behavior accordingly. This way, we're not introducing an additional dependency on the item stack's NBT internals.

TheCodex6824:
That sounds like a good idea. Although in that case, the return value should probably be changed to an `Option` in case `createIfNotFound` is false and there is no ID. Or maybe there can be a `hasIdAssigned(): boolean` method instead? What do you think?

asiekierka:
Separate `getOrCreateId` method, then, plus `getId` now returning an `Option`, would be the compromise solution I'd go for.

Issue #3560, opened 2022-09-19
latest.log : In game manual


AR2000AR:
> [13:14:07] [Client thread/WARN] [opencomputers]: Gosh I'm popular! Too many tabs were added to the OpenComputers in-game manual, so some won't be shown. In case this actually happens, let me know and I'll look into making them scrollable or something...

Found this in the log while looking for persistence error with the screens. Just following the instruction by reporting that.

Mods : 
- Commons0815-1.12.2-1.4.0.jar
- Computronics-1.12.2-1.6.6.jar
- foamfix-0.10.14-1.12.2.jar
- jei_1.12.2-4.16.1.302.jar
- malisiscore-1.12.2-6.5.1.jar
- mcjtylib-1.12-3.5.4.jar
- Mekanism-1.12.2-9.8.3.390.jar
- MekanismGenerators-1.12.2-9.8.3.390.jar
- ocdevices-MC1.12.2-1.0.0-44.jar
- OpenComputers-MC1.12.2-1.7.7+5413028.jar
- OpenPrinter-1.12.2-0.1.0.16.jar
- OpenSecurity-1.12.2-1.0-83.jar
- OptiFine_1.12.2_HD_U_G5.jar
- performant-1.11.jar
- SGCraft-2.0.5.jar

asiekierka:
After seven and a half years, someone has finally managed to hit this issue!

AR2000AR:
> After seven and a half years, someone has finally managed to hit this issue!

or no one saw this message or bothered to report it

PublicVoidUpdate:
I hit it once, but I didn't think to report it because I thought it was an issue with a different mod (⁠⊙⁠_⁠◎⁠)

brisingraerowing:
I may actually be hitting this and never noticed (I generally don't read logs unless there's a crash). I have a lot more addons than the OP.

PR #3572, opened 2022-10-31
Port OpenComputers to Minecraft 1.16.5

KosmosPrime:
This is a port of the 1.12 branch (as it appeared in May) for Minecraft 1.16.5. I've done some testing on the OC features to make sure they work (the mod is very much playable), but it's a big mod so I haven't tested every case yet (similar with mod compat: it works but it might behave differently in some cases).
There are some things I've had to change, for example, GUI input works differently in Minecraft and so the semantics of the keyboard are different as well. Because of this, OPPM packages might be incompatible (but I've not removed/looked into that issue yet) and there have been some changes to the API.
By the way, I'm making this as a pull request for the 1.12 branch (because that's what it's based on) but the intent is to create a new branch.


asiekierka:
Impressive work! However, there's a bit of a problem... (a) payonel and I are both effectively stuck in 1.12.2 land, not willing to move beyond it; (b) Sangar is probably more willing to push OpenComputers 2 on 1.16+ as the de-facto sequel. However, I'll ask around.

src/main/scala/li/cil/oc/integration/enderstorage/DriverFrequencyOwner.java
```diff
@@ -84,9 +84,10 @@ public Object[] setFrequency(final Context context, final Arguments args) {
             return null;
         }
 
-        @Callback(doc = "function():string -- Get the name of the owner, which is usually a player's name or 'global'.")
+        @Callback(doc = "function():string or nil -- Get the name of the owner, which is usually a player's name or nil.")
```
asiekierka, line 87: Keeping compatibility with older code might be nice - so it might be worth returning "global" here, instead of null, as in 1.7.10.

src/main/scala/li/cil/oc/server/machine/luac/LuaStateFactory.scala
```diff
@@ -137,7 +137,12 @@ abstract class LuaStateFactory {
   // loaded at the time the initializer runs.
   private def prepareLoad(lib: String): Unit = jnlua.NativeSupport.getInstance().setLoader(new Loader {
     def load(): Unit = {
-      System.load(lib)
+      // This redirects to `System.load(lib)`. It is required because
+      // Java expects the correct ClassLoader to also load the native
+      // library. Simply calling load here doesn't work in dev,
+      // because the JNLua jar only gets merged when the mod is built.
+      // Production environments are unaffected.
+      jnlua.NativeSupport.proxyLoad(lib)
```
asiekierka: Any changes to JNLua will need to go into a separate PR: https://github.com/MightyPirates/OC-JNLua - we actually have a consistent build system for it now.

src/main/scala/li/cil/oc/integration/enderstorage/DriverFrequencyOwner.java
asiekierka: Except it has been returning null in 1.12. Well then... @payonel - I'd say we fix the 1.12 docs to document the nil behaviour.

build.gradle
```diff
@@ -170,13 +189,21 @@ allprojects {
...
+            exclude '*', 'META-INF', 'META-INF/**'
         }
     }
     manifest {
-        attributes FMLCorePlugin: "li.cil.oc.common.launch.TransformerLoader"
-        attributes FMLCorePluginContainsFMLMod: "true"
-        attributes FMLAT: "oc_at.cfg"
+        attributes([
+            "Specification-Title": "opencomputers",
+            "Specification-Vendor": "li.cil.oc",
```
fnuecke, line 222: Is there a reason this is using the string directly here, but the variable below? Otherwise also using `"${config.mod.group}"` here to keep it homogenous would be nice.

src/main/java/li/cil/oc/api/FileSystem.java
```diff
@@ -22,29 +23,24 @@
...
      *
-     * @param clazz  the class whose containing JAR to wrap.
-     * @param domain the domain, usually your mod's ID.
-     * @param root   an optional subdirectory.
-     * @return a file system wrapping the specified folder.
+     * @param loc the location where the file system's contents are stored.
+     * @return a file system wrapping the specified resource.
      */
-    public static li.cil.oc.api.fs.FileSystem fromClass(final Class<?> clazz, final String domain, final String root) {
+    public static li.cil.oc.api.fs.FileSystem fromResource(final ResourceLocation loc) {
```
fnuecke: Minor (feel free to ignore): would suggest not shortening words in parameters, i.e. loc -> location.

src/main/java/li/cil/oc/api/detail/ItemAPI.java
```diff
@@ -41,25 +42,27 @@ public interface ItemAPI {
...
      * @param name    the label and identifier to use for the loot disk.
+     * @param loc     the location where the disk's contents are stored.
 	 * @param color   the color of the disk, as a Minecraft color.
      * @param factory the callable to call for creating file system instances.
      * @param doRecipeCycling whether to include this floppy disk in floppy disk cycling.
      * @return an item stack representing the registered loot disk, to allow
      * adding a recipe for your loot disk, for example.
      */
-    ItemStack registerFloppy(String name, EnumDyeColor color, Callable<li.cil.oc.api.fs.FileSystem> factory, boolean doRecipeCycling);
+    ItemStack registerFloppy(String name, ResourceLocation loc, DyeColor color,
```
fnuecke, line 64: Minor (feel free to ignore): would suggest not shortening words in parameters, i.e. loc -> location.

fnuecke:
Very very cool, thanks so much for taking the time to do this. As asie said, it's a lot, I'll try to help review this, but it'll take some time.

Assuming this PR is good (which it looks like from what I've seen so far!): as asie said, the other maintainers will probably stick to pre-1.16. Would you be able to/interested in maintaining the 1.16.5 version going forward?

src/main/resources/assets/opencomputers/bcbuilder/blocks/adapter.json
```diff
@@ -1,12 +0,0 @@
-[{
```
fnuecke, line 1: I take it this was BuildCraft Builder integration, from the name of the dir? Is this handled differently now, or was compat removed?

src/main/resources/META-INF/mods.toml
```diff
@@ -0,0 +1,29 @@
+modLoader="scorge"
+loaderVersion="[${sversion},)"
+license="MIT, CC0 (see LICENSE)"
+issueTrackerURL="https://github.com/MightyPirates/OpenComputers/issues"
+
+[[mods]]
+modId="opencomputers"
```
fnuecke, line 7: Unless there's a very *very* good reason for changing the mod ID, I'd rather it doesn't:
- that also means people integrating with the mod who've done so before will definitely stumble over this.
- it adds a lot of noise to the PR (in particular *all* the manual pages having to be touched).

src/main/resources/assets/opencomputers/loot/network/data/usr/bin/nc.lua
```diff
@@ -38,7 +38,7 @@ local function handleTcp()
                 elseif e[2] == "message" then
                     term.write(e[4])
                 end
-            elseif e[1] == "key_up" and channel then
+            elseif e[1] == "key_up" and channel and e[3] ~= 0 then
```
fnuecke: Is this is a fix that would also make sense to backport to 1.12 and such? @payonel, what do you think?

src/main/scala/li/cil/oc/CreativeTab.scala
```diff
@@ -1,11 +1,17 @@
 package li.cil.oc
 
-import net.minecraft.creativetab.CreativeTabs
+import li.cil.oc.common.init.Items
+import net.minecraft.item.ItemGroup
+import net.minecraft.item.ItemStack
+import net.minecraft.util.NonNullList
 
-object CreativeTab extends CreativeTabs(CreativeTabs.getNextID, OpenComputers.Name) {
+object CreativeTab extends ItemGroup(-1, OpenComputers.Name) {
```
fnuecke: There should be a constructor overload that just takes the name, so the magical `-1` could be removed.

src/main/scala/li/cil/oc/client/KeyBindings.scala
```diff
@@ -1,44 +1,56 @@
...
+    extendedTooltip.getKeyModifier match {
+      // KeyModifier.NONE does not accept pure modifier keys by default, so check for that.
+      case KeyModifier.NONE if KeyModifier.isKeyCodeModifier(extendedTooltip.getKey) => isActive(extendedTooltip.getKey)
+      case mod if mod.isActive(extendedTooltip.getKeyConflictContext) => isActive(extendedTooltip.getKey)
+      case _ => false
+    }
+  }
 
-  def isPastingClipboard: Boolean = isKeyBindingPressed(clipboardPaste)
+  def isAnalyzeCopyingAddress: Boolean = analyzeCopyAddr.isDown
```
fnuecke, line 38: Shouldn't this be `isAnalyzerCopyingAddress`? (missing `r`)

src/main/scala/li/cil/oc/client/PacketHandler.scala
```diff
@@ -115,23 +111,28 @@ object PacketHandler extends CommonPacketHandler {
...
       case _ => // Invalid packet.
     }
 
   def onAnalyze(p: PacketParser) {
     val address = p.readUTF()
-    if (Keyboard.isKeyDown(Keyboard.KEY_LCONTROL) || Keyboard.isKeyDown(Keyboard.KEY_LCONTROL)) {
-      GuiScreen.setClipboardString(address)
-      p.player.sendMessage(Localization.Analyzer.AddressCopied)
+    if (KeyBindings.isAnalyzeCopyingAddress) {
+      RenderSystem.recordRenderCall(new IRenderCall {
```
fnuecke, line 124: Hm, using `RenderSystem.recordRencerCall` here feels somewhat out of place? I'd say it makes more sense to make the `NetworkEvent.Context` available to the methods needing synchronization (new parameter, probably) and use `context.enqueueWork` instead?

src/main/scala/li/cil/oc/client/gui/widget/Widget.scala
```diff
@@ -1,5 +1,8 @@
 package li.cil.oc.client.gui.widget
 
+import com.mojang.blaze3d.matrix.MatrixStack
+
+@Deprecated
```
fnuecke, line 5: Just curious, what's the motivation behind marking this deprecated?

src/main/scala/li/cil/oc/client/renderer/HighlightRenderer.scala
```diff
@@ -20,227 +21,70 @@ object HighlightRenderer {
...
-        t.draw()
 
-        RenderState.disableBlend()
-        RenderState.popAttrib()
-        GlStateManager.popMatrix()
+        stack.popPose()
       }
     }
-
-    if (hitInfo.typeOfHit == RayTraceResult.Type.BLOCK) e.getPlayer.getEntityWorld.getTileEntity(hitInfo.getBlockPos) match {
```
fnuecke, line 100: Has the remaining been moved somewhere else, or why has it been removed?

src/main/scala/li/cil/oc/client/renderer/PetRenderer.scala
```diff
@@ -35,7 +39,8 @@ object PetRenderer {
     "076541f1-f10a-46de-a127-dfab8adfbb75" ->(0.2, 1.0, 0.1), // vifino
     "e7e90198-0ccf-4662-a827-192ec8f4419d" ->(0.0, 0.2, 0.6), // Izaya
     "f514ee69-7bbb-4e46-9e94-d8176324cec2" ->(0.098, 0.471, 0.784), // Wobbo
-    "f812c043-78ba-4324-82ae-e8f05c52ae6e" ->(0.1, 0.8, 0.5) // payonel
+    "f812c043-78ba-4324-82ae-e8f05c52ae6e" ->(0.1, 0.8, 0.5), // payonel
+    "1db17ee7-8830-4bac-8018-de154340aae6" ->(0.0, 0.5, 1.0) // Kosmos
```
fnuecke, line 43: Cheeky ;) I'm OK with it though.

src/main/scala/li/cil/oc/client/renderer/RenderCache.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public void render(MatrixStack stack) {
+        // Apply transform globally so we don't have to update stored vertices.
+        RenderSystem.pushMatrix();
+        RenderSystem.multMatrix(stack.last().pose());
+
+        cached.forEach(frame -> {
+            frame.type().setupRenderState();
+            DrawState state = frame.state();
+            state.format().setupBufferState(MemoryUtil.memAddress(frame.data()));
```
fnuecke, line 105: Hmm, this seems uncomfortably low-level; does this have massive benefits over using the more high-level `BufferBuilder.State` MC provides (via BufferBuilder.getState/restoreState)?

src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
```diff
@@ -65,99 +76,108 @@ object HologramRenderer extends TileEntitySpecialRenderer[Hologram] with Callabl
...
-  override def render(hologram: Hologram, x: Double, y: Double, z: Double, f: Float, damage: Int, alpha: Float) {
-    if (failed) {
-      HologramRendererFallback.render(hologram, x, y, z, f, damage, alpha)
-      return
-    }
+  private val renderQueue = new ArrayDeque[Hologram]
 
-    this.hologram = hologram
-    RenderState.checkError(getClass.getName + ".render: entering (aka: wasntme)")
+  // Defer actual rendering until now so transparent things render correctly.
```
fnuecke, line 81: Could you elaborate on this a bit?

Are BlockEntities not depth-sorted before rendering?

Does this still work correctly if a transparent block or non-hologram BlockEntity is visually in front of the hologram?

src/main/resources/data/opencomputers/tags/blocks/adapter.json
```diff
@@ -0,0 +1,6 @@
+{
```
fnuecke, line 1: Does it make sense for each block and item to have its own tag that just contains that single block/item?
Wouldn't some classifying tags make more sense, such as "upgrade" for all upgrades, "component" for things that provide components, for example?

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
```diff
@@ -1,133 +1,140 @@
 local keyboard = require("keyboard")
 
-keyboard.keys["1"]           = 0x02
```
asiekierka: This is a very breaking change. To keep compatibility with existing OC1 code, it would be recommended to move this to a translation layer Scala-side, IMO.

KosmosPrime:
> Very very cool, thanks so much for taking the time to do this. As asie said, it's a lot, I'll try to help review this, but it'll take some time.
> 
> Assuming this PR is good (which it looks like from what I've seen so far!): as asie said, the other maintainers will probably stick to pre-1.16. Would you be able to/interested in maintaining the 1.16.5 version going forward?

Sure, I'm willing to continue working on it.

src/main/scala/li/cil/oc/server/machine/luac/LuaStateFactory.scala
KosmosPrime: I've integrated the new lua build pipeline, and made the development environment run from a mod jar (like FancyGradle does in the 1.12 branch) and so a PR is no longer necessary.

asiekierka: Excellent!

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
KosmosPrime: There's a test branch in my repository using the LWJGL codes with software translation, but I'm worried about undocumented key codes (on either side).
The old LWJGL codes seem to match current scancodes mostly, I'd prefer using those but they're not an exact match either.
Also, I've found an issue here in that old LWJGL used to send control characters for some keys (like enter and backspace) and those are necessary at least for the Plan9K OS, so that's another open (and undocumented for the most part) question.
And finally there's the new divide between key events and character events which I'm very wary of bridging.

asiekierka: Regarding plan9k, we are planning to drop bundling it in OpenComputers 1.8.0 regardless. Don't worry too much about it, but it's good to use as a testcase for OC1 API compatibility.

From personal experience developing other things - the divide between key and character events is done because key scancodes (attached to physical key location) != character scancodes (affected by keymap, IME, etc).

I'm not sure how LWJGL 3's specific implementation works, but this might be a good first try:

* For key events, ignore all scancodes which map to *printable* characters (ASCII range 32-126 is a good first start).
* For character events, ignore all events which map to *non-printable* characters (ASCII range 0-31 + 127).
* Create mapping tables from key scancodes to characters where applicable, and from characters to key scancodes where applicable. (See https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua for more information. If some scancode which is emitted isn't there, it might be a good idea to PR for OC 1.8.0 regardless). 
* Tentatively ignore character events with characters outside the 32-126 range. Add a TODO comment to examine interactions with IMEs (for, say, typing Japanese directly into the OpenComputers machine). This is best skipped for a first alpha.

As for undocumented keycodes - remember: OC1 has to try its best to keep compatibility with the existing base of OC1 programs, especially given that OC2 technically exists. However, I don't see a problem with fixing compatibility issues "as they're discovered"... Ideally, *ignore* any and all keycodes/characters/scancodes/etc. which you don't know how to map - this will prevent people writing broken code.

asiekierka:
Personally speaking, I'm willing to merge this as soon as:

* OC1 API compatibility is restored (even if a bit rough around the edges),
* @payonel and @fnuecke approve the changes.

Not all bugs need to be fixed, just getting a beta release for 1.16.5 out would be magnificent.

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
SquidDev: > I'm not sure how LWJGL 3's specific implementation works, but this might be a good first try:

LGJGL 3 just uses [GLFW key codes](https://www.glfw.org/docs/latest/group__keys.html), so it should be possible to make an exhaustive list and map between the two.

Afraid I didn't bother to do this for ComputerCraft or I'd share mine. Sorry!

src/main/scala/li/cil/oc/common/nanomachines/ControllerImpl.scala
```diff
@@ -340,33 +341,33 @@ class ControllerImpl(val player: EntityPlayer) extends Controller with WirelessE
...
 
   private def isServer = !isClient
 
   private def cleanActiveBehaviors(reason: DisableReason): Unit = {
     if (activeBehaviorsDirty) {
       configuration.synchronized(if (activeBehaviorsDirty) {
         val newBehaviors = configuration.behaviors.filter(_.isActive).map(_.behavior)
-        val addedBehaviors = newBehaviors -- activeBehaviors
-        val removedBehaviors = activeBehaviors -- newBehaviors
+        val addedBehaviors = newBehaviors.clone --= activeBehaviors
```
fnuecke, line 369: It's been a while since I used Scala, this reads like it'd also update the cloned list? Why is this necessary? Or if my guess is wrong, what does this do?

src/main/scala/li/cil/oc/integration/waila/ModWaila.scala
```diff
@@ -2,12 +2,10 @@ package li.cil.oc.integration.waila
...
-import net.minecraftforge.fml.common.event.FMLInterModComms
+import net.minecraftforge.fml.InterModComms
 
 object ModWaila extends ModProxy {
   override def getMod = Mods.Waila
 
-  override def initialize() {
-    FMLInterModComms.sendMessage(Mods.IDs.Waila, "register", "li.cil.oc.integration.waila.BlockDataProvider.init")
-  }
+  override def initialize() = ()
```
fnuecke: Can this class be removed since this is just a no-op now?

src/main/scala/li/cil/oc/server/fs/FileSystem.scala
```diff
@@ -57,50 +60,27 @@ object FileSystem extends api.detail.FileSystemAPI {
...
-    }
-    val file = url.map(url => new io.File(url.toURI)).recoverWith {
-      case _: URISyntaxException => url.map(url => new io.File(url.getPath))
-    }.getOrElse(new io.File(codeSource))
+  override def fromResource(loc: ResourceLocation): api.fs.FileSystem = {
+    val innerPath = "/assets/" + loc.getNamespace + "/" + (loc.getPath.trim + "/")
 
-    if (isArchive) {
+    val modInfo = FMLLoader.getLoadingModList().getModFileById(loc.getNamespace)
+    val file = modInfo.getFile().getFilePath().toFile()
```
fnuecke, line 67: Does this work for both assets in JARs and not just for things in the assets dir on the host FS (as it did before)?

src/main/java/li/cil/oc/api/network/SimpleComponent.java
```diff
@@ -1,114 +0,0 @@
...
- *             if ("greet".equals(method)) {
- *                 return new Object[]{String.format("Hello, %s!", args.checkString(0))};
- *             } else {
- *                 throw new NoSuchMethodException();
- *             }
- *         }
- *     }
- * </pre>
- */
-public interface SimpleComponent {
```
fnuecke, line 89: I might be misremembering, but I thought this was the preferred way by many when adding integrations.

Is it too complex to add with the new JavaScript based core modding system? Is this something you plan on re-introducing in a follow-up PR?

src/main/resources/assets/opencomputers/sounds/preload.cfg
```diff
@@ -1,16 +0,0 @@
-assets/opencomputers/sounds/computer_running.ogg
```
fnuecke, line 1: Is this no longer needed? Or is it no longer possible? IIRC there were very noticeable freezes when sounds played the first time, hence preloading them once at startup.

src/main/scala/li/cil/oc/common/command/SimpleCommand.scala
```diff
@@ -1,23 +0,0 @@
...
-
-import net.minecraft.command.CommandBase
-import net.minecraft.command.ICommandSender
-import net.minecraft.server.MinecraftServer
-import net.minecraftforge.fml.common.FMLCommonHandler
-
-import scala.collection.convert.WrapAsJava._
-import scala.collection.mutable
-
-abstract class SimpleCommand(val name: String) extends CommandBase {
```
fnuecke, line 13: Any reason all the commands were removed? It surely is still possible to register custom commands?

These were always quite useful when testing, in particular.

src/main/scala/li/cil/oc/server/component/UpgradeChunkloader.scala
```diff
@@ -142,7 +151,8 @@ class UpgradeChunkloader(val host: EnvironmentHost) extends AbstractManagedEnvir
         throw new Exception("this dimension is blacklisted")
       }
     } else {
-      ticket = Option(ForgeChunkManager.requestTicket(OpenComputers, host.world, ForgeChunkManager.Type.NORMAL))
+      // This ticket is a lie, but ChunkloaderUpgradeHandler won't crash or load it.
+      ticket = Some(new ChunkPos(0, 0))
```
fnuecke, line 155: Does this mean this upgrade is currently not functioning?

fnuecke (commented):
Firstly, hats off, very impressed you powered through all this! Thanks again for putting in the effort.

Just a couple of minor suggestions/questions, very few major points (mod id, commands removal, simple component removal).

Did not yet get around to actually running it, yet. Might not get to that for a few more days... my internet is currently down, and the earliest the ISP "can" send someone to check is Wednesday... So I'm using my phone's hotspot in the meantime, but I have a pretty tight data limit there. So downloading all those libraries is something I'll avoid until my regular internet is back. 😅

@asiekierka, @payonel, do you know if there were any changes to the 1.12 branch in the meantime that should also be integrated in this PR? I imagine a full rebase will be quite painful, so if there's nothing major I'm considering creating the 1.16 branch this PR will target from the 1.12 state back when this was started. Thoughts?

---

Side note, probably not for this PR, but maybe interesting for future maintenance effort reduction: I pulled out the in-game manual into [a separate mod](https://github.com/MightyPirates/MarkdownManual) for TIS-3D and OC2, might be worth looking into using this here, too. I saw the change from depth buffer to scissors rect, which is a great idea. I'll try to get that into that library mod as soon as possible. If there's any other changes (unrelated to the port) that I missed in this area, please let me know, thanks!

build.gradle
KosmosPrime: In an effort to keep the specification from changing I've used strings here, but variables would work just as well.

src/main/resources/assets/opencomputers/bcbuilder/blocks/adapter.json
KosmosPrime: When updating dependencies I couldn't find a Buildcraft version for MC 1.16 so I removed this part as well so as to not leave it lingering. It can probably be re-added but I'd rather know what I'm ensuring compatibility with before I do.

src/main/resources/META-INF/mods.toml
KosmosPrime: To my knowledge, the mod ID has always been lowercase (see [mcmod.info](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.12/src/main/resources/mcmod.info#L2)).

src/main/scala/li/cil/oc/server/fs/FileSystem.scala
KosmosPrime: Yes, I can load OpenOS both in standalone client/server with mod jar as well as dev (which used to give me the resources path). In fact, I've had to use resource locations explicitly because in standalone the old classpath-based method got broken by the new `modjar` URL scheme (which lack a `CodeSource`). I'd like to eventually use resource loaders instead of relying on disk reads, but it works fine for now.

src/main/java/li/cil/oc/api/network/SimpleComponent.java
KosmosPrime: I'm open to reintroduce the SimpleComponent in the future, indeed the removal of the stubs was probably rushed as, even without class transformer support (which I have no experience with), modders could call them manually.

src/main/scala/li/cil/oc/common/command/SimpleCommand.scala
KosmosPrime: Commands were changed significantly and I simply didn't see them as a priority this early on, but I'll add them back if they're still necessary.

src/main/scala/li/cil/oc/server/component/UpgradeChunkloader.scala
KosmosPrime: It worked in the limited testing I've done for it. The method immediately below turns it into a "real" ticket - I'm going to use vanilla's new chunk ticket system in the future, because I can have a single ticket load a 3x3 with minimal hassle.

build.gradle
asiekierka: :+1:

src/main/resources/assets/opencomputers/bcbuilder/blocks/adapter.json
asiekierka: BuildCraft is extremely unlikely to come back to Forge. It's acceptable to remove integration here.

src/main/resources/META-INF/mods.toml
asiekierka: It wasn't on 1.7.10. It was changed [in 1.11](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.11/src/main/scala/li/cil/oc/OpenComputers.scala#L20). This is okay.

src/main/scala/li/cil/oc/integration/waila/ModWaila.scala
asiekierka: WTHIT exists, so I'd say either put a TODO to restore integration, or remove the class if integration is no longer necessary.

src/main/java/li/cil/oc/api/network/SimpleComponent.java
asiekierka: As far as I know, JavaScript-based coremods don't allow you to target arbitrary classes, only a pre-determined list. If that is the case, a new approach would be necessary. We have Capabilities now (if Forge didn't remove them), so that might be the right way to go...

src/main/scala/li/cil/oc/common/command/SimpleCommand.scala
asiekierka: For anything that's removed since the 1.12.2 port, please add a TODO - in the form of an open issue (once this is merged).

src/main/scala/li/cil/oc/server/component/UpgradeChunkloader.scala
asiekierka: Hacks for the initial port beta are fine, but please add a TODO or open issue.

asiekierka:
Regarding the new in-game manual: I would absolutely move over to it, especially as the pre-1.16 manual [has known bugs](https://github.com/MightyPirates/OpenComputers/issues/3560). This does not have to be part of the initial PR, however.

asiekierka:
Regarding 1.12: The changes made in master-MC1.12 are significant in my opinion, and I would still wish to continue maintaining 1.7.10 if something is reported back from GT:New Horizons - they're still contributing features from 1.7.10. Therefore, I do expect Kosmos to be able to merge in code from the MC1.12 branch from time to time - however, this does not need to be part of the initial release or PR; it should be done soon after, though.

I do not foresee myself or GT:NH doing additional major changes that don't involve mod compatibility, however; I just don't have the time, and GT:NH primarily focuses on mod compat - and the number of mods which have a remotely similar API in 1.7.10 and 1.16.5 is vanishingly small :-)

build.gradle
fnuecke: Ah, sure. Though the package name/group should similarly be non-changing, right? So I'd either inline both or neither, but up to you :)

src/main/resources/assets/opencomputers/bcbuilder/blocks/adapter.json
fnuecke: Ahhh, I was just aware that it was still maintained, but missed the no longer on Forge part. Got it!

src/main/scala/li/cil/oc/integration/waila/ModWaila.scala
fnuecke: Isn't the integration still working via the [plugin annotation in the other class](https://github.com/KosmosPrime/OpenComputers/blob/master-MC1.16/src/main/scala/li/cil/oc/integration/waila/BlockDataProvider.scala#L29)?

src/main/scala/li/cil/oc/server/fs/FileSystem.scala
fnuecke: Awesome, thanks for clarifying :)

src/main/java/li/cil/oc/api/network/SimpleComponent.java
fnuecke: Ah, I see, yeah, that makes it more complicated/a bigger change than I expected. Switching this over to (potentially event injected) capabilities sounds like a good new approach in a next step.

src/main/scala/li/cil/oc/common/command/SimpleCommand.scala
fnuecke: Hmm, yeah, I'd say it'd be good to get these back, but can be as a follow up. As @asiekierka said, tracking the things that got removed for the initial port but should be re-added would be very valuable. If you have a list in your head already, maybe already add it to the description of the PR for clarity, thanks!

src/main/scala/li/cil/oc/server/component/UpgradeChunkloader.scala
fnuecke: Ah, I see. Cool. Yeah, ditto then :)

fnuecke:
Re branch: got it. [Created a branch this can target](https://github.com/MightyPirates/OpenComputers/tree/master-MC1.16) based on the last commit before this PR for now then. Can be updated as a follow up.

Edit: fixed branch name.

src/main/resources/META-INF/mods.toml
fnuecke: Ahh, wrong assumption then, my bad. So there was some legacy `oc` usage in OreDict logic, which now caused the `oc` -> `opencomputers` change in the in-game manual?

asiekierka: Yes, that's exactly what happened - the ore dictionary used `oc:` as a prefix.

src/main/scala/li/cil/oc/client/renderer/font/StaticFontRenderer.scala
```diff
@@ -16,7 +21,7 @@ import scala.io.Source
  */
 class StaticFontRenderer extends TextureFontRenderer {
```
asiekierka, line 22: I recommend to drop this functionality right here. It's a pointless maintenance burden which I planned I swear nobody really uses it anymore - it's all the Unifont .HEX file.

asiekierka: There's an old VBO rewrite here for 1.7, but the code is not very good. Either way, it might give you some inspiration: https://github.com/MightyPirates/OpenComputers/pull/3138

src/main/resources/data/opencomputers/tags/blocks/adapter.json
KosmosPrime: I was looking to bring this up eventually, as I'd prefer to make the current `mayPlace` depend on tags (which are synchronized) as opposed to the existence of a driver. The reason for almost all of these tags existing is that they do in OC (and were actually used for recipes most of the time).

src/main/scala/li/cil/oc/common/nanomachines/ControllerImpl.scala
KosmosPrime: It's because scala 2.13.4 doesn't have the difference method `--` anymore, this creates a new list and subtracts the `activeBehaviors` from that in-place. The next line would probably still work but I didn't notice that it uses a Set at the time (nevertheless removing an `IterableOnce` from a mutable `Set` is already deprecated).

src/main/scala/li/cil/oc/integration/waila/ModWaila.scala
KosmosPrime: WAILA integration is working (except drones), indeed it currently uses an annotation-based plugin system (and thus doesn't need to be initialized anymore, I don't think the mod class is required anymore).

asiekierka: If it's not required, I recommend removing it.

src/main/resources/data/opencomputers/tags/blocks/adapter.json
fnuecke: Ah, got it. Can be a follow up then 👍🏻

src/main/scala/li/cil/oc/common/nanomachines/ControllerImpl.scala
fnuecke: Got it, thanks for the explanation.

src/main/resources/assets/opencomputers/sounds/preload.cfg
KosmosPrime: Preloading is now a feature in `sounds.json`, apparently I forgot to add it there. I didn't notice any freezes though. Was the intent to buffer the file in memory, or only metadata?

asiekierka: Given that this stuff was worked on almost 10 years ago, you might need to try running from an HDD to find out ^^ - I'd say it's safe to remove until someone complains again.

fnuecke: To answer your question: it was to load the actual audio data into memory ahead of time.

SSDs/M.2 being rather common now is a good point though. Probably makes this a non-issue nowadays anyway. Feel free to resolve.

src/main/scala/li/cil/oc/client/renderer/font/StaticFontRenderer.scala
KosmosPrime: I'll probably remove it in the same pass as converting to the dedicated manual mod then, it appears to function for now.
I've already implemented a buffered approach for screens using MC's vertex buffers to properly support the `MatrixStack`.

asiekierka: Excellent!

src/main/scala/li/cil/oc/client/KeyBindings.scala
KosmosPrime: The intent was that it determines whether to copy the address as you analyze, but I don't feel strongly about it.

src/main/scala/li/cil/oc/client/PacketHandler.scala
KosmosPrime: I believe this is necessary as it calls `setClipboard`, which then calls the GLFW method (which probably requires the GLFW context). I'm not sure if `context.enqueueWork` guarantees this.

src/main/scala/li/cil/oc/client/gui/widget/Widget.scala
KosmosPrime: Vanilla minecraft now has a `net.minecraft.client.gui.widget.Widget` which is supported by all GUIs. Because there's only 1 widget class I'd like to part that over to avoid maintaining this.

src/main/scala/li/cil/oc/client/PacketHandler.scala
asiekierka: Agreed.

src/main/scala/li/cil/oc/client/renderer/RenderCache.java
KosmosPrime: I was originally merging different render calls  of the same `RenderType` but had to abandon because of Z-fighting issues. Currently it's indeed much the same as a list of `BufferBuilder.State` so I might replace it with that.

src/main/scala/li/cil/oc/client/KeyBindings.scala
fnuecke: Ah, as in referring to the operation. Sure. Me neither :D

src/main/scala/li/cil/oc/client/PacketHandler.scala
fnuecke: Ah, got it. Thanks for the explanation!

src/main/scala/li/cil/oc/client/gui/widget/Widget.scala
fnuecke: Oh, very nice, got it.

src/main/scala/li/cil/oc/client/renderer/RenderCache.java
fnuecke: Got it, just wanted to check. Up to you if/when to change it.

src/main/scala/li/cil/oc/client/renderer/HighlightRenderer.scala
KosmosPrime: Yes, block highlights (even for prints) is now vanilla MC based on `Block.getShape`.

src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
KosmosPrime: The problem was that clouds and water rendered in front of holograms partially. I've tested this in 1.12 and it appears to suffer from similar issues, but only for transparent blocks. The problem is that depth-sorting ignores the current depth value in the buffer (which this renderer relies on).
Sadly no, this does get holograms occluded by transparent blocks in front of them. However, I believe making this work requires at least depth-sorting all the hologram quads every frame which ruins caching and thus performance. I have some ideas, but I've yet to figure out how to inject quads into the sorted queue while allowing for rotation.

src/main/scala/li/cil/oc/client/renderer/HighlightRenderer.scala
fnuecke: Ah, very nice, thanks for clarifying.

src/main/scala/li/cil/oc/client/renderer/tileentity/HologramRenderer.scala
fnuecke: Right. Fair to keep it like this for now then.

Might be an interesting experiment to try rendering the hologram itself as part of regular MC block rendering (not via BlockEntity rendering). IIRC MC does depth sorting for transparent blocks/chunks anyway, updating the sorting whenever the player/eye pos moved a significant amount.

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
KosmosPrime: Hmm, I'm reading the LWJGL 2 code currently. At least on windows, it seems to be unable to report a non-documented keycode. In addition, it appears to be doing the exact "correlate key and character events" wizardry I was afraid of implementing myself. I'll have to look into this more, and whether it works for other OSes too (and how to avoid stuck keys).

Regarding mapping scancodes to characters and back, I don't think that's feasible. I'd have to know the keyboard layout (every keyboard layout) to be able to do this correctly, just like a physical OS would. And I don't have the keyboards to test it.

asiekierka: Tricky.

I do agree that separating key and text input events OC-side going forward would be a good idea; possibly by introducing two new events and deprecating the old one, then emulating the new events in 1.7 and 1.12 as well. This is because once you add IMEs to the mix, there is no longer a correlation between text input and keypresses (LWJGL2 just didn't support IMEs in the first place).

However, what would be the best way to do that while minimizing impact to legacy code? My proposal: We keep treating key_down/key_up as scancode events, with a scancode-derived QWERTY-assuming character value for "char" to provide at least a pretense of legacy support. We also add a text_input event. On 1.7/1.12, key_down also implies emitting a matching text_input event. This way, code which targets OC 1.8.0+ can simply listen to `text_input` events - across all supported MC versions - rather than `key_down` events. (And you could probably easily write a Lua-side compatibility shim with `computer.pushSignal` for servers stuck on older OC versions.)

src/main/resources/assets/opencomputers/sounds/preload.cfg
KosmosPrime: I've gone to test this on my HDD before realizing that windows probably cached the entire folder in memory. Have fixed this in `sounds.json`.

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
asiekierka: I talked to payonel, and he seems to agree that a new `text_input` event + a non-ideal compatibility layer for `key_down` would be the best idea. See [this](https://paste.asie.pl/KPtr) for a simple patch to emulate `text_input` on < 1.13. Do you think that would be sufficient?

```
12:49 <+asie> on 1.7-1.12 we just emulate an additional event, LWJGL2 doesn't let us do much more
12:49 <+asie> on 1.16+ we emit text_input calls for Unicode codepoints, and key_down calls for scancodes - with the 'char' field populated as a best guess to preserve *some* compat with existing code
```

KosmosPrime: I'm in favor of a `text_input` and GLFW conveniently uses ASCII codes for printable keys (32-126) which I can enhance with a crude shift-to-uppercase check. It'll still be wrong on non-US keyboards though. Would it be OK if I continue firing 2 `key_down` events for characters, one with only the key and one with only the character (in addition to the new `text_input`)?

asiekierka: My opinion is that `key_down` should *only* react to scancode events (and translate them to characters for some subset of keys), while `text_input` should *only* react to character events. The idea is that code which uses only `key_down` will work well enough for the time being, and eventually software can be migrated to use `text_input` for printables.

asiekierka: Now, if you really want to go the extra mile, you could make character events in the 32-126 range emulate `key_down` calls on top of emitting `text_input`, while scancode events which result in a 32-126 range character are ignored. This will solve most of the keymap problem.

fnuecke:
Oh, one more thing: I'm fine with the rest of the changes in 1.12 since back then coming later, but could you please pick over the Github workflows/actions for PR validation and publishing? (i.e. the contents of `/.github/workflows/` on current 1.12)

Thanks.

fnuecke (approved):
My internet graciously sporadically came back, so I've been able to test it locally.

Looks great for the most part! Didn't test *everything* because, well, there's a lot, but just a few things I've noticed, in case they're not already on your list of known issues:

- Tablets seem to have some issues:
    - console output doesn't seem to work immediately -- at least the creative one. Seems to boot audio-wise, but no console output being shown, does work after quitting and reloading.
    - Does not seem to properly persist state, always boots up fresh when opening up after quitting and reloading.
- Robots don't seem to properly use tools (`PlayerInteractionManagerHelper.onBlockClicked` seems to always fail).
- Net splitter side configuration does not visually update adjacent cable (functionally it does).
- For some reason I don't get a runServer gradle task, so I couldn't test behavior on a dedicated server.
- Robots are transparent in the middle when running (pretty sure they shouldn't be? Could be misremembering, but looks like a bug). See screenshot: 
- These warnings on startup, but might be those were already present in 1.12? Dunno.
  ```
  [23:25:06] [Render thread/WARN] [OpenComputers/]: Invalid glyph char code detected in font.hex. Expected 0<= charCode <10000, got: 1D300
  [23:25:06] [Render thread/WARN] [OpenComputers/]: 87 total invalid glyph char codes detected in font.hex
  ```

So all in all I'd say no massive blockers for this PR, stuff that can be fixed later, and in the meantime maybe an alpha/beta/pre-release/whatever release can be made to get some more eyes on it.

Thanks again!

asiekierka:
```
[23:25:06] [Render thread/WARN] [OpenComputers/]: Invalid glyph char code detected in font.hex. Expected 0<= charCode <10000, got: 1D300
[23:25:06] [Render thread/WARN] [OpenComputers/]: 87 total invalid glyph char codes detected in font.hex
```
This is fine. OpenComputers got font upgrades without getting patches to support Unicode codepoints outside the basic multilingual plane (BMP) - OC 1.8.0 has the necessary patches (they required reworking a bunch of Unicode logic and TextBuffer code, so they were kept for a new "big" release instead of an OC 1.7.x patch release), but Kosmos didn't pull them in yet.

Either way, personally, I'd say the keyboard thing remains a serious blocker - we don't want people to start writing incompatible OC1(1.16.5) code that doesn't work with OC1(1.7.10) or OC1(1.12.2), where possible to avoid. However, that remains the only merge blocker in my opinion.

KosmosPrime:
- Creative tablets' screens don't render, but the rest are fine. I'm guessing the client giving it the wrong buffer ID somehow but I'm tracking that.
- As far as I know tablets turn off when they pass through being serialized (e.g. save game, get put into a chest for too long). I'd prefer them not do that but it appears consistent with 1.12 behavior.
- Noted, I planned to use the official FakePlayer for them but should at least get placing & mining blocks working.
- Noted, apparently toggling faces doesn't send the update required for cables to update their block state.
- I tested on a 'real' server and forgot to add it, should be a simple fix. I should eventually look into the data generator task also.
- Robots are indeed transparent even though the black square in the center makes it seem like they shouldn't be.

I agree that the keyboard input issue is one we need to resolve first (especially because I haven't merged the code which translates to LWJGL 2 codes yet).

asiekierka:
There's some fixes for tablets in the later commits, FYI.

fnuecke:
FTR, the robot render thing: if it's also like this in 1.12, I'm guessing the render mode of the "running effect" changed from "emissive opaque" to "emissive translucent" at some point before. Possibly during a previous port. Originally it wasn't see-through. E.g. see [the trailer](https://youtu.be/pbETzD-tuQo?t=27)... has it really been 8 years? Damn.

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
KosmosPrime: I've looked into manual translation to/from characters again and I'm even more convinced that it's not possible without a key map (GLFW doesn't provide one) as it becomes impossible to type symbols which requires a modifier key, even on US keyboards. Instead I've rewritten the system to wait for character events on all keys, joining the first character with the most recent key (following characters are dropped, as in 1.12.2). Finally, during a GUI tick the client sends key events that weren't associated with any character (doing this after events are polled would be nice, but there's no Forge hook for it).
Additionally, Minecraft's GUIs can only take a char (not a code point) but `KeyboardListener` splits them up correctly, so I've taken the liberty of rejoining high & low surrogate pairs similar to keys & characters (for `text_input` only, surrogates don't get delivered for `key_down` and `key_up` at all). The reason for doing this is that in UTF-8, non-BMP is encoded differently from the concatenation of a high and low surrogate - the latter isn't even valid UTF-8. Note that I do not have an IME to test this with non-BMP codes.

Edit: Made a tracking pull request KosmosPrime/OpenComputers#1

KosmosPrime:
I've made a tracking pull request for the keyboards feature. With the simpler changes already in, KosmosPrime/OpenComputers#1 should be the only blocking issue (for the initial pull anyway).

src/main/resources/assets/opencomputers/loot/openos/lib/core/full_keyboard.lua
asiekierka: This is all very good. I wasn't worried about symbols which require a modifier key, as I'd expect common applications (like `edit` and the shell) to be patched to properly support `text_input` - however, I'm impressed that you went the full mile to support it. Note that until you pull in the latest `MC1.12` changes, OpenComputers will *not* correctly parse non-BMP key codes in all cases.

asiekierka: I've briefly reviewed the PR, and I like it - my only concern is that whether or not it is possible for the key and char diverge significantly (say, is it possible for me to press 'A' and 'B' simultaneously in such a way that the char for 'A' and the scancode for 'B' gets emitted, or vice versa?), and how rare such an occurence would be.

KosmosPrime: I don't think that's an issue, key and char callbacks are only called while the main thread runs `GLFW.glfwPollEvents` and are executed in whichever order the OS delivers them. Specifically (on windows), when you [translate](https://learn.microsoft.com/en-us/windows/win32/inputdev/using-keyboard-input#translating-character-messages) a key code into a character, that event is always delivered next so unless GLFW mixes them it's safe to correlate them.
What does happen (even in LWJGL 2 / OC 1.12.2) is that pressing an accent key (which doesn't generate a character immediately) followed by a non-modifier key will result in the accent replacing whatever character you would have gotten. This results in surreal combinations where the enter `key_down` event has the right key code, but the accent character. Same method, same... feature I guess.

asiekierka: Okay. In that case, I think we're good.

KosmosPrime:
The keyboard changes are now in, so everything's OK from my side.

Regarding the changes since I started working on this - should I prefer to cherry-pick (and edit) commits (as I have with the workflow stuff) or merge them?

asiekierka:
Merge. In general, changes in OpenComputers are applied from lowest applicable version to the most recent supported, so you want to be up to speed for future merges as well.

Anyhow, this is almost good for me! The only issue is the small smidgen of Lua changes - we'd want to merge them from 1.7.10 upwards if they are applicable to older versions, or get rid of them altogether if they're no longer necessary.

kraetzin:
I've been trying out builds of this PR. I noticed that there is no LuaBIOS EEPROM (item or recipe), and the same is true of the different floppies such as the OpenOS one. Is this supposed to be the case?

KosmosPrime:
> I've been trying out builds of this PR. I noticed that there is no LuaBIOS EEPROM (item or recipe), and the same is true of the different floppies such as the OpenOS one. Is this supposed to be the case?

I've just tested with the most recent automatic build, both Lua BIOS and OpenOS are present and working (though it looks like the recipe for the Lua BIOS EEPROM doesn't).
Could you elaborate on which build you used and how you've found these things to be missing (and whether other parts of the mod do exist)?

kraetzin:
>  Could you elaborate on which build you used and how you've found these things to be missing (and whether other parts of the mod do exist)?

I am currently using the build from here: https://github.com/MightyPirates/OpenComputers/actions/runs/3406556101

I have had the same issue with running the build locally on [abfd464](https://github.com/MightyPirates/OpenComputers/pull/3572/commits/abfd464805412d33245adcbab1fa39d279bd72a0).

I'm running it in a fresh instance of 1.16.5 in PolyMC with Forge version 36.2.34, with Scorge 3.1.3 also installed.

Neither the LuaBIOS EEPROM variant or floppy variants show up. I do have the base eeprom and base floppy items showing:







Is there something that I might be missing?

Edit:

Just to add, I also noticed these errors in the log:

```
[18:42:03] [Render thread/WARN] [OpenComputers/]: Attempt to get ItemStack for block Block{opencomputers:robotafterimage} without item form 
[18:42:05] [Worker-Main-13/WARN] [minecraft/ModelBakery]: Unable to load model: 'opencomputers:floppy#inventory' referenced from: opencomputers:floppy#inventory: java.io.FileNotFoundException: opencomputers:models/item/floppy.json 
```

KosmosPrime:
Looks like they don't appear in the search tab, but do in the OC tab. I believe Minecraft builds the list to search before the server has a chance to send floppies to the client, but the EEPROM should be present from the start.
Regarding the second error, I'm not sure why the game insists on loading that but it's never used, floppies exclusively use colored models.

kraetzin:
Understood, sorry for the confusion, I should have looked in the OpenComputers tab! I can confirm that everything is there as it should be. When I couldn't craft the BIOS, and it wasn't showing up in search like it used to, it threw me off a bit.

asiekierka:
Don't worry about either of those two issues, for the time being. The only objection I have is ensuring any changes to the Lua code (OpenOS etc.) are either undone or propagated from 1.7.10 onwards.

KosmosPrime:
There's currently 4 lua changes relative to the branch point:
- A change in data's `nc.lua` which stops it from sending unprintable keys. It used to be more critical when pressing printable keys generated an extra event without a character attached, but the original code still causes NULs to be sent.
- Two cases related to hard-coded key values to use constants instead. This is no longer relevant since we're using old key codes again.
- Part of the Lua upgrade commmit which I've included (the version number change) because I included the new library build pipeline but not the Lua 5.4 integration (yet). The 1.16 branch matches the 1.12 branch as it is now though, so safe to leave I think.

asiekierka:
In that case, I'd do the following:

* split out the `nc.lua` change into a separate PR, targetting `master-MC1.7.10`, it will make its way to `master-MC1.16` in due time;
* whether or not you include the hard-coded key values to constants change is up to you, it's a code clarity thing I guess;
* the Lua copyright date change can stay, it will fix itself in due time as above.

asiekierka:
Thank you!

I'm waiting on payonel to analyze the `text_input` event addition in light of OpenOS's architecture; he'll do his best to get it done this weekend. Once that is done (and the `text_input` event is added to 1.7.10/1.12.2 branches), it's a merge.

th3impasta:
small issue in build https://github.com/MightyPirates/OpenComputers/actions/runs/3406556101
robot names dont show on the one probe

payonel:
@KosmosPrime i'm really excited about this work. thank you
a couple of things i found
1. I noticed while typing in the shell (testing in openos) that some keys are not printed. I found that spamming a key sometimes sends char:0 instead of the correct char. Example, 'a' should send '97 30', but sometimes sends '0 30' (note the code is always correct. I noticed this char:0 only happens for keys that also signal text_input. char:0 is sent about 20% of the time
2. it seems no text commands are registered. is this a known issue?

KosmosPrime:
1. I can't replicate this behavior (on windows), so I assume you're on unix, and I don't have a machine around to test it. In any case, my best guess (if the `text_input` event appears still) is that the GLFW character event gets delivered a frame too late and the code fails to correlate them (this is why I've advocated for the `text_input` event in the first place).
- I noticed while testing this that I accidentally broke key repeat event handling (fixed on catchup branch).
2. I've not looked into commands yet, they didn't seem important at the time while requiring a significant rework for 1.16.5's autocomplete system. I'll definitely make a follow up including those.

payonel:
yeah i was going to mention the repeat bug too. glad you noticed

yes i'm on linux, i'll retest on windows

embeddedt:
I've submitted a small fix to allow using this on newer versions of Java (i.e. 11). https://github.com/KosmosPrime/OpenComputers/pull/2

embeddedt:
> I noticed while testing this that I accidentally broke key repeat event handling (fixed on catchup branch).

Keys still seem not to repeat properly on `catchup-MC1.16`. When using the `edit` program I was unable to hold down an arrow key to keep scrolling (it only registers as one press).

Is that the best branch to be testing on/working with? And on that note, is there anything I could do to help expedite this?

KosmosPrime:
Apologies, they were on my local branch and I failed to push them.
Currently `catchup-MC1.16` is the main branch, pushing to master messes with this PR so I'm avoiding it until the `text_input` changes have bubbled up. To my knowledge, the latter is the only issue still blocking this PR.

kristibektashi:
> 1. I can't replicate this behavior (on windows), so I assume you're on unix, and I don't have a machine around to test it. In any case, my best guess (if the `text_input` event appears still) is that the GLFW character event gets delivered a frame too late and the code fails to correlate them (this is why I've advocated for the `text_input` event in the first place).
> 
> 
>     * I noticed while testing this that I accidentally broke key repeat event handling (fixed on catchup branch).
> 
> 
>     2. I've not looked into commands yet, they didn't seem important at the time while requiring a significant rework for 1.16.5's autocomplete system. I'll definitely make a follow up including those.

You could download [VirtualBox](https://www.virtualbox.org/) and use it to install a Linux system such as Ubuntu or Linux Mint in a virtual machine. It won't mess with your Windows installation or your files at all

kristibektashi:
Also when will 1.18.2 be supported?

embeddedt:
Someone will need to have time and interest to port it as was done for 1.16. Keep in mind that it often takes quite a lot of effort to port mods of this size. That being said, porting to future versions should hopefully be more straightforward.

KosmosPrime:
I have since reproduced the issue in a Ubuntu VM (was surprised to see it even supporting holograms & shaders) and as expected, the issue is that char events get delivered too late.
However, looking at [the GLFW code](https://github.com/glfw/glfw/blob/master/src/x11_window.c#L1244) I don't see why this happens, the key (except duplicates) and char events get delivered at the same time, and during the `glfwPollEvents` method. I could allow events to appear a tick late but that's hardly a fix: high frame-rate could require a higher delay, which negatively impacts low frame-rate situations.

embeddedt:
Which Minecraft event are you using to get the key inputs? I would like to try this in a standalone mod and see if the same phenomenon occurs. (I use Arch Linux so I should get the same results as Ubuntu.)

KosmosPrime:
> Which Minecraft event are you using to get the key inputs? I would like to try this in a standalone mod and see if the same phenomenon occurs. (I use Arch Linux so I should get the same results as Ubuntu.)

See https://github.com/KosmosPrime/OpenComputers/blob/catchup-MC1.16/src/main/scala/li/cil/oc/client/gui/traits/InputBuffer.scala
The `keyPressed`, `keyReleased` and `charTyped` methods do capturing/bubbling of events, and are fed almost directly by the GLFW callbacks.

embeddedt:
The events seem to be correlated with respect to each time a GUI frame is rendered, at least with this simple test code. (Both `keyPressed` and `charTyped` print the same frame number.) Do you see the same behavior?

```java
import com.mojang.blaze3d.matrix.MatrixStack;
import net.minecraft.client.gui.screen.Screen;
import net.minecraft.util.text.ITextComponent;
import net.minecraft.util.text.StringTextComponent;
import org.lwjgl.glfw.GLFW;

public class TestScreen extends Screen {
    protected TestScreen() {
        super(new StringTextComponent("hello world"));
    }

    public int frame = 0;

    @Override
    public void render(MatrixStack p_230430_1_, int p_230430_2_, int p_230430_3_, float p_230430_4_) {
        super.render(p_230430_1_, p_230430_2_, p_230430_3_, p_230430_4_);
        this.frame++;
    }

    @Override
    public boolean charTyped(char ch, int p_231042_2_) {
        System.out.println("frame " + frame + " char " + ch);
        return false;
    }

    @Override
    public boolean keyPressed(int pKeycode, int pScancode, int pModifiers) {
        System.out.println("frame " + frame + " keycode " + GLFW.glfwGetKeyName(pKeycode, pScancode));
        return false;
    }
}
```

Output:

```
[18:49:00] [Render thread/INFO] [STDOUT/]: [TestScreen:keyPressed:30]: frame 342 keycode d
[18:49:00] [Render thread/INFO] [STDOUT/]: [TestScreen:charTyped:24]: frame 342 char d
[18:49:01] [Render thread/INFO] [STDOUT/]: [TestScreen:keyPressed:30]: frame 368 keycode d
[18:49:01] [Render thread/INFO] [STDOUT/]: [TestScreen:charTyped:24]: frame 368 char d
```

KosmosPrime:
Not from within the OC branch, no. In fact, using `render` over `tick` makes it much less likely to recognize characters (from 50% to less than 1%).
(Do note that I have to use these separate hooks and generate a single `key_down` event with both key & char.)

kristibektashi:
> I have since reproduced the issue in a Ubuntu VM (was surprised to see it even supporting holograms & shaders) and as expected, the issue is that char events get delivered too late. However, looking at [the GLFW code](https://github.com/glfw/glfw/blob/master/src/x11_window.c#L1244) I don't see why this happens, the key (except duplicates) and char events get delivered at the same time, and during the `glfwPollEvents` method. I could allow events to appear a tick late but that's hardly a fix: high frame-rate could require a higher delay, which negatively impacts low frame-rate situations.

Minecraft on Linux has a bug where key presses are sometimes sent twice, this is also why sometimes the letter 't' (or whatever your keybind for chat is) appears in the chatbox when you open chat. This is a vanilla bug and happens on all versions afaik

embeddedt:
Ah, good point! I assume [MC-122477](https://bugs.mojang.com/browse/MC-122477) is the issue you're referring to? There is a good writeup [here](https://github.com/RecursiveG/Mc122477Fix#what-is-this-bug) about it and it does sound related.

kristibektashi:
Any progress?

kristibektashi:
> Ah, good point! I assume [MC-122477](https://bugs.mojang.com/browse/MC-122477) is the issue you're referring to? There is a good writeup [here](https://github.com/RecursiveG/Mc122477Fix#what-is-this-bug) about it and it does sound related.

Yea that's the one. What is interesting is that it doesn't happen when using [MultiMC](https://multimc.org/) but often happens when using the vanilla launcher

Source: I use Linux and the bug stopped happening when I switched to MultiMC from vanilla launcher

Ryryy700 (approved):

asiekierka:
To be honest, given that the OpenComputers development team has *de facto* disbanded, with me not having been an original member at any time and not feeling confident enough to touch some parts of the codebase... I think your best option might be to make a fork (like CC:Tweaked forked off ComputerCraft).

Lockl00p:
> To be honest, given that the OpenComputers development team has _de facto_ disbanded, with me not having been an original member at any time and not feeling confident enough to touch some parts of the codebase... I think your best option might be to make a fork (like CC:Tweaked forked off ComputerCraft).

I was actually about to say the same thing! You (the pull request author) should make a fork (like maybe OC Rebuilt). If You do, I'll fork that to make a restitched version.

KosmosPrime:
The original intent was to port OC to have something while OC2 isn't ready for production yet. OC2 was more active at the time and I thought a port would be good for the time being. If I'd designed for a fork I would have probably rewritten the mod around some challenging aspects (I'm not very good with scala, IDs and metadata are no longer a thing, library changes, datapack support, etc.). So I settled for a port because I felt that a fork would also imply being maintained parallel to OC2's existence.
As mentioned I'm fine with maintaining the 1.16 branch, maybe later a 1.18/1.19 too, but I'm unsure where to take this from here.

embeddedt:
I'd like to see this get a release even with the keyboard bug still present (as it may not even really be OC's fault). That's infinitely better than no post-1.12 release at all.

alexhorner:
> The original intent was to port OC to have something while OC2 isn't ready for production yet. OC2 was more active at the time and I thought a port would be good for the time being. If I'd designed for a fork I would have probably rewritten the mod around some challenging aspects (I'm not very good with scala, IDs and metadata are no longer a thing, library changes, datapack support, etc.). So I settled for a port because I felt that a fork would also imply being maintained parallel to OC2's existence.
> As mentioned I'm fine with maintaining the 1.16 branch, maybe later a 1.18/1.19 too, but I'm unsure where to take this from here.

OC2 being completely different, it'd be a shame to set OC aside once OC2 becomes production ready, unless of course OC2 suddenly brings in full compatibility with OC however I don't see that happening considering they take two completely different stances on the backend (Linux Vs Lua APIs)

Lockl00p:
> The original intent was to port OC to have something while OC2 isn't ready for production yet. OC2 was more active at the time and I thought a port would be good for the time being. If I'd designed for a fork I would have probably rewritten the mod around some challenging aspects (I'm not very good with scala, IDs and metadata are no longer a thing, library changes, datapack support, etc.). So I settled for a port because I felt that a fork would also imply being maintained parallel to OC2's existence.
> As mentioned I'm fine with maintaining the 1.16 branch, maybe later a 1.18/1.19 too, but I'm unsure where to take this from here.

If you’re planning to set it aside I could attempt to start developing it instead. Maybe add datapack support.

Mai-Lapyst:
> If you’re planning to set it aside I could attempt to start developing it instead. Maybe add datapack support.

I tried to port it to a higher MC version (as well as to java) some time back but I lack experience in Scala and OC's backend code is heavily make use of Scala features; If the port is being made in Java (which would be more of a rewrite than a plain port) I would see If I can help with development.

It would really be a shame to not have the lua / script-based OC available in modern versions of mc.

Lockl00p:
> > If you’re planning to set it aside I could attempt to start developing it instead. Maybe add datapack support.
> 
> I tried to port it to a higher MC version (as well as to java) some time back but I lack experience in Scala and OC's backend code is heavily make use of Scala features; If the port is being made in Java (which would be more of a rewrite than a plain port) I would see If I can help with development.
> 
> It would really be a shame to not have the lua / script-based OC available in modern versions of mc.

Wait it uses scala? Oh god, guess I won’t be porting it.

Piripe:
> I'd like to see this get a release even with the keyboard bug still present (as it may not even really be OC's fault). That's infinitely better than no post-1.12 release at all.

@embeddedt You can download the latest release [here](https://github.com/KosmosPrime/OpenComputers/actions/runs/5179736609).

Lockl00p:
Any attempt to compile it, throws an error, saying
 > Failed to notify project evaluation listener.
   > Cannot mutate content repository descriptor 'BUNDELED_-174096774(file:/C:/Users/me/.gradle/caches/forge_gradle/bundled_deobf_repo)' after repository has been used
   > Could not find method prepareRuns() for arguments [build_dj7izmay26m3csqhkflr0jpl3$_run_closure14$_closure50@164c1d85] on root project 'OpenComputers' of type org.gradle.api.Project.

Was that due to gradle version or something?

asiekierka:
Yes. With ForgeGradle and other Gradle-based modding solutions, you *must* use the same major Gradle version as the project, due to how much they tend to need to hack into the internals of Gradle.

Try using the `./gradlew` script, not your system's Gradle installation.

Lockl00p:
> Yes. With ForgeGradle and other Gradle-based modding solutions, you _must_ use the same major Gradle version as the project, due to how much they tend to need to hack into the internals of Gradle.
> 
> Try using the `./gradlew` script, not your system's Gradle installation.

But I did, and it still threw the error. I ran gradlew and it did not work.

Lockl00p:
> Yes. With ForgeGradle and other Gradle-based modding solutions, you _must_ use the same major Gradle version as the project, due to how much they tend to need to hack into the internals of Gradle.
> 
> Try using the `./gradlew` script, not your system's Gradle installation.

I have found the issue. He was trying to pr his master:1.16.5 so I figured that was where he did his work on. It was on his catchup-1.16.5 branch.

KosmosPrime:
The reason for that is that every time I push to master now, it'll update this PR, requiring a re-run of the CI (and possibly requesting a new review). The catchup branch merges changes from the 1.12 branch since I started the port.
I see that compiling the master branch locally or through github CI does produce this error now... I have no idea what it means though, and it certainly didn't use to do that. It's also not just the gradle version, same happens if I select gradle 7.6.1 instead, but the catchup branch is somehow fine.

kristibektashi:
Any news in this?

finnaminope:
Someone needs to take this on or the mod will die. i dont know any java tho

PublicVoidUpdate:
speak for yourself, i still actively use java 1.12.2 mods, and only play newer versions in modpacks.

PublicVoidUpdate:
also gregtech new horizons is still on 1.7.10 and its not dying anytime soon.

leonpano2006 (approved):

kristibektashi:
@asiekierka @KosmosPrime so the changes got approved. Is this enough to get this PR merged?

KosmosPrime:
> After a pull request is opened, anyone with read access can review and comment on the changes it proposes.

This isn't the first such review and I don't know what motivates it. Github doesn't show it under "Show all reviewers" at the bottom anyway.

Unless this person has write access, #3709 probably still applies.

asiekierka:
I thought I have made it clear, but: with the retirement of all of the project's lead developers, this PR will not be merged, because there's nobody to maintain it regardless. For minor fixes, at least I or someone else can step in to handle the existing codebase, but this is a relatively large change which would require us to re-learn the codebase.

I continue to encourage KosmosPrime to start a fork to carry it on. Call it LibreComputers or something like that. I will do my best to point people to it if it proves itself to be reliable.

GriffinSego (approved):

leonpano2006 (approved):

finnaminope:
I Am Thinking That Maybe We Could Make An OpenComputers Like Mod For Luanti. Similar Mods Already Exist So We May Be Able To Make It Work. I Would Also Want It To Be Mostly Compatible With The Existing OpenComputers Lua API.

uis246:
> I continue to encourage KosmosPrime to start a fork to carry it on. Call it LibreComputers or something like that. I will do my best to point people to it if it proves itself to be reliable.

Maybe upstream it, but separate maintainers for new versions? I mean something similar to how linux-next and linux LTS kernels work, where first is maintained by Linus, while older versions by Greg. Here it would be old team does versions up to 1.12.2 and new team does higher versions.

kristibektashi:
> > I continue to encourage KosmosPrime to start a fork to carry it on. Call it LibreComputers or something like that. I will do my best to point people to it if it proves itself to be reliable.
> 
> Maybe upstream it, but separate maintainers for new versions? I mean something similar to how linux-next and linux LTS kernels work, where first is maintained by Linus, while older versions by Greg. Here it would be old team does versions up to 1.12.2 and new team does higher versions.

My understanding is that the older team is gone, and they may not want to give control to the repo to someone else. Forking is probably the best idea, for now at least.

PR #3577, opened 2022-11-16
Fix typo

Philippus:


asiekierka:
I mean... it's only a comment. But it is a fix!

PR #3582, opened 2022-11-23
Faster Computer Boot

igorkll:
playing strings is a non-blocking operation, and the string "." plays like a computer.beep(1000, 0.2)


igorkll:
I'm a little confused in the pull requests, sorry, there's nothing there except the updated bios

asiekierka:
@payonel Is there a good reason to have blocking beeps at all?...

igorkll:
you check the code, the computer is writing, but this call does not block the work

PR #3594, opened 2023-01-03
Backport text_input event to 1.7.10

embeddedt:
This is the same patch suggested by asie here: https://github.com/MightyPirates/OpenComputers/pull/3572#discussion_r1014625140

I tested that it compiles and works as expected with basic alphabet keys in 1.7.10.


asiekierka:
I have an implementation of this ready locally, it's just stuck on feedback from @payonel

PR #3623, opened 2023-06-06
Fix headers in integration readme

Starlii10:
I was looking through the integration readme for curiosity's sake, then found that the headers within weren't working. On closer inspection, they were missing a space between the hashtags... Whoops.


asiekierka:
Thank you!

PR #3644, opened 2023-08-03
use openContainer instead of sent coords

Glease:
this provide protection against malicious players sending forged packets to mess up unprotected computers without ever getting close to it.


asiekierka:
Thank you!

PR #3647, opened 2023-08-11
Added base ukrainian localization

DinoSlavik:
All but most of the manuals have been translated (only the home page and "basic" manuals from the "general" folder have been translated)


asiekierka:
That's a rather messy PR.

In general, we accept PRs to `master-MC1.7.10`, from which they are ported by us to newer Minecraft versions. The PR should, as such, be derived from `master-MC1.7.10` and prepared for that Minecraft version.

DinoSlavik:
> All but most of the manuals have been translated (only the home page and "basic" manuals from the "general" folder have been translated)

I don't know why the other files are also listed as modified, since I only changed the files necessary for the translation.

DinoSlavik:
> That's a rather messy PR.
> 
> In general, we accept PRs to `master-MC1.7.10`, from which they are ported by us to newer Minecraft versions. The PR should, as such, be derived from `master-MC1.7.10` and prepared for that Minecraft version.

Sorry if this sounds rude, but does this mean I need to create a new PR for `master-MC1.7.10`? I just haven't fully figured out GitHub yet...

PR #3659, opened 2023-09-15
Adjust tank indexing in the Transposer's transferFluid method

yut23:
This updates the `sourceTank` parameter to be 1-based when called from Lua.

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/traits/InventoryTransfer.scala
+++ b/src/main/scala/li/cil/oc/server/component/traits/InventoryTransfer.scala
@@ -48,7 +48,7 @@ trait InventoryTransfer extends traits.WorldAware with traits.SideRestricted {
     val sinkSide = checkSideForAction(args, 1)
     val sinkPos = position.offset(sinkSide)
     val count = args.optFluidCount(2)
-    val sourceTank = args.optInteger(3, -1)
+    val sourceTank = args.optInteger(3, -1) - 1
 
     onTransferContents() match {
       case Some(reason) =>


src/main/scala/li/cil/oc/server/component/traits/InventoryTransfer.scala
```diff
@@ -48,7 +48,7 @@ trait InventoryTransfer extends traits.WorldAware with traits.SideRestricted {
     val sinkSide = checkSideForAction(args, 1)
     val sinkPos = position.offset(sinkSide)
     val count = args.optFluidCount(2)
-    val sourceTank = args.optInteger(3, -1)
+    val sourceTank = args.optInteger(3, -1) - 1
```
asiekierka, line 51: @repo-alt What do you think? This is a change we pulled from GT:NH in 2022, and using 1-based indexes does make sense, but how much production code would breaking this break?

repo-alt: It will break some code of course, and there is no easy way to notify the user about that change, but the change makes sense indeed, so I don't think you should look back at us.

asiekierka: Understood. In this case I'll probably keep it for a larger release, like 1.9.0 (if it ever happens - at this rate, I doubt it!).

yut23:
I'm fine with removing the indexing change for now, but the transfer limit fix is pretty important for precise fluid movement (I originally found it when trying to automate a Tinker's smeltery).

asiekierka:
Understood. I'll try to ship a patch release soon. I apologize for taking so long; the mod no longer has active maintainers, and I don't have much of an interest in Minecraft mod development anymore.

yut23:
I just rebased and updated the PR description to reflect just the indexing change. Thanks for the maintenance you are doing, however minimal it may be :smiley:

PR #3675, opened 2023-11-04
Add MineOS Compat to bios.lua

finnaminope:


PublicVoidUpdate:
Why not just... Rename Mineos 'os.lua' to 'init.lua'

finnaminope:
that will work

PR #3694, opened 2024-03-25
Added a 3d-printer skin upgrade for robots

Compaszer:
This PR adds a new upgrade for robots: The skin upgrade.
This upgrade allows for adding 3d-printed blocks as a skin to robots. They will be rendered at the location of the robot.
In addition 3d-printed blocks can be set as the "arms" of a robot. This is purely decorational, but allows for some cool applications, like building a working GLaDOS from portal (an example can be found below).


asiekierka:
That's a really cool PR for an abandoned mod!

asiekierka:
That's a really cool PR for an abandoned mod!

However, I need to ask: Is there anything in place to ensure that people don't make almost-invisible robots for, say, a PvP context? What I'm trying to say is, I'm not sure if I like the idea of "hiding the full robot body".

src/main/scala/li/cil/oc/server/component/UpgradeSkin.scala
```diff
@@ -0,0 +1,334 @@
...
+      result(true)
+    }
+
+    private def updateClient(): Unit = host match {
+      case robot: internal.Robot => robot.synchronizeSlot(robot.componentSlot(node.address))
+      case _ =>
+    }
+
+    def checkIfTierIsHighEnoughForPart(part: Int): Boolean = {
+      ((part == 2 || part == 5) && tier >= 1) || ((part == 3 || part == 6) && tier >= 2) || (part == 0 || part == 1 || part == 4)
```
asiekierka, line 167: This would probably be much more readable if we had an Enum for the part number; a converter could try to match the integer value to an Enum, or emit an error if it is not a valid part index.

Compaszer: Yeah, I thought about that. The design consideration here was, that you use these numbers in the ingame Lua programs, so it should be referenced as numbers in the code aswell.
But if you disagree, I'll change it asap.

asiekierka: An user can always create their own constants for such numeric values in Lua code - for example, OpenOS comes with `/lib/sides.lua` and `/lib/colors.lua`. I think we should do the same in Scala code; the Enum will document what the part numbers are by virtue of the ordinal.

Compaszer:
> That's a really cool PR for an abandoned mod!
> 
> However, I need to ask: Is there anything in place to ensure that people don't make almost-invisible robots for, say, a PvP context? What I'm trying to say is, I'm not sure if I like the idea of "hiding the full robot body".

Yeah, I didn't thought about this. Maybe the nametag shouldn't be hidable if the body skin occupies less than 10% of the visual space. Because the nametag allways gives away the robots position and therefore should be sufficient for an PvP context.

asiekierka:
That sounds like a good solution; plus a configuration option which allows the server administrator to disable certain forms of "hiding", I think.

PR #3698, opened 2024-04-03
Fix screen turnOn/turnOff return value docs

Hawk777:
These component method calls actually both return two booleans, but only the first one was documented.


asiekierka:
Thank you both!

PR #3702, opened 2024-04-11
Add ability to robots to obtain nearby experience orbs

### LINKED ISSUE #3289: Robots do not gain experience from villager trading
# Outline

I have a robot equiped with an experience upgrade in a Tier 3 upgrade container, as well as an internal trading upgrade. The aim is to have the robot trade with villagers and pick up the experience, leveling up the experience upgrade. This would then be rotated out to other robots.

# Issue:

After executing a trade, experience orbs drop around the villagers. These are not picked up by the robot, unlike in mining, for example. Calling `robot.suck()` on the experience orbs does nothing either.

# Expected behaviour:

The robot should automatically gain the experience resulting from an executed trade, if it has an experience upgrade.

Version information:
`OpenComputers-MC1.12.2-1.7.5.192`

zeng-github01:
Temporary fix #3289

Existing bug: The format of settings.conf is wrong and cannot be parsed correctly, which will cause a crash.

A temporary fix has been made to this bug. Looks like a past patch oversight caused the crash, so it needs to be reviewed

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1652,7 +1652,7 @@ opencomputers {
 
     # This setting assigns the budget call cost to invoke bitblt to write vram
     # to a screen. Video ram can bitblit to a screen which can cause real life
-    # network laod the defaults settings put bitblit network impact close to gpu.set
+    # network load the defaults settings put bitblit network impact close to gpu.set
     # Increase these values to throttle bitblt more. The cost tier N is bitbltCost * 2^(tier)
     # default is .5, which gives: .5, 1, 4
     bitbltCost: 0.5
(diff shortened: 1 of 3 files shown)


zeng-github01:
It looks like the cause of the crash is two missing configuration entries, something the past patch missed.

zeng-github01:
@asiekierka After investigation, Minecraft does not fire any events when a transaction is triggered. Only some experience orbs will be spawned This problem cannot be fixed through ordinary solutions in the current version.

The only solution might be to add a custom event trigger to “useRecipe” through a mixin or ASM

A feasible method is to imitate this submission and add a logic to absorb experience balls in Trade.scala when the complateTrade judgment is completed, but there may be a lag.

Or just like this PR, giving players the power to customize

zeng-github01:
@asiekierka  The patch has been tested and meets the expected results

PR #3705, opened 2024-04-13
Using Leash requires that the drone's inventory already has a Lead.

### LINKED ISSUE #3376: leash upgrade result in infinite leash
One hour ago, in a forge-14.23.5.2847 server with OpenComputer 1.7.5.192, my friends and I decided to transport a few horses from a remote place to our home. 
Since it's a long way, we decided to use a squad of drones. We wrote programs, placed down waypoints, and installed navigation upgrade + leash upgrade in all the drones. Then we started the drones. Everything was fine, and the horses were pulled into the air and being dragged toward our base.
And then, one of my friends got an idea: "Why don't we ride on those horses, so that we don't need to walk by ourselves?!"
We all thought that was a good idea, so we did it. However, just right after we got onto the horse, they(the horses) fell to the ground and broke the leash. However, what surprised us is that a leash *item* dropped to the ground. This made us think if there is a glitch. So we tried again, and that resulted in another leash item dropping out, without damaging the drone or the leash upgrade at all.
And that doesn't look right.
With that glitch, a player can just ride on a horse, then have a drone leash it, then move the drone high up and make the leash broke, and then move back down and repeat it. By doing that, we can get infinite leash. 

We hope the leash upgrade could be re-implemented to solve this problem:

- The leash upgrade now requires another call to a method that makes it take some leash item from the drone's internal storage, just like the generator upgrade.
- when the leash method is called successfully, a leash is consumed
- when the unleash method is called successfully, a leash is returned
- in this way, if the leash broke, the program won't call the unleash method. Thus the leash won't be returned, and it won't dupe.

That's it. Also, thanks for the great mod! It was fun writing programs to solve 'real world' issues :)

zeng-github01:
fix #3376

Requires review and testing

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/UpgradeLeash.scala
+++ b/src/main/scala/li/cil/oc/server/component/UpgradeLeash.scala
@@ -2,7 +2,6 @@ package li.cil.oc.server.component
 
 import java.util
 import java.util.UUID
-
 import li.cil.oc.Constants
 import li.cil.oc.api.driver.DeviceInfo.DeviceAttribute
 import li.cil.oc.api.driver.DeviceInfo.DeviceClass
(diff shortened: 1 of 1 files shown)


zeng-github01:
This patch looks a bit rough, I don't know if there are any performance issues

zeng-github01:
Use the following code to test, it has no effect

`local leash
local drone

for address, componentType in component.list("leash") do
    leash = component.proxy(address)
end

for address, componentType in component.list("drone") do
    drone = component.proxy(address)
end

drone.move(0,2,0)


while true do
    leash.leash(0)
end`

PR #3706, opened 2024-04-14
Add checkDamage Boolean for li.cil.oc.util.InventoryUtils#haveSameItemType

### LINKED ISSUE #3288: Incorrect villager trade emulation in Trading Upgrade
OC: OpenComputers_MC1.7.10_1.7.5.1290_universal
MC: 1.7.10

By default villagers ignore metadata. So u could trade with them by using different items for example:
Charcoal instead of coal,
broken tools,
etc.
Component trading doesn't ignore metadata.
I think its a defect and should be fixed in the next release.
The main reason why it is so important.
image

zeng-github01:
maybe fix #3288 

Need to test

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/traits/InventoryAnalytics.scala
+++ b/src/main/scala/li/cil/oc/server/component/traits/InventoryAnalytics.scala
@@ -49,7 +49,7 @@ trait InventoryAnalytics extends InventoryAware with NetworkAware {
     DatabaseAccess.withDatabase(node, dbAddress, database => {
       val dbSlot = args.checkSlot(database.data, 2)
       val dbStack = database.getStackInSlot(dbSlot)
-      result(InventoryUtils.haveSameItemType(localStack, dbStack, args.optBoolean(3, false)))
+      result(InventoryUtils.haveSameItemType(localStack, dbStack, args.optBoolean(3, false), checkDamage = true))
     })
   }
 }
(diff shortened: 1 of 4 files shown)


asiekierka:
While this PR may help in a specific case, I am worried other mods may be patching the villager trade code to check damage for their specific items - as such, fixing this could lead to more devastating issues in cross-mod interaction.

In my opinion, the best way would be to try and re-use the same codepath vanilla uses for executing villager trades, but as I outlined in the initial comment - this would probably not be too easy.

zeng-github01:
> While this PR may help in a specific case, I am worried other mods may be patching the villager trade code to check damage for their specific items - as such, fixing this could lead to more devastating issues in cross-mod interaction.
> 
> In my opinion, the best way would be to try and re-use the same codepath vanilla uses for executing villager trades, but as I outlined in the initial comment - this would probably not be too easy.

You mean there might be other mods that modify the OC code? That would cause glitches?

zeng-github01:
@asiekierka I have a new idea, can you help me test it? I will submit a new commit about it soon

[Test branch](https://github.com/zeng-github01/OpenComputers/tree/Vanilla-trading-test)

Considering that this patch is very irregular, I chose to submit it to the test branch
I found the vanilla transaction code by querying useRecipe, the above code is an attempt to manually build the relevant class

zeng-github01:
> While this PR may help in a specific case, I am worried other mods may be patching the villager trade code to check damage for their specific items - as such, fixing this could lead to more devastating issues in cross-mod interaction.
> 
> In my opinion, the best way would be to try and re-use the same codepath vanilla uses for executing villager trades, but as I outlined in the initial comment - this would probably not be too easy.

I think this problem can be solved by setting checkDamage to true by default. This way, no changes will be required to the old code.

Issue #3663, opened 2023-09-20
Internet API Throws Exception for HTTP Error Response (again)


kodai2199:
I'm trying to use a REST API I built, but I cannot handle HTTP error codes in-game, because attempting to iterate over the request handle throws the following error: Server returned HTTP response code: 400 for URL: ..., and getmetatable(handle).__index.response() simply returns nil so I am unable to check for certain response codes there either.

I believe this issue was said to be fixed here [https://github.com/MightyPirates/OpenComputers/issues/3398](https://github.com/MightyPirates/OpenComputers/issues/3398). However,  in the latest version, this bug is still here. I am tempted to look at the differences between the relevant files in the latest release and in the older ones to see what's going on.

I am now using [https://http.codes/](https://http.codes) for testing purposes, and I can confirm that I can correctly get codes for responses like 200 OK. But with 400, 404, ... I only get nil, and an error if I try to read the response.
Thank you in advance for your support!

kodai2199:
For now I was able to circumvent the problem with a pretty ugly solution. I parse the response in a protected call, and if I get an error, I use string.match() to find the error code within the error message. Unfortunately the error message changes for codes 404 and 410, leaving no information on which of those two error codes was received. Since 404 is way more frequent, I decided to return 404 in that situation.

Of course a fix would be better.
```
function requestWithResponseCode(url, post_data, headers, method)
    local handle = internet.request(url, post_data, headers, method)
    local result = ""
    local success, error = pcall(function() for chunk in handle do result = result..chunk end end)
    if success == false and error ~= nil then
        local error_code = error:match("code: (%d%d%d) ")
        if error_code ~= nil then
            return error_code, result
        else
            return 404, result
        end
    end

   if success == true then
       local mt = getmetatable(handle)
       local code, message, headers = mt.__index.response()
       return code, result
   end
end
```

asiekierka:
It was fixed, but then it was reverted as not every Minecraft configuration actually had the Apache HTTP client library, nor could I figure out in the limited maintenance time I have for the mod: https://github.com/MightyPirates/OpenComputers/commit/555ae0e695db0ae252aa153a4a8f330568c41dab

asiekierka:
The problem, I suppose, is that:

* a bunch of programs (like wget) depend on the old behaviour to some extent (they don't check the response code at all, assuming exception = failure),
[...]

PR #3708, opened 2024-04-15
Work around #3663 - allow accessing getErrorStream() for reading unsu…

zeng-github01:
…ccessful HTTP responses' output,

@asiekierka  Unable to build for unknown reasons. Therefore not tested

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/InternetCard.scala
+++ b/src/main/scala/li/cil/oc/server/component/InternetCard.scala
@@ -1,36 +1,22 @@
 package li.cil.oc.server.component
 
 import com.google.common.net.InetAddresses
+import li.cil.oc.{Constants, OpenComputers, Settings}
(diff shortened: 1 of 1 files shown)


asiekierka:
This was deliberately (albeit not definitively) not done in this way. Doing so breaks the expectations of `wget` and - presumably - many other OpenOS applications. As OpenComputers 1 is effectively on long-term support and only available for legacy Minecraft versions, not breaking API compatibility is much more important than adding new functionality.

zeng-github01:
> This was deliberately (albeit not definitively) not done in this way. Doing so breaks the expectations of `wget` and - presumably - many other OpenOS applications. As OpenComputers 1 is effectively on long-term support and only available for legacy Minecraft versions, not breaking API compatibility is much more important than adding new functionality.

Looks like this is meant to just fix the exception and discard all but the 200 response.

PR #3725, opened 2024-07-01
Add Proper Hostname Support

PewPewCricket:
The seemingly half-baked lua code that is meant to handle hostnames in /boot/94_shell.lua does not work, I have moved hostname setting to /etc/profile.lua and it simply reads one line from /etc/hostname and sets $HOSTNAME to that, the default $HOSTNAME_SEPARATOR is now ": ", you can set your hostname by running this command: echo "hostname" > /etc/hostname

After changing you hostname to "foobar" using the above method, upon reloading of the shell, your new prompt would be: foobar: /home #

This is a very minor and easy improvement and i hope it will not be too much trouble to merge into OpenOS, I have tested this and it works on my OC systems in 1.7.10


PewPewCricket:
image
here is an image of a system with a hostname set to "SystemM"

PR #3732, opened 2024-08-17
Added Ukrainian localization

dnrovs:


asiekierka:
For upstream localization PRs, we need both a 1.7.10 and a 1.12.2 PR, as both versions are still kept in sync.

Issue #2733, opened 2018-01-27
[Suggestion] Filter-Upgrade for Relays


GHXX:
# Why is this suggestion useful?
From what i have seen it is not possible to specify the number of times a packet may pass through relays.
I am assuming this is because people might set it to a number that's extremely high which would cause a massive server load if loops happen. Maybe we could have a card that re-sets the lifetime counter 
(back to 5). This however wouldn't solve the loop issue. 

So what if instead its a card the both resets the counter AND is able to detect looping. If it loops through it 5 times or more its being blocked and not relayed.

# What is my suggestion?
- Add a new upgrade card for relays
- This card will reset the relay-counter so it can be relayed another 5 times. If it goes through another relay with this upgrade the counter will reset again
- To avoid loops there should be a filter functionality built into the card which is able to detect looping packets and blocks them if the very same packet goes through it 5 or more times.

So what do you think about it?

# Alternative solution

An alternative would be to add some sort of custom switch, maybe a network interface or something like that. Basically a micro controller that can interact with network cables. This Block however would have to be able to choose where the packet came from (effectively spoofing the sender address).

skyem123:
Ideally you would wire the system in a way that doesn't loop, although that can get tricky... especially with wireless networks! (maybe that is something else to wonder about...)

The way I would rather it be done is by allowing people to make their own switches by being able to sniff and spoof packets, but other people here don't like my idea. :p

Maybe the "card" could instead be a fully programming thing, similar to a microcontroller? That would have some interesting implications...

Sorry for kinda hijacking this...
oh also why 5 times, specifically?

GHXX:
as the wiki (iirc) and fnuecke say:

> Sounds like you have a loop in your network, in which case that's expected and intended behavior. Network packets have a TTL of 5 hops, so this sounds like the switch is forwarding it those 5 times.

But yeah, custom relays could be very cool actually. Ill update the issue.

skyem123:
oh I see now, a card that does more detailed inspection of packets, while also resetting the TTL, I think it's not even needed to have the the limit as high as 5?
The main problem I see is that what if people do send the same message over and over and want their data back? That would require every packet to have a hidden unique ID which is... eeeh.

GHXX:
or there could be a hidden, salt (a value that counts up converted to letters for example) added to the packet. And a TTL of 5 wont hurt. So you dont need a special switch every hop.

GHXX:
[...]

PR #3745, opened 2024-11-10
Added configurable network packet TTL

timothe-grisot:
While I'm having a great time playing with OpenComputers in 1.12.2, I think the default network packet TTL is quite restrictive when thinking about building huge networks.
Upon searching for a way to fix the issue, I stumbled upon issue #2733 which is kind of related to what I wanted (that is, simply a higher base TTL).

I decided to take a quick look at the source code to see where the hard-coded TTL value was. As I was feeling quite comfortable modifying the code, I decided to make all changes in order to make it configurable by players.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1259,6 +1259,13 @@ opencomputers {
     # string from the clipboard (Shift+Ins on a screen with a keyboard).
     maxClipboard: 1024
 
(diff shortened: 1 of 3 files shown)


asiekierka:
https://github.com/MightyPirates/OpenComputers/commit/d19255dda9c4c40f152181fdd484ac945c6b043a

asiekierka:
Thank you!

PR #3752, opened 2025-01-07
Maybe fix Coroutine.wrap?

### LINKED ISSUE #3687: Coroutine.wrap does not propagate errors raised in a coroutine
The following code should raise an error in the main thread, propagating one thrown by the coroutine body:

```lua
print(coroutine.wrap(function()
  error("test")
end)())

print("should be unreachable")
```

In native Lua and in the BIOS environment none of the print calls are reachable. In OpenOS, however, the wrapper intercepts the error and returns the error object; the code above thus prints the following:

```
/home/test.lua:2: test
should be unreachable
```

OC 1.8.3. I think this may be a regression in one of the newer OC versions — I certainly remember `coroutine.wrap` working as expected a while ago (1.7-ish).

zeng-github01:
Fix #3687

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/boot/01_process.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/boot/01_process.lua
@@ -52,7 +52,9 @@ end
 _coroutine.wrap = function(f)
   local thread = coroutine.create(f)
   return function(...)
-    return select(2, coroutine.resume(thread, ...))
+    local result, reason = coroutine.resume(thread, ...)
+    assert(result, reason)
+    return reason
   end
 end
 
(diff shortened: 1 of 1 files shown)


SquidDev:
I think this needs to handle the case where the coroutine returns multiple values. I think if you use `table.pack`, you can get something closer to what PUC Lua does:

```lua
    local result_pack = table.pack(coroutine.resume(thread, ...))
    if not result_pack[1] then error(result_pack[2], 2) end
    return table.unpack(result_pack, 2, result_pack.n)
```

zeng-github01:
> I think this needs to handle the case where the coroutine returns multiple values. I think if you use `table.pack`, you can get something closer to what PUC Lua does:
> 
> ```lua
>     local result_pack = table.pack(coroutine.resume(thread, ...))
>     if not result_pack[1] then error(result_pack[2], 2) end
>     return table.unpack(result_pack, 2, result_pack.n)
> ```

This looks like a restoration of the past implementation.
Why did they change it to this? It's really strange.

Do you consider submitting a PR? I'm not good at Lua. I'm worried about new problems.

PR #3755, opened 2025-01-09
[1.7.10]Fixed villager trading being inconsistent with vanilla

### LINKED ISSUE #3288: Incorrect villager trade emulation in Trading Upgrade

zeng-github01:
Fix #3288

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/Trade.scala
+++ b/src/main/scala/li/cil/oc/server/component/Trade.scala
@@ -78,7 +78,7 @@ class Trade(val info: TradeInfo) extends AbstractValue {
                     if (!hasRoomForRecipe(inventory, recipe)) {
                       result(false, "not enough inventory space to trade")
                     } else {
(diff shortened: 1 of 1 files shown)


zeng-github01:
@asiekierka I've come up with a better fix. This doesn't break the existing API. The only cost is creating your own PR for both versions of Minecraft.

I have created separate PRs for both versions.

zeng-github01:
@asiekierka I've come up with a better fix. This doesn't break the existing API. The only cost is need creating  PR for both versions of Minecraft.

I have created separate PRs for both versions.

This patch does not modify the internal API. In theory, it will not cause compatibility issues.

PR #3767, opened 2025-03-25
fix openos error handling

### LINKED ISSUE #3727: [1.12.2] Regression in 1.8.4 related to OpenOS exception handling
In 1.8.4, when I cause an exception to be thrown in OpenOS, occasionally the console will be spammed with text, which boils down to the following error: `/lib/process.lua:63: in function </lib/process.lua:59>attempt to yield across a C-call boundary`.

image

This does not happen in 1.8.3, _but it requires a fresh install of OpenOS_, for unknown reasons. This commit https://github.com/MightyPirates/OpenComputers/commit/9d4f7ea297953c2fd8ccfd24fe549d5e9576400f looks potentially suspicious as a cause for the problem as it was introduced in 1.8.4.

RedstoneParkour:
this is a different fix than [the one described by Jason](https://github.com/MightyPirates/OpenComputers/issues/3727#issuecomment-2601843881), but it seems to work as well
i haven't done a lot of testing of the fix though, so there may be a (smaller) bug left

closes #3727

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/process.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/process.lua
@@ -68,15 +68,15 @@ function process.load(path, env, init, name)
           if type(msg) == "table" and msg.reason == "terminated" then
             return msg.code or 0
           end
-          return debug.traceback()
+          return {msg, debug.traceback()}
         end, ...)
     }
 
-    if not result[1] and type(result[2]) ~= "number" then
+    if not result[1] and type(result[2]) == "table" then
       -- run exception handler
       xpcall(function()
-          local stack = result[2]:gsub("^([^\n]*\n)[^\n]*\n[^\n]*\n","%1")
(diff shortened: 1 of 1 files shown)


JasonS05:
I haven't tested it, but looking at the code I'm concerned that it won't correctly report an `out of memory` error. Why not just make a pull request with my fix? Although I realize now that I should have done that in the first place.

asiekierka:
@JasonS05 This one seemed more idiomatic due to not adding another `local`, but instead passing everything in a table. They seemed mostly identical to me other than that.

JasonS05:
The issue is that if an out of memory error occurs then the error handler is *not* called, but instead `xpcall` simply returns an out of memory error message. With this code I don't see how that would get reported since the `type(result[2]) == "table"` check will fail. With my fix, the error handler just returns the error message instead of a table meaning that if an out of memory error occurs, it will still work out properly since the code already expects `xpcall` to just return a string.

PR #3770, opened 2025-04-01
Fix #3769

### LINKED ISSUE #3769: Empty error when using pastebin.
Image

The computer returns an empty error with a callstack.
Tested on a fresh 1.12.2 forge 14.23.5.2860 and also on FTB Ultimate Reloaded.

ff66theone:
Added user agent of a real browser (mine) to bypass what looks like an **anti-scraping protection**.
This is an *experimental* fix to avoid the 403 HTTP (forbidden access) code that is the hidden origin of that issue.
**Tested with the MineOS installer and the pastebin tested from #3769.**


asiekierka:
I'll solve it in a different way. Hard-coding an user agent that doesn't belong to us is not very nice; CC:Tweaked uses "computercraft/$version" and that seems to work fine for them, so I'll add a default user agent configuration option to match (and to fix legacy software!).

PR #3774, opened 2025-04-11
Fix Juke box driver on 1.12

kebufu:
Since minecraft 1.12 changed the SFX ID for the jukebox from 1005 to 1010 and reassigned the old SFX ID to the iron door opening sound, the OpenComputers adapter can no longer control jukebox playback properly. Attempting to play or stop playback triggers the iron door opening sound instead. This commit fixes the issue. It is currently unclear whether this issue exists in other versions.

Relevant code in 1.7.10:
image

Relevant code in 1.12:
image
image


asiekierka:
Thank you very much!

PR #3778, opened 2025-04-27
Fix #3682

### LINKED ISSUE #3682: the "flash" program in openOS does not warn that the EEPROM is readonly
when trying to flash read only eeprom, the flash program does not display a message stating that the flash failed

ff66theone:
Potential fix for #3682, declaration of `eeprom` moved to main chunk.

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/bin/flash.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/bin/flash.lua
@@ -5,11 +5,6 @@ local eeprom = component.eeprom
 
 local args, options = shell.parse(...)
 
--- No funtion available to check if the EEPROM is read-only
-if not eeprom.setLabel(eeprom.getLabel()) then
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/opencomputers/loot/openos/bin/flash.lua
```diff
@@ -53,8 +57,6 @@ local function writeRom()
     io.write("Beginning to flash EEPROM.\n")
   end
 
-  local eeprom = component.eeprom
-
```
asiekierka, line 57: This breaks the ability of an EEPROM to be inserted after starting the program, but before flashing.

ff66theone: I moved the warning message to that function and a line that I added should fix this issue.

asiekierka:
I made a slightly more future-proof fix.

PR #3779, opened 2025-04-30
fix `os.sleep(0)` causing `too long without yielding`

### LINKED ISSUE #3773: Asking community for help with debugging and fixing .lua script that crashes on new version of OC
Hello!

There is mod called "Just a Stargate Mod" and for the "Stargate" they made a lua script AGS (AuspexGateSystem) and it works perfectly fine on version 1.8.6. (1.8.7. & 1.8.8.) and then any version after that one it keeps freezing and crashing during use, we have been yet to solve the issue as we can't find any apparent cause/reason of the issue to occur. Any feedback would be useful. Code can be found [here](https://github.com/TheRedCreations/AuspexGateSystems/blob/release/ags/SG_Dialer.lua) (GitHub Repository Link).

Thanks in advance,
Luke

Ocawesome101:
It appears commit [7fecb07](https://github.com/MightyPirates/OpenComputers/commit/7fecb0732ce36b2e6246b5fd85ca298e00a79e2b) broke `os.sleep(0)`:

```lua
while true do os.sleep(0) end
```

as of oc 1.8.7/1.8.8 this cannot be interrupted, and will crash with `too long without yielding`, where in 1.8.6 and before this was not the case.  may fix #3773

### DIFF
--- a/src/main/resources/assets/opencomputers/loot/openos/lib/event.lua
+++ b/src/main/resources/assets/opencomputers/loot/openos/lib/event.lua
@@ -143,7 +143,7 @@ function event.pullFiltered(...)
   local deadline = computer.uptime() + (seconds or math.huge)
   repeat
     local waitTime = deadline - computer.uptime()
-    if waitTime <= 0 then
+    if waitTime < 0 then
       break
     end
     local signal = table.pack(computer.pullSignal(waitTime))


TheRedCreations:
Hi. AGS-Dev here. 
You are right about the broken os.sleep(). I just changed the os.sleep() to os.sleep(0.1) and the program works again.
Thank you.

ff66theone:
Why not using `os.sleep(0.05)` since a Minecraft ticks lasts 1/20 of a second (0.05s) ?
I know it only gains 0.05 seconds, but that's nearer from the pre-1.8.6 version.

TheRedCreations:
I got told that 0.05 wouldn't work and 0.1 is minimum. but it works with 0.05 too.
Maybe i change that, or leave it as it is cause its only 0.05 difference.

zeng-github01:
This is great, I have an unattended program that has this problem and it will definitely trigger this problem after running for a long time.

I use OC to control the reactor, and this problem has caused my reactor to melt down many times.

Thank you so much

PR #3782, opened 2025-06-07
Fix HTTP error response handling in Internet API (#3663)

### LINKED ISSUE #3663: Internet API Throws Exception for HTTP Error Response (again)

31415-n:
Previously, HTTP error responses (4xx, 5xx) would throw exceptions instead of allowing access to response information. This made it impossible to handle API errors gracefully in Lua scripts.

Changes:
- Capture response code/message/headers before attempting to read stream
- Use getErrorStream() for HTTP error responses instead of getInputStream()
- Provide empty stream fallback when getErrorStream() returns null
- Preserve existing behavior for successful responses (2xx)

This allows scripts to check response codes and read error response bodies using the existing response() and read() methods.

Fixes #3663

### DIFF
--- a/src/main/scala/li/cil/oc/server/component/InternetCard.scala
+++ b/src/main/scala/li/cil/oc/server/component/InternetCard.scala
@@ -525,21 +525,25 @@ object InternetCard {
               out.close()
             }
 
-            // Finish the connection. Call getInputStream a second time below to re-throw any exception.
-            // This avoids getResponseCode() waiting for the connection to end in the synchronized block.
-            try {
-              http.getInputStream
-            } catch {
-              case _: Exception =>
-            }
-
             HTTPRequest.this.synchronized {
               response = Some((http.getResponseCode, http.getResponseMessage, http.getHeaderFields))
             }
 
-            // TODO: This should allow accessing getErrorStream() for reading unsuccessful HTTP responses' output,
-            // but this would be a breaking change for existing OC code.
-            http.getInputStream
+            // For successful responses (2xx), use getInputStream()
+            // For error responses (4xx, 5xx), use getErrorStream() if available, otherwise return empty stream
+            val responseCode = http.getResponseCode
+            if (responseCode >= 200 && responseCode < 300) {
+              http.getInputStream
+            } else {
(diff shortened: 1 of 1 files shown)


asiekierka:
> // TODO: This should allow accessing getErrorStream() for reading unsuccessful HTTP responses' output,
> // but this would be a breaking change for existing OC code.

The problem with the solution is still that a lot of existing OC programs *do not* handle error responses which are not exceptions; since the mod is on life support, breaking existing OC programs is unacceptable.

31415-n:
Okay, I’ve added a new parameter to the request method. This change maintains full backward compatibility with existing code.

You can run this Lua script with automated tests to verify the behavior: https://pastebin.com/v6jpmDS7

The test confirms that backward compatibility is preserved: when allowErrorBody is set to false (or not provided at all in the request() method), only HTTP status codes 200, 201, and 202 are treated as successful responses. All other codes (such as 400, 404, 500, etc.) result in an exception being thrown, which matches the original behavior.

However, when allowErrorBody is set to true, the method follows the new logic: even for HTTP error responses (4xx, 5xx), the response body is returned instead of throwing an exception. This allows developers to handle error responses programmatically without interrupting execution.

In summary, the addition of the allowErrorBody parameter enhances the flexibility of the request method while preserving full backward compatibility.

PR #3783, opened 2025-06-07
Fix TCP socket finishConnect() exception handling and connection logic

31415-n:
- Fix bug where finishConnect() required multiple calls before throwing exceptions
- Fix logic where address resolution didn't immediately attempt connection
- After address is resolved, immediately call channel.finishConnect()
- Ensures immediate error reporting for failed socket connections
- Improves reliability of socket connection error handling


asiekierka:
Could you provide an example scenario in which this issue manifests?

PR #3793, opened 2025-08-15
Fix Redstone IO Block particle missing on destory

### LINKED ISSUE #3792: Redstone IO Block texture issue
When the redstone IO is destroyed, the particle texture is wrong. It appears as a textureless particle.

zeng-github01:
Fixes #3792


asiekierka:
Thank you!

Issue #1112, opened 2015-04-29
Non-blocking computer.beep()


SolraBizna:
I'm working on... something semi-secret. It would greatly benefit from a way to beep without blocking the whole machine. Currently, `computer.beep()` blocks until the beep is finished.

Two possible approaches:
- Make `computer.beep()` work differently somehow; for instance, blocking only until the _previous_ beep has concluded.
- Add an API that would let an Architecture module emit its own beeps, requiring the Architecture module to do whatever bookkeeping/throttling makes sense for that Architecture.

I, personally, would prefer the second method as it would grant a great deal of flexibility.

Flood of other issues coming soon, after the semi-secret thing is released :)

Vexatos:
Check out the Computronics Beep Card. It does up to 8 beeps at the same time with different lengths, also the calls are non-blocking. For looking for how to use it, you may want to check out my [Song API](https://github.com/OpenPrograms/Vexatos-Programs/blob/master/song/song.lua) which allows playing whole songs with this card (See the comment in the first lines for how to use it if you are interested).

fnuecke:
For now you can work around this by using OC's internal `li.cil.oc.server.PacketSender.sendSound` methods. I guess I could add them to the API somehow. Either globally or via the `Machine` interface... the former obviously being more flexible, but needing some kind of `Util` API thinger, which I kind of dislike, because that'll attract clutter like a supermagnet :P

SolraBizna:
Vexatos: The Beep Card is in an addon mod, though, and I'd prefer a solution that doesn't require one. _However_, now that I know it exists, I won't design polyphonic sound into my Architecture; and I'll be adding optional support for the Beep Card (for improved sound quality) to the semi-secret thing... assuming, that is, I can _find_ a copy of Computronics, as the website seems to be down...

fnuecke: It makes more sense conceptually to put it in the Machine interface, I think. Meanwhile, I'll try using `PacketSender.sendSound`.

fnuecke:
Yeah, the old Computronics page has been dead for a while now, but a lot of stuff still links there.
Current one is http://wiki.vex.tty.sh/

Vexatos:
I changed the URL everywhere I have acess to, sorry for the inconveience...

If you are interested, [this](https://github.com/asiekierka/Computronics/blob/master/src/main/java/pl/asie/computronics/oc/DriverCardSound.java#L129-L162) is how Computronics handles beeping.

igorkll:
[...]

PR #3798, opened 2025-09-16
Add async option to beep function to fix #3789

### LINKED ISSUE #3789: Change computer.beep() to be non-blocking system-wide
OpenComputers Mod Version: 1.8.9a for Minecraft 1.12.2
OpenOS version of test system: OpenOS 1.8.8
LUA Version: Lua 5.3

As discussed in issue [#1112](https://github.com/MightyPirates/OpenComputers/issues/1112) `computer.beep()` is a blocking call. On OpenComputers version 1.8.9a `computer.beep()` will block all execution on the computer until the call completes, including if `computer.beep()` was called from a detached thread.

I would appreciate it if instead of `computer.beep()` blocking execution system-wide, calling it only blocks within the thread it was executed from. The consequence of this change would enable users to have multiple instances of `computer.beep()` running concurrently using threads, which I believe would be an improvement to the current system.

I understand that the Computronics mod by Vexatos adds a beepcard that is non-blocking. However, the behavior of `computer.beep()` in the mainline mod should still be changed so that it at least does not block execution system-wide.

The following example code shows `computer.beep()` blocking execution of the main program thread while it runs within the detached thread:
```lua
local computer = require("computer")
local event    = require("event")
local thread   = require("thread")

local play_audio_tone = thread.create(function(tone_id)
  -- tone 1
  -- 340Hz 500ms on, 500ms off Intermittent
  local function tone1()
    local t = 0

    for i = 1, 4 do
      if (i % 2) == 1 then
        computer.beep(340, 0.50)
      else
        t = computer.uptime()
        repeat os.sleep(0) until( ( computer.uptime() - t ) >= 0.50 )
      end
    end
    computer.pushSignal("tone_complete", 1)
  end

  local PLAY_TONES = {
    [1] = function() tone1() end
  }

  -- crude system for picking which audio tone to play for testing using args passed from shell
  if type(tone_id) == "number" and tone_id > 0 and tone_id <= 8 then
    --event.timer(1, PLAY_TONES[tone_id)
    PLAY_TONES[tone_id]()
  else
    print("Invalid argument")
  end

  while true do
    local event, param1 = event.pullMultiple("interrupted", "tone_complete")

    if event == "interrupted" then
      print("interrupted")
      break
    elseif event == "tone_complete" then
      PLAY_TONES[param1]()
    end
  end
end, 1):detach()

local iter = 0

while true do
  os.sleep(1)
  print(iter)
  iter = iter + 1
end
```

Ericgi231:
Added an async flag to the existing Lua computer.beep() for non morse code beeps.
While it may introduce the potential for spamming beeps of different frequency, users could likely already exploit the morse code beeps which were already non-blocking by default,

Maintains full backwards compatibility, default behavior without new argument is the same as before.

I have not written tests or done manual testing, which would seem to be the standard.

To use in Lua - `computer.beep(340, 0.50, true)`

### DIFF
--- a/src/main/scala/li/cil/oc/server/machine/Machine.scala
+++ b/src/main/scala/li/cil/oc/server/machine/Machine.scala
@@ -443,7 +443,7 @@ class Machine(val host: MachineHost) extends AbstractManagedEnvironment with mac
   def isRunning(context: Context, args: Arguments): Array[AnyRef] =
     result(isRunning)
 
-  @Callback(doc = """function([frequency:string or number[, duration:number]]) -- Plays a tone, useful to alert users via audible feedback.""")
+  @Callback(doc = """function([frequency:string or number[, duration:number[, async:boolean]]]) -- Plays a tone, useful to alert users via audible feedback.""")
   def beep(context: Context, args: Arguments): Array[AnyRef] = {
     if (args.count == 1 && args.isString(0)) {
       beep(args.checkString(0))
@@ -454,7 +454,10 @@ class Machine(val host: MachineHost) extends AbstractManagedEnvironment with mac
       }
       val duration = args.optDouble(1, 0.1)
       val durationInMilliseconds = math.max(50, math.min(5000, (duration * 1000).toInt))
-      context.pause(durationInMilliseconds / 1000.0)
+      val async = args.optBoolean(2, false)
+      if (!async) {
+        context.pause(durationInMilliseconds / 1000.0)
+      }
       beep(frequency.toShort, durationInMilliseconds.toShort)
     }
     null


ReallySecureShell:
This pull request is in response to [Issue #3789 ](https://github.com/MightyPirates/OpenComputers/issues/3789)

`computer.beep()` was added in [Commit-d5f1280](https://github.com/MightyPirates/OpenComputers/commit/d5f1280fc9c9f463c45eff76119cfc513d8c3fd5) in response to [Issue #1112](https://github.com/MightyPirates/OpenComputers/issues/1112). The SCALA function for `computer.beep()` can be found on [Lines 446 through 461 in Machine.scala](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.12/src/main/scala/li/cil/oc/server/machine/Machine.scala#L446-L461), as referenced by this pull request.

As far as I can tell, `computer.beep(frequency, duration)` was made to be a blocking call over concerns of potentially high network packet generation, and the processing resources required by clients to generate the tone in real-time [source 1](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.12/src/main/java/li/cil/oc/api/machine/Machine.java#L147-L163). I figure the idea was that by blocking execution on the machine instance while the tone is played would result in more processing resources being available to generate the real-time tone.

In source 1, for the comment talking about *`void beep(String pattern);` it says "It has the advantage of only generating a single network packet, and generating a single, longer sound sample for the full pattern". Which may explain why `computer.beep(string)` was decided to be non-blocking.

Furthermore, apart of the comment for **`void beep(short frequency, short duration);` in source 1 reads "It is therefore recommended to not call this too frequently, and to limit the length of the sound to something relatively short (not longer than a few seconds at most)." So, if making `computer.beep(frequency, duration)` non-blocking, then putting a limit on the maximum duration of the tone seems to be the best idea.

Because it may also be of interest [this is the server-side packet constructor for building the sound packet](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.12/src/main/scala/li/cil/oc/server/PacketSender.scala#L807-L818).

*When referencing `void beep(String pattern);` in context to source 1, that method is the same as `computer.beep(string)`

**When referencing `void beep(short frequency, short duration);` in context to source 1, that method is the same as `computer.beep(frequency, duration)`

Ericgi231:
I agree putting a limit on the maximum duration of the tone would be best practice.
My understanding of the use case is not informed enough to suggest what that max should be, as such I left it un-capped.

I would leave it up to the owners discretion what magic number would best fit.

Zaklinatel:
Bumping the PR.
Looking forward to see this in the main branch.

ff66theone:
The audio patterns and "normal" beeps seem to be decoded client-side ([here](https://github.com/MightyPirates/OpenComputers/blob/master-MC1.12/src/main/scala/li/cil/oc/util/Audio.scala#L44)), so both using patterns and a single beep generate a single packet each. The only reasons I could find of making beeps blocking would be to avoid 2 beeps overlapping each other at different tones (which would be, let's be honest, unrealistic since computer buzzers aren't multi-channel audio outputs), or to avoid the server kicking/DDoSing you for playing a music file (I made such a thing happen once using OpenComputers + OpenFM + my local Apache server, so watch out).
Playing strings being currently already non-blocking, so spamming dots (say, by making a cursed call like `computer.beep("..................................")`) would eat a bit more resources that the "normal" call, and you're allowed to stack multiple ones on top of each others.
This is the reason why I'd say it would be better to implement a maximum call number of this function per second, because you can already spam strings if you want to. Long beeps have never been an issue, it's simply packet spam. The non-pattern call could have been used for music and would be able to kick users without them thinking about it, this is why it was made blocking. The pattern call wasn't because it is fiiinneee for music applications, as long as you don't call it for a single character at a time.
Btw there's also a funny thing, being that the client side supports playing patterns with different tones and "base lengths" (length of a dot) than the default, but I've never seen anything use that feature. Could be useful one day, maybe...?

Zaklinatel:
Audio could be implemented as a dedicated component with its own buffer. Reading that buffer linerly rules out the possibility to play two sound at a time.
OR implement a queue and a function like "queueSound" which queues the sound to play when device is ready to
OR ignore other `beep(...)` calls while the previous sound is playing and add a callback to the call so we could know when it's done.

ff66theone:
I don't think a callback is necessary. Just use signals. This is pretty much the same thing as making the soundcard throw an interrupt when it finished its audio buffer so the CPU can refill it (though signals have to be polled, while interrupts don't wait for you).

Zaklinatel:
Well it doesn't matter which way we use it since it still blocks the main thread because `beep` calls `context.pause(...)` under the hood.
The main point is to allow playing sounds without pausing the global context (basically the computer) - doesn't matter how (as devs decide) - just to be able to do this, so we could use it in games, for example, play soundtracks and some sfx. For now the whole system stucks for the `length` time, while the sound is playing, which makes beeping vestigial feature and basically, - hear me out! - **useless**.
I really had to make a separate device based on a microcontroller and EEPROM to act as a dedicaed sound card so I could play sounds over network without blocking. But this is so cooked. Of course there are problems with sync and delays, because of network lags and necessety to passively pull events and spend that microscopic CPU time I have for checking "is there a new sounds to play? Oh well lets spend more time for inserting the table" instead of playing notes *in time* from the buffer or something.

Talking about features, I would be glad to see something like microcontrolles integrated into programmable extension cards so we could use it in slots, just like graphics cards, but they could provide shared buffers or API calls to the system and be able to work seaprately, like deicated device. This way we could create our own sound cards, co-processors, or any kind of stuff. As for me it's a great step to extending the gameplay with this mod.

